PR #8, opened 2024-06-11
Fix model unlock on exception

### LINKED ISSUE #6: Model remains locked on erroneous context exit
An error in the lock() context will exit the context but leave the model locked.

```
>>> model = Model()
>>> with model.lock():
            duckbilledplatypus
```

exits context with error `NameError: name 'duckbilledplatypus' is not defined`  but keeps the model locked:

```
>>> model.is_locked()
True
```

Should the lock context have a `try` context so it can exit gracefully (undo the lock)?

randomir:
Fix #6.


JoelPasvolsky (approved):
Thank you!

arcondello (approved):

PR #13, opened 2024-06-12
Rename NodeObserver to Symbol and ArrayObserver to ArraySymbol

arcondello:
I think these names are much clearer.

These are both "internal" classes, in that users should not interact with them directly except via their subclasses. So I don't believe this will break any testing code.


JoelPasvolsky (approved):
Much better!

PR #26, opened 2024-06-12
Fix use of delvewheel with a namespace package

arcondello:
See https://github.com/dwavesystems/dwave-preprocessing/pull/131, https://github.com/adang1345/delvewheel/issues/38.

Should fix the errors in https://github.com/dwavesystems/dwave-system/pull/518 once it's deployed.


arcondello:
I manually confirmed that the patch is applied to the correct `__init__.py`. It's kind of hard to test in CI.

PR #36, opened 2024-06-25
Support NumPy 2.0

arcondello:


arcondello:
Oops, the latest doesn't actually install NumPy 2.0 because of the `dimod` install in `tests/requirements.txt`. So will need to rerun after https://github.com/dwavesystems/dimod/pull/1375 is deployed.

thisac (approved):
LGTM (assuming that the tests pass).

PR #37, opened 2024-06-26
Fix `DisjointBitSetsNode` state copying

arcondello:
I'll think about ways to enforce that all decision nodes have copyable states in the future. But for now let's fix the current one.


alexzucca90 (approved):
LGTM

PR #42, opened 2024-07-03
Refactor dwave.optimization.generators

arcondello:
* Clean up pep8 errors and whitespace
* Remove redundant tests in tests/test_examples.py
* Refactor input argument checking/coercion
* Slightly generalize some of the generators


dwave/optimization/generators.py
```diff
@@ -17,120 +17,64 @@
...
-    for the error message to make sense.
-    """
-    for argname, array in kwargs.items():
-        try:
-            array = np.atleast_2d(np.asarray(array))
-        except (ValueError, TypeError):
-            raise ValueError(f"`{argname}` must be a square 2D array-like of non-negative numbers")
+        # we do "size up"
+        while array.ndim < ndim:
+            array = array[np.newaxis]
```
JoelPasvolsky, line 66: I'm not quite sure what is intended with this clever trick but I doubt it is this:

```
model = job_shop_scheduling(times=[1, 2, 3, 4], machines=[4, 3, 2, 1])
ValueError: arrays must have the same shape
```

I think the code will be clearer if, for a requirement on square matrices, you check that `array.shape[0] == array.shape[1]` and don't upsize.

JoelPasvolsky (commented):
That's a lot of trailing spaces you cleaned up.

dwave/optimization/generators.py
arcondello: That error was present in the original code. Turns out it fails on `main` for every non-square times/machines. Will fix.

Re the dimension increase, I am following the pattern of [np.atleast_2d()](https://numpy.org/doc/stable/reference/generated/numpy.atleast_2d.html), which is typical in this sort of place.

arcondello: Ah, that's a different bug. Will fix that one too.

arcondello: https://github.com/dwavesystems/dwave-optimization/issues/43

JoelPasvolsky (approved):

Issue #23, opened 2024-06-12
Inconsistency for returned value type of `maybe_equals()`


arcondello:
Originally reported by @JoelPasvolsky 

Depending on whether it's tested on another symbol or on itself, `maybe_equals()` returns Boolean or integer:
```
>>> model = Model()
>>> i = model.integer(100, lower_bound=0, upper_bound=20)
>>> j = model.integer(3, lower_bound=-10, upper_bound=10)
>>> j.maybe_equals(j)
True
>>> j.maybe_equals(i)
0
```

arcondello:
This should always return a number. Specifically `0` if not equal, `1` if possibly equal, and `2` if definitely equal.

PR #49, opened 2024-07-10
Make `maybe_equals` consistently return integers

kevinchern:
Closes https://github.com/dwavesystems/dwave-optimization/issues/23

### DIFF
--- a/dwave/optimization/model.pyx
+++ b/dwave/optimization/model.pyx
@@ -1359,9 +1359,6 @@ cdef class Symbol:
 
     def maybe_equals(self, other):
         """Compare to another node. This method exists because a complete equality test can be expensive.
-        A return value of ``0`` guarantees the two nodes are not equal.
-        A return value of ``1`` is indecisive and a more expensive, complete equality test is necessary.
-        A return value of ``2`` guarantees the two nodes are equal.
 
         Args:
             other: Another node in the model's directed acyclic graph.
@@ -1369,9 +1366,24 @@ cdef class Symbol:
         Returns: integer
             Supported return values are:
 
-            *   ``0``---Not equal.
-            *   ``1``---Might be equal.
-            *   ``2``---Are equal.
+            *   ``0``---Not equal (with certainty)
+            *   ``1``---Might be equal (no guarantees); a complete equality test is necessary
+            *   ``2``---Are equal (with certainty)
+
+        Examples:
+            This example compares
+            :class:`~dwave.optimization.symbols.IntegerVariable` symbols
+            of different sizes.
+
+            >>> from dwave.optimization import Model
+            >>> model = Model()
+            >>> i = model.integer(3, lower_bound=0, upper_bound=20)
+            >>> j = model.integer(3, lower_bound=-10, upper_bound=10)
+            >>> k = model.integer(5, upper_bound=55)
+            >>> i.maybe_equals(j)
+            1
+            >>> i.maybe_equals(k)
+            0
         """
         cdef Py_ssize_t NOT = 0
         cdef Py_ssize_t MAYBE = 1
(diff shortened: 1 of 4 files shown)


kevinchern:
lol might have to view this by commit 😆 removed trailing white spaces in the first commit

kevinchern:
does this need a release note?

dwave/optimization/model.pyx
```diff
@@ -1358,7 +1358,10 @@ cdef class Symbol:
             inc(it)
 
     def maybe_equals(self, other):
-        """Compare to another node.
+        """Compare to another node. This method exists because a complete equality test can be expensive.
+        A return value of ``0`` guarantees the two nodes are not equal.
```
arcondello, line 1362: I think we want this under `Returns:`.

I see there are two docstring https://github.com/dwavesystems/dwave-optimization/blob/c36f47b8434251dd12c390827c811d788f9f00b8/dwave/optimization/model.pyx#L1360-L1372 and https://github.com/dwavesystems/dwave-optimization/blob/c36f47b8434251dd12c390827c811d788f9f00b8/dwave/optimization/model.pyx#L1661-L1683

I _think_ if we just delete the docstring for `ArraySymbol.maybe_equals()` it will be inherited from `Symbol.maybe_equals()`.

And we should then copy over
```
        Returns: integer
            Supported return values are:
            
            *   ``0``---Not equal.
            *   ``1``---Might be equal.
            *   ``2``---Are equal.         
```
keeping the best of both worlds.

arcondello:
> does this need a release note?

Yes. Something like `Fix return type of ``Symbol.maybe_equals()`` `.

arcondello:
Also, can you rebase off of https://github.com/dwavesystems/dwave-optimization/pull/51 ? Will make it easier to review in isolation.

dwave/optimization/model.pyx
```diff
@@ -1358,17 +1358,32 @@ cdef class Symbol:
             inc(it)
 
     def maybe_equals(self, other):
-        """Compare to another node.
+        """Compare to another node. This method exists because a complete equality test can be expensive.
```
arcondello, line 1361:
```suggestion
        """Compare to another node.
        
        This method exists because a complete equality test can be expensive.
```
IMO this makes sense to include in the body rather than in the summary line. See [PEP 257](https://peps.python.org/pep-0257/#multi-line-docstrings).

dwave/optimization/model.pyx
```diff
@@ -1658,35 +1673,13 @@ cdef class ArraySymbol(Symbol):
         return Max(self)
 
     def maybe_equals(self, other):
```
arcondello, line 1675:
```suggestion
    def maybe_equals(self, other):
        # note: docstring inherited from Symbol.maybe_equal()
```

arcondello (commented):
Two small changes, otherwise looking good!

arcondello (approved):
Thanks!

tests/test_symbols.py
```diff
@@ -56,17 +56,19 @@ def assertLessEqual(self, *args, **kwargs): ...
...
     def test_equality(self):
+        DEFINITELY = 2
         for x in self.generate_symbols():
-            self.assertTrue(x.maybe_equals(x))
+            self.assertEqual(DEFINITELY, x.maybe_equals(x))
             self.assertTrue(x.equals(x))
 
         for x, y in zip(self.generate_symbols(), self.generate_symbols()):
-            self.assertTrue(x.maybe_equals(y))
+            self.assertTrue(DEFINITELY, x.maybe_equals(y))
```
k8culver, line 65: I know this is already merged but was this supposed to have been updated from `assertTrue`?

arcondello: ah, good catch. I can fix it in a followup.

Issue #46, opened 2024-07-08
Consider adding `Stack` symbol or similar


arcondello:
Something like https://numpy.org/doc/stable/reference/generated/numpy.stack.html

arcondello:
Upon further review, what we actually want is an [np.concatenate](https://numpy.org/doc/2.0/reference/generated/numpy.concatenate.html) equivalent. I believe `stack(..)` is just `concatenate(...)` followed by a `reshape(...)`.

Issue #38, opened 2024-06-27
Consider adding `as_symbol()` method or similar


arcondello:
Similar to `numpy.asarray`. This would basically be
```python
def as_symbol(model, symbol):
    if isinstance(symbol, Symbol):
        return symbol
    return model.constant(np.asarray(symbol, dtype=double))
```
or something like that. We could then use it in a bunch of function signatures. Because right now it's clunky to have to do `model.constant(...)` absolutely everywhere.

PR #50, opened 2024-07-11
Add `Size` symbol

### LINKED ISSUE #48: Consider adding a Counter Node for counting the number of elements in a disjoint_list
Assuming we have the following variable `model.disjoint_lists(primary_set_size, num_disjoint_lists)`, it would be great to have a Counter Node that counts the number of elements in each `disjoint_list`.

arcondello:
Closes #48


arcondello:
I am toying with renaming these to `Length` and `LengthNode`. Need to think about it but wanted to get the logic up to be tested with.

arcondello:
Still ruminating on this, but a few thoughts.

One issue with calling it `Len` or `Length` is that the `len()` of a NumPy array is the size of the first dimension. I am worried that would be confusing.

I am currently inclined to do something like
```python
s: Size = a.size(symbol=True)
```
It's ugly, but nicely explicit.

*Alternatives*
* We could make  `ArraySymbol.size()` always return a symbol, but there are many legitimate uses that want an integer.
* If we had https://github.com/dwavesystems/dwave-optimization/issues/38 we could consider having `ArraySymbol.size()` return a symbol when it's dynamic and an int otherwise. But that's pretty ugly switching behavior.

arcondello:
Ah, if we implement [__index__()](https://docs.python.org/3/reference/datamodel.html#object.__index__) for  `Constant`  then returning a constant always from `size()` will work in most places. Interesting.

arcondello:
With the addition of https://github.com/dwavesystems/dwave-optimization/pull/72, we could now always return a symbol from `.size()`. Though another downside of that approach is that without some sort of constant caching or de-duplication scheme we might end up with many constants in the model. So returning a symbol only when dynamic might still be preferable.

arcondello:
Ended up going with the approach where we return a symbol only for dynamic nodes. I hate the switching behavior of the return type, but I think it's the least confusion option for now.

tests/test_symbols.py
```diff
@@ -1788,6 +1787,43 @@ def test_state_size(self):
...
+        a = dwave.optimization.symbols.Size(model.constant(5))
+        b = dwave.optimization.symbols.Size(model.constant([0, 1, 2]))
+        c = model.set(5).size()
+
+        with model.lock():
+            yield a
+            yield b
+            yield c
+
+    def test_dynamic(self):
```
elham-azd, line 1803: I'm trying to get the length for disjoint_lists:
```
model = Model()
model.states.resize(1)

x, ys = model.disjoint_lists(5, 3)
length0 = ys[0].size()
length1 = ys[1].size()
length2 = ys[2].size()

x.set_state(0, [[0, 1, 2, 3, 4], [], []])
with model.lock():
    print(length0.state(0))
    print(length1.state(0))
    print(length2.state(0))
```
Is there a better way for doing this? For example, can we make an array of all lengths?

arcondello: Currently no way to join them into another array unfortunately, we'd need #46 for that I think

elham-azd (approved):

PR #53, opened 2024-07-11
Update the default repr() of symbols

### LINKED ISSUE #52: Consider changing `Symbol.__repr__()` to report underlying node ID rather than python symbol ID
Right now if you have two different symbols pointing to the same underlying C++ node, you get different `repr`s. E.g.,
```python
In [4]: from dwave.optimization import Model

In [5]: model = Model()

In [6]: c0 = model.constant(5)

In [7]: c1, = model.iter_symbols()

In [8]: c0
Out[8]: <dwave.optimization.symbols.Constant at 0x7f5d22e28400>

In [9]: c1
Out[9]: <dwave.optimization.symbols.Constant at 0x7f5d203da3e0>
```
Both of these symbols are referencing the same underlying C++ node, but the `repr` does not reflect that.

We should consider replacing the `0x...` with one pointing to the underlying node pointer.

The cost of course is losing information about the Python object uniqueness, as opposed to the quantity represented uniqueness.

arcondello:
Closes #52

### DIFF
--- a/dwave/optimization/model.pyx
+++ b/dwave/optimization/model.pyx
@@ -40,6 +40,7 @@ import numpy as np
 from cpython cimport Py_buffer
 from cython.operator cimport dereference as deref, preincrement as inc
 from cython.operator cimport typeid
+from libc.stdint cimport uintptr_t
 from libcpp cimport bool
 from libcpp.utility cimport move
 from libcpp.vector cimport vector
@@ -1192,6 +1193,20 @@ cdef class Symbol:
         # disallow direct construction of symbols, they should be constructed
         # via their subclasses.
         raise ValueError("Symbols cannot be constructed directly")
+
+    def __repr__(self):
+        """Return a representation of the symbol.
+
+        The representation refers to the id of the underlying node, rather than
+        the id of the Python symbol.
+        """
+        cls = type(self)
+        # We refer to the node_ptr, which is not necessarily the address of the
+        # C++ node, as it sublasses Node.
+        # But this is unique to each node, and functions as an id rather than
+        # as a pointer, so that's OK.
+        # Otherwise we aim to match Python's default __repr__.
+        return f"<{cls.__module__}.{cls.__qualname__} at {<uintptr_t>self.node_ptr:#x}>"
 
     cdef void initialize_node(self, Model model, cppNode* node_ptr) noexcept:
         self.model = model
--- a/releasenotes/notes/fix-to_networkx-a78f0f669cc9638c.yaml
+++ b/releasenotes/notes/fix-to_networkx-a78f0f669cc9638c.yaml
@@ -1,0 +1,5 @@
+---
+features:
+  - |
+    Make ``repr()`` of symbols unique to the underlying node, rather than to the Python symbol.
+    See `#52 <https://github.com/dwavesystems/dwave-optimization/issues/52>_.
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -424,3 +424,14 @@ class TestSymbol(unittest.TestCase):
         from dwave.optimization.model import Symbol
         with self.assertRaisesRegex(ValueError, "Symbols cannot be constructed directly"):
             Symbol()
+
+    def test_repr(self):
+        model = Model()
+        c0 = model.constant(5)
+        c1, = model.iter_symbols()
+        c2 = model.constant(5)
+
+        # the specific form is an implementation detail, but different symbols
+        # representing the same underlying node should have the same repr
+        self.assertEqual(repr(c0), repr(c1))
+        self.assertNotEqual(repr(c0), repr(c2))


arcondello:
The image in the docs looks like 
image

dwave/optimization/model.pyx
```diff
@@ -845,53 +846,72 @@ cdef class Model:
...
-            # this regex is compiler specific! Don't do this for the general case
-            match = re.search("\d+([a-zA-z]+Node)", str(typeid(deref(ptr)).name()))
-            if not match:
-                raise ValueError
+            This creates the following image:
 
-            u = (match[1], <long>(ptr))
+            .. figure:: /_images/to_networkx_example.svg
+               :width: 500 px
+               :name: to-networkx-example
```
JoelPasvolsky, line 880: Would you mind renaming to something like `nlModelToNetworkxExample` for the figure's label and something like  `nl_model_to_networkx_example` for the filename above? Once the package is updated in the SDK, having package-specific names is helpful for preventing label conflicts.

dwave/optimization/model.pyx
```diff
@@ -845,53 +846,72 @@ cdef class Model:
...
-                if not match:
-                    raise ValueError
+        # Add the symbols, in order if we happen to be topologically sorted
+        G.add_nodes_from(repr(symbol) for symbol in self.iter_symbols())
 
-                v = (match[1], <long>(pptr))
+        # Sanity check. If several nodes map to the same symbol repr we'll see
+        # too few nodes in the graph
+        if len(G) != self.num_symbols():
+            raise RuntimeError("symbol repr() is not unique to the underlying node")
```
JoelPasvolsky, line 894: Why not do this once only in line 903 (or even 915)?

dwave/optimization/model.pyx
```diff
@@ -845,53 +846,72 @@ cdef class Model:
...
 
-                G.add_edge(v, u)
+        # Now add the edges
+        for symbol in self.iter_symbols():
+            for successor in symbol.iter_successors():
+                G.add_edge(repr(symbol), repr(successor))
+
+        # Sanity check again. If the repr of symbols isn't unique to the underlying
+        # node then we'll see too many nodes in the graph here
+        if len(G) != self.num_symbols():
```
JoelPasvolsky, line 903: This should not happen, right? Do you want to add some debugging info on the offending symbol for if it does and the user contacts Support?

dwave/optimization/model.pyx
```diff
@@ -1193,6 +1213,20 @@ cdef class Symbol:
         # via their subclasses.
         raise ValueError("Symbols cannot be constructed directly")
 
+    def __repr__(self):
+        """Return a representation of the symbol.
+
+        The representation refers to the id of the underlying node, rather than
+        the id of the Python symbol.
```
JoelPasvolsky, line 1220:
```suggestion
        The representation refers to the identity of the underlying node, rather than
        the identity of the Python symbol.
```

JoelPasvolsky (commented):
Looks good, thanks!

dwave/optimization/model.pyx
```diff
@@ -845,53 +846,72 @@ cdef class Model:
...
 
-        import re
-        import networkx
+            One advantage of converting to NetworkX is the wide availability
+            of drawing tools. See NetworkX's
+            `drawing <https://networkx.org/documentation/stable/reference/drawing.html>`_
+            documentation.
 
-        G = networkx.DiGraph()
+            This example uses a `DAGVIZ <https://wimyedema.github.io/dagviz/>`_ to
```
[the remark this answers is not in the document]
arcondello: We are not including `dagviz` as a requirement anywhere, partly for this reason. This is just an example of a library one can use to visualize with.

dwave/optimization/model.pyx
arcondello: These are for debugging and the location will be helpful to that end.

dwave/optimization/model.pyx
arcondello: This is for maintainers, `RuntimeError` is not user facing unless something has gone terribly wrong.

dwave/optimization/model.pyx
arcondello: How about
```suggestion
               :name: dwave-optimization-to-networkx-example
```

dwave/optimization/model.pyx
thisac, line 894: Is this only for debugging? It should be enough to just add the edges and implicitly add the nodes below.

dwave/optimization/model.pyx
thisac: Can a successor ever _not_ be in `iter_symbols`? Otherwise I don't see the use for this check either.

dwave/optimization/model.pyx
```diff
@@ -845,53 +846,72 @@ cdef class Model:
...
+
+        # Add the objective if it's present. We call it "minimize" to be
+        # consistent with the minimize() function
+        if self.objective is not None:
+            G.add_edge(repr(self.objective), "minimize")
+
+        # Likewise if we have constraints, add a special node for them
+        if self.num_constraints():
+            for symbol in self.iter_constraints():
+                G.add_edge(repr(symbol), "constraint(s)")
```
thisac, line 914: Isn't `self.iter_constraints()` just empty if there are no constraints?
```suggestion
        for symbol in self.iter_constraints():
            G.add_edge(repr(symbol), "constraint(s)")
```

dwave/optimization/model.pyx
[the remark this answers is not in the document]
thisac: Nitpicky, but it sounds better to skip the "a" in "...a DAGVIZ" and just say "...uses DAGVIZ", but I guess technically the latter could be correct as well if seeing DAGVIZ as a type of visualized graph.

thisac (approved):
Some minor comments, but nothing blocking.

dwave/optimization/model.pyx
arcondello: If the reprs are not unique, then there will be too many symbols here. Because `iter_sucessors()` makes new python objects each time. So without c706ae8 they would appear as extra nodes. This is a hedge on future changes to `__repr__()`.

dwave/optimization/model.pyx
arcondello: Same comment https://github.com/dwavesystems/dwave-optimization/pull/53#discussion_r1674699140. Specifically this can go wrong if we were to change the `__repr__()` such that
```python
>>> x = model.binary(0)
>>> repr(x)
Binary(0)
>>> y = model.binary(0)
>>> repr(y)
Binary(0)
```
these are different symbols with identical reprs and that would cause this runtime error.

dwave/optimization/model.pyx
thisac: Ah, yeah, that makes sense.

dwave/optimization/model.pyx
JoelPasvolsky: works for me, thanks!

dwave/optimization/model.pyx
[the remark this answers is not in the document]
arcondello:
```suggestion
            This example uses `DAGVIZ <https://wimyedema.github.io/dagviz/>`_ to
```

Issue #54, opened 2024-07-12
Docs reference incorrect version


arcondello:
We include `|version|` in `index.rst` https://github.com/dwavesystems/dwave-optimization/blob/c711e35f41b6888fdcc489a9849167b0ec217be6/docs/index.rst?plain=1#L17 which leads to the SDK inserting its version number in the built docs 
![image](https://github.com/user-attachments/assets/197c5db5-e450-4f75-a98e-074e3cac5885)

See also https://github.com/dwavesystems/dwave-ocean-sdk/issues/150

PR #55, opened 2024-07-12
Improve doc testing

arcondello:
Match the SDK's doc theme and add doctests to CI.

Closes https://github.com/dwavesystems/dwave-optimization/issues/54
Also fixes some doc errors introduced in #42 and #53

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -85,31 +85,15 @@ import dwave.optimization
 
 """
 
-# -- Breath ---------------------------------------------------------------
-
-# breathe_default_project = "dwave.optimization"
-# breathe_projects = dict(
-#   optimization=os.path.join(config_directory, 'build-cpp', 'xml'),
-#   )
-
-# see https://breathe.readthedocs.io/en/latest/readthedocs.html
-# if os.environ.get('READTHEDOCS', False):
-#     subprocess.call('make cpp', shell=True, cwd=config_directory)
-
 # -- Options for HTML output ----------------------------------------------
 
-# The theme to use for HTML and HTML Help pages.  See the documentation for
-# a list of builtin themes.
-#
-import sphinx_rtd_theme
-html_theme = 'sphinx_rtd_theme'
-html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
+html_theme = "pydata_sphinx_theme"
 
-# Theme options are theme-specific and customize the look and feel of a theme
-# further.  For a list of options available for each theme, see the
-# documentation.
-#
-# html_theme_options = {}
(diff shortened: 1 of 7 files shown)


docs/conf.py
```diff
@@ -85,31 +77,15 @@
```
JoelPasvolsky, line 77: Line 51:
```
language = 'en'
```

Let's start getting rid of `WARNING: Invalid configuration value found: 'language = None'. Update your configuration to a valid language code. Falling back to 'en' (English).`

docs/intro.rst
```diff
@@ -164,8 +164,7 @@ for symbols of a model without using labels.
 >>> with model.lock():
 ...     for symbol in model.iter_decisions():
 ...         symbol.set_state(0, [2])
-...     model.objective.state(0) == -4
-True
+...     assert model.objective.state(0) == -4
```
JoelPasvolsky, line 167: I'm not sure about this: it save one line but at the cost of slightly less understandable code for less experienced users and the replacement of a False output were it to not obtain by an error state, which is not desirable

JoelPasvolsky (approved):
LGTM, thanks!

docs/intro.rst
arcondello: The problem is that `NumPy` either returns `True` or `np.True_` depending on the version. So doctests may fail in either case. I believe `assert` is fairly clear even to novice Python users.

JoelPasvolsky: Or could print the value

arcondello: Is that really clearer than `assert`? Both are built-in Python operations. If we cannot assume that they know about `assert`, then can we assume they know `print`?

JoelPasvolsky: Yes, they all know print

JoelPasvolsky: Originally we moved from `print()` statements to using equals to get a True because we worried about changes to the format of the output over time (mostly unordered lists and sets) but here that's not a concern

arcondello: I am comfortable assuming that they know `assert` as well.

arcondello: > changes to the format of the output over time (mostly unordered lists and sets) but here that's not a concern
Is it not? `NumPy` could decide to change the `str()` representation of `np.True_`.

JoelPasvolsky: Sorry, I meant print the value: `print(model.objective.state(0))` and showing its output (that will be used by the doctest)

arcondello: That will return `np.float64(4)` or `4` depending on the NumPy version

PR #56, opened 2024-07-12
Add model attributes objective states to docs

JoelPasvolsky:


arcondello (approved):
LGTM though I'd like to rebase after https://github.com/dwavesystems/dwave-optimization/pull/55 is merged so we can run the docstests.

PR #57, opened 2024-07-12
Flatten symbols documentation

arcondello:


docs/reference/symbols.rst
```diff
@@ -84,21 +56,19 @@ are inherited by the :ref:`model symbols <symbols_model_symbols>`.
...
-are inherited by the :class:`~dwave.optimization.model.ArraySymbol`
-class and :ref:`model symbols <symbols_model_symbols>`.
+Model Symbols
+=============
 
-.. autosummary::
-    :toctree: generated/
+Each operation, decision, constant, mathematical function, and
+flow control is modeled using a symbol. The following is the list
+of symbols available for modelling.
```
JoelPasvolsky, line 66:
```suggestion
flow control is modeled using a symbol. The following symbols
are available for modelling.
```

no longer has a list

docs/reference/symbols.rst
```diff
@@ -7,59 +7,31 @@ Symbols
 Symbols are a model's decision variables, intermediate variables, constants, 
 and mathematical operations.
 
-.. _symbols_model_symbols:
+See :ref:`Symbols <intro_optimization_symbols>` for an introduction to
+working with symbols.
 
-Model Symbols
-=============
+.. currentmodule:: dwave.optimization.model
```
JoelPasvolsky, line 13:
```suggestion
Inherited Methods
============

.. currentmodule:: dwave.optimization.model
```

Currently the side navigation won't be helpful to new users:

image

JoelPasvolsky (approved):
I like this. 

I think it would benefit from more introductory text to explain the relationships between high-level objects to give an overview before diving in but that can be added in a separate PR

arcondello:
> I think it would benefit from more introductory text to explain the relationships between high-level objects to give an overview before diving in but that can be added in a separate PR

Totally agree

docs/reference/symbols.rst
arcondello: I don't love `Inherited`. What about `Base Symbols`?

JoelPasvolsky: Could that be "Base Classes" instead? "Base symbols" might be mistaken as a different flavor of symbols (talk about saturation of a word)

arcondello:
```suggestion
Base Classes
============

.. currentmodule:: dwave.optimization.model
```

PR #61, opened 2024-07-15
Construct C++ `NaryOpNode` from `Array*`

arcondello:


dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -269,12 +269,28 @@ NaryOpNode<BinaryOp>::NaryOpNode(Node* node_ptr)
...
+    for (Node* ptr : node_ptrs) {
+        add_node(ptr);
+    }
+}
+
+template <class BinaryOp>
+NaryOpNode<BinaryOp>::NaryOpNode(std::span<Array*> array_ptrs)
+        : Node(), ArrayOutputMixin(nonempty(array_ptrs)->shape()) {
+    for (Array* ptr : array_ptrs) {
+        add_node(dynamic_cast<Node*>(ptr));
```
alexzucca90, line 293: Should we add an assertion that the casting doesn't return `nullptr`?

arcondello: We should. Also, turns out `add_node()` casts to `Array*` .... Will clean that up too.

arcondello: Ok, added some checks. We still do two dynamic casts, but IMO the right fix is deeper. So we can/should live with it for now.

alexzucca90 (approved):
lgtm

PR #63, opened 2024-07-16
Restructure symbol tests

arcondello:
Remove the symbol test mixins and replace them with subclasses. This requires putting them in another "namespace". I also added `BinaryOpTests` with a single sanity test.

### DIFF
--- a/tests/test_symbols.py
+++ b/tests/test_symbols.py
@@ -25,225 +25,224 @@ import dwave.optimization.symbols
 from dwave.optimization import Model, logical_or, logical_and
 
 
-class SymbolTestsMixin(abc.ABC):
-    @abc.abstractmethod
-    def generate_symbols(self):
-        """Yield symbol(s) for testing.
+class utils:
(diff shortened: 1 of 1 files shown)


tests/test_symbols.py
```diff
@@ -25,225 +25,224 @@
...
-            model = x.model
-            index = x.topological_index()
+            model = Model()
+            lhs_symbol = model.constant(lhs_array)
+            rhs_symbol = model.constant(rhs_array)
 
-            # Get the symbol back
-            y, = itertools.islice(model.iter_symbols(), index, index+1)
+            op_array = self.op(lhs_array, rhs_array)
+            op_symbol = self.symbol_op(lhs_symbol, rhs_symbol)
```
wbernoudy, line 134: Should there be another line here that actually tests the output is correct?

arcondello: :laughing: yes.

arcondello: Added in 369dc04

wbernoudy (approved):

PR #68, opened 2024-07-18
Improve operator overloading for array symbols

arcondello:
* Rather than raising a `TypeError` when an operator sees an unknown type, we instead return `NotImplemented`.
* Implement `*=` and `+=` for array symbols.


arcondello:
This PR is concerned with `ArraySymbols` in general. I will make a followup PR for `Constant` specifically.

wbernoudy (approved):
LGTM, appreciate the niceness of creating `Nary` nodes with `+=` and `*=`!

PR #72, opened 2024-07-25
Support `<,>,==,!=,<=,>=,int,bool,not` operations on Constant

arcondello:
This allows us to treat `Constant` as a number in many places. And is a step towards `ArraySymbol.size() -> Constant`, see https://github.com/dwavesystems/dwave-optimization/pull/50#issuecomment-2229200436


dwave/optimization/symbols.pyx
```diff
@@ -781,6 +783,12 @@ cdef class Constant(ArraySymbol):
         # Have the parent model hold a reference to the array, so it's kept alive
         model._data_sources.append(array)
 
+    def __bool__(self):
+        if not self._is_scalar():
+            raise ValueError("the truth value of a constant with more than one element is ambiguous")
+
+        return <bool>deref(self.ptr.buff())
```
thisac, line 790: Does the `buff()` method point at the first (only in this case) location in memory? What is the difference between what's returned by `self.ptr` (I assume this is a pythonic pointer object and not actually a memory address pointer?) and `self.ptr.buff()`?

dwave/optimization/symbols.pyx
```diff
@@ -794,6 +802,54 @@ cdef class Constant(ArraySymbol):
...
+            return NotImplemented  # this should never happen, but just in case
+
+    cdef bool _is_integer(self) noexcept:
+        """Return True if the constant encodes a single integer."""
+        if not self._is_scalar():
+            return False
+
+        # https://stackoverflow.com/q/1521607 for the integer test
+        cdef double dummy
+        return modf(deref(self.ptr.buff()), &dummy) == <double>0.0
```
thisac, line 846: Interesting solution.

dwave/optimization/symbols.pyx
```diff
@@ -794,6 +802,54 @@ cdef class Constant(ArraySymbol):
...
+
+    def __richcmp__(self, rhs, int op):
+        # __richcmp__ is a special Cython method
+
+        # If rhs is another Symbol, defer to ArraySymbol to handle the
+        # operation. Which may or may not actually be implemented.
+        # Otherwise, defer to NumPy.
+        # We could also check if rhs is another Constant and handle that differently,
+        # but that might lead to confusing behavior so we treat other Constants the
+        # same as any other symbol.
```
thisac, line 821: In what was should/could it be handled differently (unless not deferring to NumPy at all)? NumPy would just use `rhs` as if a NumPy array either way since it's being deferred there.

dwave/optimization/symbols.pyx
```diff
@@ -794,6 +802,54 @@ cdef class Constant(ArraySymbol):
...
+        elif op == cpython.object.Py_GT:
+            return lhs.__gt__(rhs)
+        elif op == cpython.object.Py_LE:
+            return lhs.__le__(rhs)
+        elif op == cpython.object.Py_LT:
+            return lhs.__lt__(rhs)
+        elif op == cpython.object.Py_NE:
+            return lhs.__ne__(rhs)
+        else:
+            return NotImplemented  # this should never happen, but just in case
```
thisac, line 837:
```suggestion
        return NotImplemented  # this should never happen, but just in case
```

tests/test_symbols.py
```diff
@@ -595,6 +596,60 @@ def generate_symbols(self):
...
+        operators = [
+            operator.eq,
+            operator.ge,
+            operator.gt,
+            operator.le,
+            operator.lt,
+            operator.ne
+        ]
+
+        for op in operators:
```
thisac, line 637: (minor) It would've been nice to have this parameterized and test each operator separately.

thisac (approved):
LGTM, just a couple of minor questions.

dwave/optimization/symbols.pyx
arcondello: I think the nesting makes it clearer that it's because those are the only options.

dwave/optimization/symbols.pyx
thisac: > We could also check if rhs is another Constant and handle that differently

Just responding to this comment. Curious if there's anything that should or could be handled differently.

tests/test_symbols.py
thisac: Oh, true. I somehow missed that line, and running the test using pytest only shows this as a single test.

thisac: [pytest-subtests](https://github.com/pytest-dev/pytest-subtests) required to make pytest work with `subTest`. Good to know!

dwave/optimization/symbols.pyx
arcondello: Oh. The alternative would be, when both sides of the operation are `Constant`s, to have the result be another `Constant` or even to "demote" to a NumPy array.

PR #73, opened 2024-07-25
Flatten generator docs

arcondello:
The main benefit here is that new generators are automatically incorporated into the docs. For example, we were already missing `quadratic_assignment`.

The cons are slightly ugly formatting at the top level 
image

and no navigation within the page on the sidebar 
image

IMO the maintenance benefit is worth it. If we wanted a summary table at the top, there are sphinx plugins, e.g. https://github.com/Chilipp/autodocsumm.

Another change made is that by using `autodoc_type_aliases` we can significantly shrink the size and increase the readability of the signature. IMO we should make that change in the SDK.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -77,6 +77,10 @@ import dwave.optimization
 
 """
 
+autodoc_type_aliases = {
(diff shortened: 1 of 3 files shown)


JoelPasvolsky (approved):
Yeah, this is nice

PR #81, opened 2024-08-08
Add Symbol.id() method

arcondello:
This is useful for a few other features I am working on.


dwave/optimization/model.pyx
```diff
@@ -1328,6 +1326,41 @@ cdef class Symbol:
...
+
+        See Also:
+            :meth:`.shares_memory`: ``a.shares_memory(b)`` is equivalent to ``a.id() == b.id()``.
+            
+            :meth:`.equals`: ``a.equals(b)`` will return ``True`` if ``a.id() == b.id()``. Though
+            the inverse is not necessarily true.
+
+        """
+        # We refer to the node_ptr, which is not necessarily the address of the
+        # C++ node, as it sublasses Node.
```
wbernoudy, line 1359: Typo

dwave/optimization/model.pyx
```diff
@@ -1328,6 +1326,41 @@ cdef class Symbol:
         # We need to explicitly cast to evoke unique_ptr's operator bool
         return <bool>(self.model.states._states[index][self.node_ptr.topological_index()])
 
+    cpdef uintptr_t id(self) noexcept:
+        """Return the "identity" of the underlying node.
+
+        This identity is unique to the underlying node, rather than the identity
+        of the Python object representing it.
+        Therefore, ``symdol.id()`` is not the same as ``id(symbol)``!
```
wbernoudy, line 1334: Is there a reason not have `Symbol.__hash__()` return `id()`?

wbernoudy (approved):

dwave/optimization/model.pyx
arcondello: The reason we don't just return this number from `__hash__()` is because we don't want users inserting symbols into dictionaries/sets directly. The way Python handles that is first calling `__hash__()` and then `__eq__()`. The former would work fine, but the latter would return another symbol, and one with an ambiguous truth value. If the `__eq__()` raises an error, it gets treated as a False. The upshot is that size 1 symbols that return constants would work fine with set/dict, but nothing else would.
Much better, IMO, to explicitly raise an error from `__hash__()` and provide another method like `.id()` to serve a similar purpose, but which needs to be invoked explicitly.

PR #83, opened 2024-08-09
Implement `Logical` and `Not` symbols

arcondello:
There isn't really an analogue to `Logical` in NumPy, it does either `np.asarray(x, dtype=bool)` or `x != 0`. I chose `logical(x)` to follow NumPy's `logical_and(array)`, `logical_not(array)`, etc convention. Open to other names.

I also implemented some missing `UnaryOpNode` methods. Probably there are similar things we need to do for `NaryOpNode` and `BinaryOpNode`.

### DIFF
--- a/dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
+++ b/dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
@@ -31,6 +31,11 @@ struct abs {
 };
 
 template <class T>
+struct logical {
+    constexpr bool operator()(const T& x) const { return x; }
+};
+
+template <class T>
 struct max {
     constexpr T operator()(const T& x, const T& y) const { return std::max(x, y); }
 };
@@ -167,6 +172,9 @@ class UnaryOpNode : public ArrayOutputMixin<ArrayNode> {
 
     double const* buff(const State& state) const override;
     std::span<const Update> diff(const State& state) const override;
+    bool integral() const override;
+    double max() const override;
+    double min() const override;
 
     void commit(State& state) const override;
     void revert(State& state) const override;
@@ -183,7 +191,9 @@ class UnaryOpNode : public ArrayOutputMixin<ArrayNode> {
 };
 
 using AbsoluteNode = UnaryOpNode<functional::abs<double>>;
+using LogicalNode = UnaryOpNode<functional::logical<double>>;
 using NegativeNode = UnaryOpNode<std::negate<double>>;
+using NotNode = UnaryOpNode<std::logical_not<double>>;
 using SquareNode = UnaryOpNode<functional::square<double>>;
 
 }  // namespace dwave::optimization
--- a/dwave/optimization/mathematical.py
+++ b/dwave/optimization/mathematical.py
@@ -155,6 +155,9 @@ def logical_and(x1: ArraySymbol, x2: ArraySymbol) -> And:
         ...     y.set_state(0, [False, True, False])
         ...     print(z.state(0))
         [0. 1. 0.]
+
+    See Also:
+        :class:`~dwave.optimization.symbols.And`: equivalent symbol.
     """
     return And(x1, x2)
 
@@ -213,6 +216,9 @@ def logical_or(x1: ArraySymbol, x2: ArraySymbol) -> Or:
         ...     y.set_state(0, [False, True, False])
         ...     print(z.state(0))
         [1. 1. 0.]
+
+    See Also:
+        :class:`~dwave.optimization.symbols.Or`: equivalent symbol.
     """
     return Or(x1, x2)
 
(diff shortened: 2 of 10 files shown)


dwave/optimization/mathematical.py
```diff
@@ -97,6 +101,35 @@ def add(x1: ArraySymbol, x2: ArraySymbol, *xi: ArraySymbol) -> typing.Union[Add,
     raise RuntimeError("implementated by the op() decorator")
 
 
+def logical(x: ArraySymbol) -> Logical:
+    r"""Return the element-wise truth value on the given symbol.
```
wbernoudy, line 105: Maybe worth reiterating here that all values but 0 are truthy?

dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
```diff
@@ -30,6 +30,11 @@ struct abs {
     constexpr T operator()(const T& x) const { return std::abs(x); }
 };
 
+template <class T>
+struct logical {
+    constexpr bool operator()(const T& x) const { return x; }
```
wbernoudy, line 35: Seems slightly confusing to me that this returns a `bool` instead of `T`. I guess we're relying on the implementation of the `UnaryOpNode` to cast this back to the data type for arrays, so it should be fine. But why not return type `T` here, to be sure some use of `auto` deep in the `UnaryOpNode` doesn't go awry?

wbernoudy: I guess this is a "problem" for `std::logical_not` as well. Obviously not a big deal, just thought I would note it.

tests/cpp/tests/test_nodes_mathematical.cpp
```diff
@@ -961,4 +962,106 @@ TEMPLATE_TEST_CASE("UnaryOpNode", "", functional::abs<double>, std::negate<doubl
...
+
+            THEN("Negative has the min/max we expect") {
+                CHECK(i_ptr->min() == -3);
+                CHECK(i_ptr->max() == 8);
+                CHECK(ni_ptr->min() == -8);
+                CHECK(ni_ptr->max() == 3);
+            }
+
+            THEN("Negative is also integral") { CHECK(ni_ptr->integral()); }
+        }
```
wbernoudy, line 1028: Why no state check for `NegativeNode`?

tests/cpp/tests/test_nodes_mathematical.cpp
```diff
@@ -961,4 +962,106 @@ TEMPLATE_TEST_CASE("UnaryOpNode", "", functional::abs<double>, std::negate<doubl
...
+            auto i_ptr = graph.emplace_node<IntegerNode>(std::vector<ssize_t>{});
+            auto square_ptr = graph.emplace_node<SquareNode>(i_ptr);
+
+            THEN("The min/max are expected") {
+                CHECK(square_ptr->min() == 0);
+                // we might consider capping this differently for integer types in the future
+                CHECK(square_ptr->max() ==
+                      static_cast<std::size_t>(2000000000) * static_cast<std::size_t>(2000000000));
+            }
+        }
```
wbernoudy, line 1063: No state check?

wbernoudy (approved):

dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
arcondello: We keep the bool for a few reasons. The first is consistency with `std::logical_and`, `std::logical_not`, etc. The second reason is we [use this info for `.logical()`, `.max()` and `.min()`](https://github.com/arcondello/dwave-optimization/blob/e5eab5823cef206e1caa9064ce681091a27f73cb/dwave/optimization/src/nodes/mathematical.cpp#L935). And the third is that it's the simplest way to get a truth value.

tests/cpp/tests/test_nodes_mathematical.cpp
arcondello: It's covered by the generic tests above
```c++
TEMPLATE_TEST_CASE("UnaryOpNode", "", functional::abs<double>, functional::logical<double>,
                   functional::square<double>, std::negate<double>, std::logical_not<double>)
```

tests/cpp/tests/test_nodes_mathematical.cpp
arcondello: It's covered by the generic tests above
```c++
TEMPLATE_TEST_CASE("UnaryOpNode", "", functional::abs<double>, functional::logical<double>,
                   functional::square<double>, std::negate<double>, std::logical_not<double>)
```

dwave/optimization/mathematical.py
arcondello: Sure, I can make that explicit.

PR #86, opened 2024-08-14
Add functionality for exponents that are integers greater than 0 in __pow__

jwoo098:
This change adds functionality in ArraySymbol for exponents over 2 in __pow__() by calling Multiply() multiple times.


arcondello:
I believe the CI failures are unrelated to this PR.

dwave/optimization/model.pyi
```diff
@@ -165,7 +166,7 @@ class ArraySymbol(Symbol):
     def __le__(self, rhs: ArraySymbol) -> LessEqual: ...
     def __mul__(self, rhs: ArraySymbol) -> Multiply: ...
     def __neg__(self) -> Negative: ...
-    def __pow__(self, exponent: int) -> Square: ...
+    def __pow__(self, exponent: int) -> Union[Square, Multiply]: ...
```
arcondello, line 169:
```python
    def __pow__(self, exponent: int) -> typing.Union[Square, Multiply]: ...
```

dwave/optimization/model.pyi
```diff
@@ -14,6 +14,7 @@
 
 import tempfile
 import typing
+from typing import Union
```
arcondello, line 17: Let's just use `typing.` everywhere.

dwave/optimization/model.pyx
```diff
@@ -1764,7 +1764,24 @@ cdef class ArraySymbol(Symbol):
         if exponent == 2:
             from dwave.optimization.symbols import Square  # avoid circular import
             return Square(self)
-        raise ValueError("only squaring is currently supported")
+        # check if exponent is an integer greater than 2
+        elif (isinstance(exponent, int) and exponent > 2) or (isinstance(exponent, float) and exponent.is_integer() and exponent > 2):
```
arcondello, line 1768: A bit simpler
```python
        elif isinstance(exponent, numbers.Number) and exponent > 0 and int(exponent) == exponent:
```
Also, paired with the below, we can support `exponent=1` so might as well cover that as well.

dwave/optimization/model.pyx
```diff
@@ -1764,7 +1764,24 @@ cdef class ArraySymbol(Symbol):
...
+                if nodes_exponents[-1] * 2 <= exponent:
+                    nodes.append(Square(nodes[-1]))
+                    nodes_exponents.append(nodes_exponents[-1] * 2)
+                else:
+                    indexer = 2
+                    while nodes_exponents[-1] + nodes_exponents[-indexer] > exponent:
+                        indexer += 1
+                    nodes.append(Multiply(nodes[-indexer], nodes[-1]))
+                    nodes_exponents.append(nodes_exponents[-indexer] + nodes_exponents[-1])
+            return nodes[-1]
```
arcondello, line 1783: We can take advantage of some other operators
```Python
            expanded = itertools.repeat(self, int(exponent))
            out = next(expanded)  # get the first one
            for symbol in expanded:
                out *= symbol  # multiply it by the remainder
            return out
```

arcondello (commented):
Also needs some tests. I'd suggest as a method or two in https://github.com/dwavesystems/dwave-optimization/blob/e4ee421cfc8fb17c1443e98ef7b99ec663072cde/tests/test_model.py#L29

Finally needs a release note. I just noticed the README doesn't tell you how to do it, so see [release note](https://github.com/dwavesystems/dimod?tab=readme-ov-file#release-notes). Something like "Support exponents other than 2 for array symbols.".

dwave/optimization/model.pyi
jwoo098: Done

dwave/optimization/model.pyi
jwoo098: Done

dwave/optimization/model.pyx
jwoo098: Makes sense, thanks.
I changed numbers.Number to numbers.Real to cover that edge case a little bit more cleanly.
`if isinstance(exponent, numbers.Real) and exponent > 0 and int(exponent) == exponent:`

dwave/optimization/model.pyx
jwoo098: Understood. Thank you!

jwoo098: I added this change into the commit bd3a1ae0eff222ae43c548480dc28521b23762e6. However this appears to be causing the tests to fail. I haven't been able to figure out why yet.

randomir: @jwoo098, your previous implementation might be more verbose, but it creates a lot more efficient encoding than this naive multiplication.

Something like $O(\log_2^2{n})$ `Square` + $O(\log_2{n})$ `Multiply` operations, vs $O(n)$ `Multiply` operations for the naive implementation.

If anything, I'd suggest you update it to be more efficient, rather than less. By using basic method of [exponentiation by squaring](https://en.wikipedia.org/wiki/Exponentiation_by_squaring), you should be able to get $O(\log_2{n})$ `Square` + $O(\log_2{n})$ `Multiply` ops.

randomir: And btw, the test is failing because it's [expecting `Square` symbol](https://github.com/dwavesystems/dwave-optimization/blob/0578cc0e7aba80f6255e3b77f94dcf5dd5197e26/tests/test_model.py#L75) as a result of `x**2`, and it's getting `NaryMultiply`.

jwoo098: > @jwoo098, your previous implementation might be more verbose, but it creates a lot more efficient encoding than this naive multiplication.
> 
> Something like O ( log 2 2 ⁡ n ) `Square` + O ( log 2 ⁡ n ) `Multiply` operations, vs O ( n ) `Multiply` operations for the naive implementation.
> 
> If anything, I'd suggest you update it to be more efficient, rather than less. By using basic method of [exponentiation by squaring](https://en.wikipedia.org/wiki/Exponentiation_by_squaring), you should be able to get O ( log 2 ⁡ n ) `Square` + O ( log 2 ⁡ n ) `Multiply` ops.

@randomir 
When using *= or __imul__(), it appears that I only get one additional node from __pow__ no matter how high the exponent becomes. However, when I use Multiply() (as in the verbose implementation) I get the ceiling of log2(exponent)-1 additional nodes in my more verbose implementation of __pow__ using Multiply(). Is the verbose implementation still better despite the higher number of nodes created in the model?

randomir: You're getting only one node, `NaryMultiply` that multiples the input $n-1$ times.

IMO, the best option would be to implement `Pow` symbol in C++, with the underlying optimized exponentiation.

Second best would be exponentiation by squaring (or similar method) in Python, with $O(\log_2{n})$ nodes created (these are essentially multiply nodes, square being a special case).

And probably the least efficient is $n-1$ multiplications in `NaryMultiply`. But that one has the benefit of being the simplest, most straightforward implementation.

My assumptions about runtime performance of additional $O(\log_2{n})$ (multiply) nodes vs a single $n-1$ multiply node might be wrong, though, as I haven't done such benchmarks. I.e. I'm not sure about the order of constant factors that come into play here. @arcondello should weight in.

jwoo098:
> Also needs some tests. I'd suggest as a method or two in
> 
> https://github.com/dwavesystems/dwave-optimization/blob/e4ee421cfc8fb17c1443e98ef7b99ec663072cde/tests/test_model.py#L29
> 
> Finally needs a release note. I just noticed the README doesn't tell you how to do it, so see [release note](https://github.com/dwavesystems/dimod?tab=readme-ov-file#release-notes). Something like "Support exponents other than 2 for array symbols.".

I will work on this

dwave/optimization/model.pyx
arcondello: Yes, the best option would be to implement a C++ `Pow` symbol.

In the mean time, it's generally more efficient to have fewer "large" symbols than many "smaller" symbols. So my expectation is that unless the exponent is truly large (which would lead to many other problems), that having a single node will be faster in practice.

Also, because we intend to replace with with a `Pow` symbol in the near future, having a single node and a simple implementation makes that transition easier. Users won't suddenly see a change in the number of nodes etc etc.

For those reasons, I'd like to use the `NaryMultiply` node approach as I suggested above.

randomir: In that case, the simplicity of your suggested approach makes perfect sense!

dwave/optimization/model.pyx
```diff
@@ -1764,7 +1764,15 @@ cdef class ArraySymbol(Symbol):
...
             from dwave.optimization.symbols import Square  # avoid circular import
             return Square(self)
-        raise ValueError("only squaring is currently supported")
+        # check if exponent is an integer greater than 0
+        elif isinstance(exponent, numbers.Real) and exponent > 0 and int(exponent) == exponent:
+            expanded = itertools.repeat(self, int(exponent))
+            out = next(expanded)  # get the first one
+            # multiply self by itself exponent times
+            for symbol in expanded:
+                out = out.__imul__(symbol)
```
arcondello, line 1773: Why not just use `*=`? It's the same thing and a bit more readable. Also more immune to later changes we might make.

jwoo098: I will change it back. I changed from `*=` to `__imul__` because I was wondering if that was the cause of failing the compatibility tests (which it was not).

releasenotes/notes/add-supported-arraysymbol-exponents-c7be57d09c097671.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Add support for exponentiation of ``ArraySymbol`` for positive integer exponents other than 2
```
arcondello, line 4:
```suggestion
    Add support for exponentiation of ``ArraySymbol`` for positive integer exponents other than 2.
```

PR #89, opened 2024-08-15
Feature/disjoint more mutate methods

wbernoudy:


dwave/optimization/include/dwave-optimization/nodes/collections.hpp
```diff
@@ -221,12 +221,18 @@ class DisjointListsNode : public DecisionNode {
...
+    void rotate_sublist_right(State& state, ssize_t list_index, ssize_t start, ssize_t stop,
+                              ssize_t num_rotations) const;
+
     void swap_in_list(State& state, ssize_t disjoint_list, ssize_t element_i,
                       ssize_t element_j) const;
 
     void pop_to_list(State& state, ssize_t from_disjoint_list, ssize_t element_i,
                      ssize_t to_disjoint_list, ssize_t element_j) const;
 
+    void set_full_list_state(State& state, ssize_t list_index,
```
arcondello, line 233: Feels like `set_state(...)` covers this and is a bit more succinct?

Also, we should accept a `span` rather than a `vector`.

dwave/optimization/src/nodes/collections.cpp
```diff
@@ -464,6 +464,47 @@ struct DisjointListStateData : NodeStateData {
         }
     }
 
+    // "Manually" set the full state of one of the disjoint lists. This will not check that
+    // the new state is a valid configuration (given the state of the other lists), so it
```
arcondello, line 468: Do we want to check valid when asserts are on? Feels like that will save some bugs in the future.

dwave/optimization/include/dwave-optimization/nodes/collections.hpp
```diff
@@ -221,12 +221,18 @@ class DisjointListsNode : public DecisionNode {
 
     void rotate_in_list(State& state, ssize_t list_index, ssize_t dest_idx, ssize_t src_idx) const;
 
+    void rotate_sublist_right(State& state, ssize_t list_index, ssize_t start, ssize_t stop,
```
arcondello, line 224: Is this implemented?

dwave/optimization/src/nodes/collections.cpp
wbernoudy: This comment is slightly misleading as there is no way to check for correctness in this function. The function must be called multiple times on different lists to end up in a valid state if you are moving items between lists.

dwave/optimization/include/dwave-optimization/nodes/collections.hpp
wbernoudy: Nope, will remove

dwave/optimization/src/nodes/collections.cpp
arcondello: I see. Maybe a validity check when `propagate()` is called then?

arcondello (approved):

arcondello (approved):

PR #90, opened 2024-08-15
Build with meson

arcondello:
Try replacing `setuptools_dso` with [meson](https://mesonbuild.com/).


arcondello:
Pattern match: https://github.com/scipy/scipy/pull/20321

arcondello:
Ok, have macos working (see https://github.com/mesonbuild/meson/issues/13554). Onto windows next :smile:

arcondello:
I'll squash the history before merging.

There is definitely more that can be done with `meson` (e.g. updating the C++ tests to use meson), but this should be sufficient for now.

meson.build
```diff
@@ -0,0 +1,82 @@
...
+    ).stdout(),
+)
+
+# We want some debugging symbols
+add_project_arguments(['-g1'], language: ['cpp'])
+
+py = import('python').find_installation(pure: false)
+
+dwave_optimization_include = include_directories('dwave/optimization/include/')
+dwave_optimization_src = [
```
arcondello, line 25: To head off an obvious question, [meson doesn't allow wildcards](https://mesonbuild.com/FAQ.html#why-cant-i-specify-target-files-with-a-wildcard).

thisac (approved):
Overall LGTM. Tested on both Windows 10 and WSL2, and the tests pass.

meson.build
```diff
@@ -0,0 +1,82 @@
+project(
```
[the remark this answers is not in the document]
arcondello: We haven't included them in our `setup.cfg`, `pyproject.toml`, etc before... Also `meson` gets a bit grumpy about `project()` not being the first line.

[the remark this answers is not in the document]
thisac: Ah, I compared it mostly to `setup.py`. It looks too much like a Python file.

pyproject.toml
```diff
@@ -1,18 +1,56 @@
...
+]
+
+[project.readme]
+file = "README.rst"
+content-type = "text/x-rst"
+
+[project.urls]
+Changelog = "https://github.com/dwavesystems/dwave-optimization/releases"
+Documentation = "https://docs.ocean.dwavesys.com"
+"Source Code" = "https://github.com/dwavesystems/dwave-optimization"
```
thisac, line 43: Why these url names? I've used e.g., [this](https://daniel.feldroy.com/posts/2023-08-pypi-project-urls-cheatsheet) or [this](https://github.com/patrick91/links-demo) as references.

arcondello: :shrug: Why not these? There doesn't seem to be any consensus even within `pypa`

https://github.com/numpy/numpy/blob/ae926d7e01053baede90a989fb9bf98515d23a22/pyproject.toml#L53-L59
https://github.com/pypa/cibuildwheel/blob/3a30bec9ef7f8521c8f98e77a7091deafb3938a9/pyproject.toml#L91-L93
https://github.com/pypa/setuptools/blob/8e963821d37216f6ba416b14a51c628606be7f76/pyproject.toml#L31-L33
https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#urls

These are the ones I like to see so went with them. I am not sure this is a place where consistency matters much.

thisac: Fair enough. I guess the only thing would be getting relevant icons for PyPI, but as you say it seems to work fine no matter what's used.

wbernoudy (approved):

PR #93, opened 2024-08-16
Bin packing generator

eordog:


dwave/optimization/generators.py
```diff
@@ -145,6 +146,69 @@ def quadratic_model(qm):
 
     return model
 
+def bin_packing(weights: numpy.typing.ArrayLike,
+                bin_capacity: float
```
arcondello, line 150: We should keep the argument names consistent with https://github.com/dwavesystems/dimod/pull/1384. IMO `bin_` is implied here, so I think just `capacity` is fine.

dwave/optimization/generators.py
```diff
@@ -145,6 +146,69 @@ def quadratic_model(qm):
...
+    # Construct model
+    model = Model()
+
+    # Add constants
+    weights = model.constant(weights)
+    bin_capacity = model.constant(bin_capacity)
+
+    # Create disjoint bit sets to represent bins
+    max_bins = num_items
+    main, bins = model.disjoint_bit_sets(num_items, max_bins)
```
arcondello, line 193: This could also potentially be formulated with `DisjointSets`. The main benefit there is we could use the `Size` symbol (added in https://github.com/dwavesystems/dwave-optimization/pull/50) to determine set emptiness. Not sure about the performance implications though.

dwave/optimization/generators.py
```diff
@@ -145,6 +146,69 @@ def quadratic_model(qm):
...
+    for i in range(max_bins):
+        bin_array = bins[i]
+        model.add_constraint((bin_array*weights).sum() <= bin_capacity)
+
+    # Minimize the number of bins that are non-empty
+    constant_one = model.constant(1)
+    count_helper = model.constant(np.ones(num_items))
+    obj_val = model.constant(0)
+    for i in range(max_bins):
+        obj_val += ((bins[i]*count_helper).sum() >= constant_one)
```
arcondello, line 205: Aren't the bins themselves `0/1`? In which case I think we can use `bins[i].sum()` directly.

dwave/optimization/generators.py
eordog: Do you mean `DisjointLists`?

arcondello: Oh, I forgot that doesn't exist yet. It's to `DisjointLists` what `Set` is to `List`. But I guess we haven't implemented it 😳

eordog: Okay, should I wait until it exists, or use `DisjointBitSets` (and maybe update when `DisjointSets` exists), or use `DisjointLists` (probably not that last one)?

arcondello: I think it makes sense to stick with `DisjointBitSets` for now, and update later if we like.

dwave/optimization/generators.py
eordog: All suggested changes made, thanks!

arcondello (approved):
LGTM! Thanks

PR #96, opened 2024-08-26
Add helper class for node states

arcondello:
I tried to minimize the changes, so there is definitely more we could do, but this should reduce some of the code duplication as well as provide a centralized place for bug fixes and/or performance updates.

The main cost is that now all of the non-dynamic nodes keep a redundant `ssize_t previous_size_` on their state, plus the very minor runtime cost of checking it. We could avoid that with templates, but IMO the extra complexity isn't worth it.

### DIFF
--- a/dwave/optimization/src/nodes/mathematical.cpp
+++ b/dwave/optimization/src/nodes/mathematical.cpp
@@ -16,28 +16,12 @@
 
 #include <ranges>
 
+#include "_state.hpp"
 #include "dwave-optimization/utils.hpp"
 
 namespace dwave::optimization {
 
 // BinaryOpNode ************************************************************
-
-struct BinaryOpNodeData : public NodeStateData {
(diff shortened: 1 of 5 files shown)


dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -1034,29 +998,23 @@ double UnaryOpNode<UnaryOp>::min() const {
...
 
-    if (node_data->diff.size()) Node::propagate(state);
+    if (node_data->updates.size()) Node::propagate(state);
 }
 
 template <class UnaryOp>
 void UnaryOpNode<UnaryOp>::revert(State& state) const {
-    auto node_data = data_ptr<UnaryOpNodeStateData>(state);
-
-    for (const auto& update : node_data->diff) {
```
wbernoudy, line 1055: Whoops!

wbernoudy (approved):
LGTM!

PR #100, opened 2024-08-28
Add non-`const` overload of `::operands()`

arcondello:
Also fixes a badly named test file.


dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
```diff
@@ -83,7 +87,7 @@ class BinaryOpNode : public ArrayOutputMixin<ArrayNode> {
     // There are redundant, because we could dynamic_cast each time from
     // predecessors(), but that gets tedious as well as having a very minor
```
wbernoudy, line 88: Just because I'm reading this comment for the third time this week: we should probably change this to (sometimes) having a very major performance hit! :smile:

wbernoudy (approved):

PR #103, opened 2024-08-29
Support Python 3.13

arcondello:


requirements.txt
```diff
@@ -1,6 +1,6 @@
 # Install requirements
 numpy==1.21.6; python_version < '3.10'  # support numpy.typing
-oldest-supported-numpy; python_version >= '3.10'
+numpy==2.1.0; python_version >= '3.10'
```
arcondello, line 3: `oldest-supported-numpy` is deprecated. And now that we don't require NumPy as build time we can be a lot less finicky with it.

arcondello:
Looks like there isn't yet free-threaded support for Cython. Will check back in on this in a few weeks.

arcondello:
I am not 100% sure what's happening with the macos x86_64 build. Seems to be an interaction between `meson` and `cibuildwheel`. I'll let it sit for a few weeks in case it's something that will be ironed out.

thisac (approved):
LGTM

PR #105, opened 2024-09-03
Fix assertion errors seen with `_GLIBCXX_ASSERTIONS`

arcondello:
Discovered while messing around with building tests with `meson`. Have not activated `_GLIBCXX_ASSERTIONS` in CI yet, will do that in a followup PR (pending decisions on tests-with-`meson`).

### DIFF
--- a/dwave/optimization/src/nodes/indexing.cpp
+++ b/dwave/optimization/src/nodes/indexing.cpp
@@ -527,15 +527,19 @@ void AdvancedIndexingNode::initialize_state(State& state) const {
     const ssize_t size = std::get<ArrayNode*>(indices_[first_array_index_])->size(state);
 
     std::vector<ssize_t> offsets(size);
-    for (ssize_t index = 0; index < static_cast<ssize_t>(indices_.size()); ++index) {
-        if (std::holds_alternative<ArrayNode*>(indices_[index])) {
-            assert(array_strides[index] % static_cast<ssize_t>(itemsize()) == 0);
-            ssize_t stride = array_strides[index] / static_cast<ssize_t>(itemsize());
+    if (size) {
+        for (ssize_t index = 0; index < static_cast<ssize_t>(indices_.size()); ++index) {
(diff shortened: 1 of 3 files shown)


dwave/optimization/src/nodes/indexing.cpp
```diff
@@ -527,15 +527,19 @@ void AdvancedIndexingNode::initialize_state(State& state) const {
...
-                ++it;
+    if (size) {
+        for (ssize_t index = 0; index < static_cast<ssize_t>(indices_.size()); ++index) {
+            if (std::holds_alternative<ArrayNode*>(indices_[index])) {
+                assert(array_strides[index] % static_cast<ssize_t>(itemsize()) == 0);
+
+                assert(index < static_cast<ssize_t>(array_strides.size()));
+                ssize_t stride = array_strides[index] / static_cast<ssize_t>(itemsize());
+
+                auto it = offsets.begin();
```
wbernoudy, line 538: I'm assuming the issue is with this call to `begin()`, what is the assertion error that you get here?

arcondello: same as if we did `offsets[0]`. Honestly I am a bit unsure why it raises the assertion error here and not when doing things like
```c++
std::reduce(vec.begin(), vec.end());
```
which it doesn't seem to mind :shrug:

wbernoudy: But `offsets[0]` when `offsets` is empty is undefined behavior, while this call to `begin()` is well-defined and benign.

arcondello: I agree, and yet this is the line that throws the assertion. I suspect it's because we then go and use it in the loop, so the compiler is translating it into a reference access or something.

arcondello: Also it's certainly harmless to bypass the entire for-loop in this way, and possibly even beneficial.

wbernoudy (approved):

PR #107, opened 2024-09-07
Add CVRPTW generator

elham-azd:


arcondello:
Closes https://github.com/dwavesystems/dwave-optimization/issues/104

dwave/optimization/generators.py
```diff
@@ -195,7 +197,7 @@ def bin_packing(weights: numpy.typing.ArrayLike,
     # Ensure that the weight for each bin does not exceed capacity
     for i in range(max_bins):
         bin_array = bins[i]
-        model.add_constraint((bin_array*weights).sum() <= capacity)
+        model.add_constraint((bin_array * weights).sum() <= capacity)
```
arcondello, line 200: I'd prefer to keep the diff minimal by not reformatting other parts of the code. It makes it harder to review and messes with the git history. Formatting could be done as a separate PR.

dwave/optimization/generators.py
```diff
@@ -407,6 +409,243 @@ class as the decision variable being optimized, with permutations of its
     return model
 
 
+def capacitated_vehicle_routing_time_window(demand: numpy.typing.ArrayLike,
```
arcondello, line 412:
```python
def capacitated_vehicle_routing_with_time_windows(demand: numpy.typing.ArrayLike,
```
I know it's super long, but I think this is the "real" name.

dwave/optimization/generators.py
```diff
@@ -407,6 +409,243 @@ class as the decision variable being optimized, with permutations of its
...
+    be served after time_window_open and before time_window_close. Vehicles have a
+    specified delivery capacity, and on the routes to locations and then back
+    to the depot, no vehicle is allowed to exceed its carrying capacity.
+
+    Args:
+        demand:
+            Customer demand, as an |array-like|_. The first element is the depot and must
+            be zero.
+        number_of_vehicles:
+            Number of available vehicles, as an integer.
```
arcondello, line 435:
```
            Number of available vehicles, as a positive integer.
```

dwave/optimization/generators.py
```diff
@@ -407,6 +409,243 @@ class as the decision variable being optimized, with permutations of its
...
+        raise ValueError(f"Total demand, {demand.sum()}, exceeds the "
+                         f"total capacity, {number_of_vehicles * vehicle_capacity}.")
+
+    time_distances_array = _require("time_distances", time_distances, ndim=2, nonnegative=True, square=True)
+
+    time_window_open = _require("time_window_open", time_window_open, dtype=float, ndim=1, nonnegative=True)
+    time_window_close = _require("time_window_close", time_window_close, dtype=float, ndim=1, nonnegative=True)
+    service_time = _require("service_time", service_time, dtype=float, ndim=1, nonnegative=True)
+
+    if len(time_distances_array) < 2:
```
arcondello, line 486: `len()` in this case returns the size of the first dimension. IMO we should use either `.size` which returns the number of elements or `.shape[0]`. The error message should probably reflect that as well.

dwave/optimization/generators.py
elham-azd: Got it. That makes sense. I used PyCharm automated formatting and it has changed other lines as well. Will make sure that I undo changes to other lines in future.

dwave/optimization/generators.py
```diff
@@ -407,6 +409,243 @@ class as the decision variable being optimized, with permutations of its
     return model
 
 
+def capacitated_vehicle_routing_with_time_windows(demand: numpy.typing.ArrayLike,
+                                                  number_of_vehicles: int,
+                                                  vehicle_capacity: float,
+                                                  time_distances: typing.Optional[numpy.typing.ArrayLike] = None,
+                                                  time_window_open: typing.Optional[numpy.typing.ArrayLike] = None,
+                                                  time_window_close: typing.Optional[numpy.typing.ArrayLike] = None,
+                                                  service_time: typing.Optional[numpy.typing.ArrayLike] = None,
```
arcondello, line 418: If these are allowed to be `None`, the docs should explain what they get defaulted to.

dwave/optimization/generators.py
```diff
@@ -407,6 +409,243 @@ class as the decision variable being optimized, with permutations of its
     return model
 
 
+def capacitated_vehicle_routing_with_time_windows(demand: numpy.typing.ArrayLike,
+                                                  number_of_vehicles: int,
+                                                  vehicle_capacity: float,
+                                                  time_distances: typing.Optional[numpy.typing.ArrayLike] = None,
```
arcondello, line 415: It doesn't look like `time_distances` can actually be `None` in the code.

dwave/optimization/generators.py
```diff
@@ -407,6 +409,243 @@ class as the decision variable being optimized, with permutations of its
...
+            if client_idx == 0:
+                # index of the first customer visited, if length of the route is zero,
+                # points to the last (0) element
+                idx = where(condition, routes[vehicle_idx][:1].sum(), range_helper[-1])
+                this_t_serving_time.append(maximum(t_depo[idx], t_open[idx]))
+                this_t_leaving.append(this_t_serving_time[-1] + t_service[idx])
+
+            else:
+                # index of the previous customer visited
+                previous_idx = (condition * routes[vehicle_idx][client_idx - 1:client_idx].sum()
```
arcondello, line 607: Why aren't we using a `where` here? I see that replacing this statement with
```python
                previous_idx = where(condition, routes[vehicle_idx][client_idx - 1:client_idx].sum(), range_helper[-1])
```
leads to test failures. But shouldn't it be the same? I feel like I am missing something obvious.

arcondello: Oh, duh, the test that's failing is checking the number of nodes. So that's expected.

tests/test_generators.py
```diff
@@ -361,6 +359,228 @@ def test_state_serialization(self):
...
+            vehicle_capacity=10,
+            time_distances=[[1, 6, 3], [2, 2, 8], [4, 6, 7]],
+            time_window_open=[0, 0, 0],
+            time_window_close=[20, 20, 20],
+            service_time=[0, 0, 0])
+
+        self.assertEqual(model.num_decisions(), 1)
+        self.assertEqual(model.num_constraints(), 12)
+        self.assertEqual(model.num_nodes(), 177)
+        self.assertEqual(model.num_edges(), 286)
```
arcondello, line 508: I am not sure these tests help much. At the least we should change to `self.assertLessEqual(...)` to support future improvements. But IMO we should drop them.

dwave/optimization/generators.py
elham-azd: Thank you, I missed this line when adding where nodes. Will fix this.

tests/test_generators.py
elham-azd: Sounds reasonable. I'm going to drop those.

dwave/optimization/generators.py
elham-azd: Good catch. Actually I think none of the input parameters can be `None`.  Fixed

PR #110, opened 2024-09-18
Fix circleci time skew bug

arcondello:


arcondello:
Turns out to be my error. My system time wasn't synced. To be fair, meson does a funny thing whereby they set the time by the last git commit, see https://github.com/scipy/scipy/issues/19158. So my system was setting the commit time into the future.

Closing.

PR #111, opened 2024-09-18
Fix combined indexing

arcondello:
Previously indexing an array symbol by a mixture of arrays, integers, and non-empty slices would always result in an error.

### DIFF
--- a/dwave/optimization/generators.py
+++ b/dwave/optimization/generators.py
@@ -605,23 +605,19 @@ def job_shop_scheduling(times: numpy.typing.ArrayLike, machines: numpy.typing.Ar
 
     # Ensure that for each job, its tasks do not overlap
     for j in range(num_jobs):
-        ends = end_times[j, :][machines[j, :-1]]
-        starts = start_times[j, :][machines[j, :]][1:]
+        ends = end_times[j, machines[j, :-1]]
+        starts = start_times[j, machines[j, 1:]]
+
         model.add_constraint((ends <= starts).all())
 
     # Ensure for each machine, its tasks do not overlap
     for m in range(num_machines):
         order = orders[m]
 
-        ends = end_times[:, m]
-        starts = start_times[:, m]
+        ends = end_times[order[:-1], m]
+        starts = start_times[order[1:], m]
 
-        # todo: with combined indexing we wouldn't need this loop
-        for t in range(num_jobs - 1):
-            end = ends[order[t]]  # end of the t'th job on machine m
-            start = starts[order[t+1]]  # start of the (t+1)th job on machine m
-
-            model.add_constraint(end <= start)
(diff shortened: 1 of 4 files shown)


dwave/optimization/model.pyx
```diff
@@ -1706,8 +1706,8 @@ cdef class ArraySymbol(Symbol):
                 # order may be more efficient in some cases, but for now let's do the simple thing
 
                 basic_indices, advanced_indices = _split_indices(index)
-                basic = dwave.optimization.symbols(self, *basic_indices)
-                return dwave.optimization.symbols(basic, *advanced_indices)
+                basic = dwave.optimization.symbols.BasicIndexing(self, *basic_indices)
```
alexzucca90, line 1709: fun bug!

dwave/optimization/generators.py
```diff
@@ -605,23 +605,19 @@ def job_shop_scheduling(times: numpy.typing.ArrayLike, machines: numpy.typing.Ar
...
+        ends = end_times[j, machines[j, :-1]]
+        starts = start_times[j, machines[j, 1:]]
+
         model.add_constraint((ends <= starts).all())
 
     # Ensure for each machine, its tasks do not overlap
     for m in range(num_machines):
         order = orders[m]
 
-        ends = end_times[:, m]
```
alexzucca90, line 616: now the question is: here we were using one single basic indexing and reused that all over. Unless I'm missing something, we are increasing the number of nodes, right?

dwave/optimization/generators.py
```diff
@@ -605,23 +605,19 @@ def job_shop_scheduling(times: numpy.typing.ArrayLike, machines: numpy.typing.Ar
...
-
-        # todo: with combined indexing we wouldn't need this loop
-        for t in range(num_jobs - 1):
-            end = ends[order[t]]  # end of the t'th job on machine m
-            start = starts[order[t+1]]  # start of the (t+1)th job on machine m
+        ends = end_times[order[:-1], m]
+        starts = start_times[order[1:], m]
 
-            model.add_constraint(end <= start)
+        model.add_constraint((ends <= starts).all())
```
alexzucca90, line 620: but then we decrease the number of nodes with this..

dwave/optimization/generators.py
arcondello: I guess we could save `order[:-1]` and `order[1:]` if we wanted to... we really just need to add node deduplication as a presolve step. Maybe I'll look at that.

dwave/optimization/generators.py
arcondello: But yes, net far fewer nodes this way.

alexzucca90 (approved):
Not sure if you have other things in mind for the node counts, but other than that it LGTM

PR #112, opened 2024-09-19
Fix behavior for scalar indices in adv indexing

wbernoudy:
This should achieve the expected behavior for scalar (0-dimensional) `ArrayNode` indices for `AdvancedIndexingNode`. Previously there were a few asserts and an exception raised in one case but it was not properly prevented, and also not properly implemented.

### DIFF
--- a/dwave/optimization/src/nodes/indexing.cpp
+++ b/dwave/optimization/src/nodes/indexing.cpp
@@ -212,7 +212,7 @@ struct AdvancedIndexingNodeData : NodeStateData {
 };
 
 struct AdvancedIndexingNode::IndexParser_ {
-    IndexParser_(Array* array_ptr, std::vector<array_or_slice>&& indices)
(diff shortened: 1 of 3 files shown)


dwave/optimization/src/nodes/indexing.cpp
```diff
@@ -212,7 +212,7 @@ struct AdvancedIndexingNodeData : NodeStateData {
 };
 
 struct AdvancedIndexingNode::IndexParser_ {
-    IndexParser_(Array* array_ptr, std::vector<array_or_slice>&& indices)
+    IndexParser_(Array* array_ptr, std::vector<array_or_slice>&& indices, ssize_t item_size)
```
arcondello, line 215: Why not get the `itemsize` from the array? Also, for consistency, I would use `itemsize` one word.

arcondello:
Just needs a release note!

arcondello (approved):
LGTM!

Issue #47, opened 2024-07-08
Consider allowing reduction along a single axis


arcondello:
We should add an `axis` keyword argument to reduction operations. E.g.,
```python
x.sum(axis=0)
```

See for instance https://numpy.org/doc/stable/reference/generated/numpy.ndarray.sum.html#numpy.ndarray.sum

Issue #113, opened 2024-09-19
Specify a standard version of `clang-format`


wbernoudy:
Since we expect the C++ code to be formatted with `clang-format`, but there are inconsistencies between different versions, we should specify the version one is expected to use when doing development.

arcondello:
We do not currently enforce clang-formatted code. We could consider doing so as well.

wbernoudy:
Agreed! For now at least, it would be nice to just have a note in the readme.

arcondello:
I don't think this is easy.

The ecosystem as a whole doesn't have a lot of consensus.

* VSCode uses 18.1.7 (I think?) https://github.com/microsoft/vscode-cpptools/blob/1e89750b0bb9629ba0262a6eb44d1ab41eb77d51/Extension/CHANGELOG.md?plain=1#L137
* `apt install clang format` pulls in 14.0.0
* The LLVM source tree itself doesn't seem to be well formatted, albeit as of [March 2022](https://clang.llvm.org/docs/ClangFormattedStatus.html).

For internal repos, I am reasonably comfortable enforcing a version. But for public repos like this, where we want to encourage contribution from users, I don't think it's reasonable to force folks to install a specific version of clang simply to make a PR against this repo.

arcondello:
FWIW, https://github.com/ssciwr/clang-format-wheel feels like it might actually help. Could make it part of the `dev` dependencies and enforce a version.

arcondello:
Also has a pre-commit option: https://github.com/ssciwr/clang-format-wheel#use-from-pre-commit

PR #114, opened 2024-09-20
Feature/partial sum

alexzucca90:
Added partial sum node that allows to do stuff like 
```python
from dwave.optimization import Model
model = Model()
x = model.integer((3, 3))
y = x.sum(axis=0)
z = x.sum(axis=1)
```


alexzucca90:
Just realized, this (partially) closes #47

releasenotes/notes/partial-sum-node-ca3356c3bddfacde.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - Add C++ index raveling-unraveling utilities functions.
+  - Add C++ ``PartialReduceNode``.
+  - Add Python ``PartialSum`` symbol and added ``axis`` keyword in ``ArraySymbol.sum`` function.
```
arcondello, line 5:
```suggestion
  - Add Python ``PartialSum`` symbol and added ``axis`` keyword in ``ArraySymbol.sum()`` function.
```

dwave/optimization/include/dwave-optimization/array.hpp
```diff
@@ -737,6 +738,27 @@ std::vector<ssize_t> broadcast_shape(const std::span<const ssize_t> lhs,
...
+/// Convert a flat index to multi-index
+std::vector<ssize_t> unravel_index(const std::span<const ssize_t> strides, ssize_t index);
+
+/// Convert multi index to flat index
+ssize_t ravel_multi_index(const std::span<const ssize_t> strides,
+                          const std::span<const ssize_t> indices);
+
+ssize_t ravel_multi_index(const std::span<const ssize_t> strides,
+                          std::initializer_list<ssize_t> indices);
+
```
arcondello, line 761: For now I would have these live in `mathematical.cpp` alongside the node implementation. It doesn't look like we use them anywhere else. We can always promote them later if we want them.

Or do you think they would be insufficiently tested if we don't have access to them publicly?

dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
```diff
@@ -151,6 +151,73 @@ using NaryMaximumNode = NaryOpNode<functional::max<double>>;
 using NaryMinimumNode = NaryOpNode<functional::min<double>>;
 using NaryMultiplyNode = NaryOpNode<std::multiplies<double>>;
 
+/// TODO: support multiple axes
+template <class BinaryOp>
+class PartialReduceNode : public ArrayOutputMixin<ArrayNode> {
```
arcondello, line 156: I haven't gone back and done it for old classes. But I'd like new ones to have the methods in alphabetical order. The old ones are driving me somewhat nuts.

dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -531,6 +531,272 @@ template class NaryOpNode<functional::min<double>>;
 template class NaryOpNode<std::multiplies<double>>;
 template class NaryOpNode<std::plus<double>>;
 
+// PartialReduceNode *****************************************************************
+template <class BinaryOp>
+struct ExtraArrayData {
+    void commit() {}
+    void revert() {}
+};
```
arcondello, line 539: I am guessing this is for future-proofing? Because it's in the `.cpp` file there are no binary compatibility issues so I'd remove it for now. That would also allow you to use `ArrayNodeStateData` directly I believe.

dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -531,6 +531,272 @@ template class NaryOpNode<functional::min<double>>;
...
+    }
+
+    /// 3. Iterate through the parent on the correct axis.
+    double val = init;
+    auto func = op();
+    for (ssize_t i = begin; i < array_ptr_->shape()[this->axis_]; ++i) {
+        ssize_t p_index =
+                start_idx + i * this->parent_strides_c_[this->axis_] / array_ptr_->itemsize();
+        assert(p_index < static_cast<ssize_t>(array_ptr_->size(state)));
+        double other = array_ptr_->view(state)[p_index];
```
arcondello, line 737: It feels a bit redundant to keep the strides as if the parent was contiguous, then also use `view` to account for the parent's strides.

Maybe worth just working with the parent's actual strides directly?

dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -531,6 +531,272 @@ template class NaryOpNode<functional::min<double>>;
...
+                "without supplying an initial value");
+    } else if (array_ptr_->size() < 1) {
+        throw std::invalid_argument(
+                "cannot do a reduction on an empty array with an operation that has no identity "
+                "without supplying an initial value");
+    }
+
+    /// TODO: support negative axis
+    if (axis_ < 0 || axis_ >= array_ptr_->ndim()) {
+        throw std::invalid_argument("Axis should be an integer between 0 and n_dim - 1");
```
arcondello, line 601:
```c++
        throw std::invalid_argument("Axis should be an integer between 0 and ndim - 1");
```
for consistency with the method name

dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -531,6 +531,272 @@ template class NaryOpNode<functional::min<double>>;
...
+template <class BinaryOp>
+void PartialReduceNode<BinaryOp>::initialize_state(State& state) const {
+    int index = topological_index();
+    assert(index >= 0 && "must be topologically sorted");
+    assert(static_cast<int>(state.size()) > index && "unexpected state length");
+    assert(state[index] == nullptr && "already initialized state");
+
+    std::vector<double> values(size(state));
+
+    for (ssize_t i = 0; i < size(state); ++i) {
```
arcondello, line 631:
```suggestion
    for (ssize_t i = 0, stop = size(state); i < stop; ++i) {
```
See https://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop. Not quite the same, but the principal applies. Especially since `size()` is a virtual function.

dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -531,6 +531,272 @@ template class NaryOpNode<functional::min<double>>;
...
+        assert(i < static_cast<ssize_t>(values.size()));
+        values[i] = reduce(state, i);
+    }
+
+    state[index] = std::make_unique<PartialReduceNodeData<op>>(std::move(values));
+}
+
+template <class BinaryOp>
+bool PartialReduceNode<BinaryOp>::integral() const {
```
arcondello, line 640: Apologies for this, but these'll have to be updated along the lines of https://github.com/dwavesystems/dwave-optimization/pull/115

dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -531,6 +531,272 @@ template class NaryOpNode<functional::min<double>>;
...
+template <>
+void PartialReduceNode<std::plus<double>>::propagate(State& state) const {
+    auto ptr = data_ptr<PartialReduceNodeData<op>>(state);
+
+    auto& values = ptr->buffer;
+    auto& changes = ptr->updates;
+
+    for (const auto& [p_index, old, value] : array_ptr_->diff(state)) {
+        auto index = _map_parent_index(state, p_index);
+        auto previous = values[index];
```
arcondello, line 681: Style thing, can be ignored. But IMO `auto` isn't adding clarity here.
```c++
        const ssize_t index = _map_parent_index(state, p_index);
        const double previous = values[index];
```

dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
```diff
@@ -151,6 +151,73 @@ using NaryMaximumNode = NaryOpNode<functional::max<double>>;
...
+        return std::span<Array* const, 1>(&array_ptr_, 1);
+    }
+    std::span<const Array* const> operands() const {
+        assert(predecessors().size() == 1);
+        return std::span<const Array* const, 1>(&array_ptr_, 1);
+    }
+
+    const std::optional<double> init;
+
+    ssize_t axis() const { return axis_; }
```
arcondello, line 197: If we think there is a decent chance we want to support multiple axis in the future, then IMO
```c++

    std::span<const ssize_t> axis() const { ... }
```
will be more future proof. And we should also move the implementation to the `.cpp` for binary compatibility.

dwave/optimization/symbols.pyx
```diff
@@ -2217,6 +2219,52 @@ cdef class Or(ArraySymbol):
...
+    def _from_zipfile(cls, zf, directory, Model model, predecessors):
+        if len(predecessors) != 1:
+            raise ValueError("PartialSum must have exactly one predecessor")
+
+        with zf.open(directory + "axis.json", "r") as f:
+            return PartialSum(*predecessors, json.load(f))
+
+    def _into_zipfile(self, zf, directory):
+        encoder = json.JSONEncoder(separators=(',', ':'))
+        zf.writestr(directory + "axis.json", encoder.encode(self.ptr.axis()))
```
arcondello, line 2261: I would save the list of axis for future-proofing.

dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
arcondello: Or should it be `axes`?

dwave/optimization/src/nodes/mathematical.cpp
alexzucca90: yeah, in case of e.g. a product along an axis, then we'd need this extra array data. And it's gotta be an array and not scalars of course. I can remove this extra machinery or keep it commented out for future.

arcondello: I'd remove for now, rather than commenting out. It'll be in the git history if we need it later.

dwave/optimization/src/nodes/mathematical.cpp
alexzucca90: Good point, let me rethink about the logic. I'll reply here later on.

dwave/optimization/symbols.pyi
```diff
@@ -151,6 +151,9 @@ class Not(ArraySymbol):
 class Or(ArraySymbol):
     ...
 
+class PartialSum(ArraySymbol):
```
alexzucca90, line 154: forgot an extra line here

dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
alexzucca90: > Or should it be axes?
For consistency with numpy we can leave it as `axis` though that's a keyword argument rather that the name of the function. However the return type `std::span<const ssize_t> ` suggests to me to call it `axes`.

arcondello: I think we should keep the input keyword argument as `axis` for consistency with NumPy. But I see they do use both [axis and axes](https://numpy.org/doc/stable/glossary.html#term-axis) so I would rename this method and other places where we store multiple as `axes`.

dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
arcondello: Yes. IMO once the number of methods gets high enough, the grouping becomes more confusing than helpful. Most times you're looking for a specific method and being able to find it quickly is helpful.

alexzucca90: roger!

dwave/optimization/src/array.cpp
```diff
@@ -38,8 +38,7 @@ bool SizeInfo::operator==(const SizeInfo& other) const {
     if (this->array_ptr != other.array_ptr) return false;
 
     return (this->multiplier == other.multiplier && this->offset == other.offset &&
-            this->min.value_or(0) == other.min.value_or(0) &&
-            this->max == other.max);
+            this->min.value_or(0) == other.min.value_or(0) && this->max == other.max);
```
arcondello, line 41: Please don't edit unrelated functions for formatting. It clutters the git history and without https://github.com/dwavesystems/dwave-optimization/issues/113 it's not even deterministic.

dwave/optimization/src/array.cpp
```diff
@@ -192,9 +191,7 @@ bool array_shape_equal(const Array* lhs_ptr, const Array* rhs_ptr) {
 
     return false;
 }
-bool array_shape_equal(const Array& lhs, const Array& rhs) {
-    return array_shape_equal(&lhs, &rhs);
-}
+bool array_shape_equal(const Array& lhs, const Array& rhs) { return array_shape_equal(&lhs, &rhs); }
```
arcondello, line 194: Please don't edit unrelated functions for formatting. It clutters the git history and without https://github.com/dwavesystems/dwave-optimization/issues/113 it's not even deterministic.

dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -663,6 +663,279 @@ template class NaryOpNode<functional::min<double>>;
...
+        // or an integer default init
+        return array_ptr_->integral() && is_integer(init.value_or(0));
+    }
+
+    assert(false && "not implemeted yet");
+    unreachable();
+}
+
+template <class BinaryOp>
+double PartialReduceNode<BinaryOp>::max() const {
```
arcondello, line 790: This also needs an update along the lines of https://github.com/dwavesystems/dwave-optimization/pull/115

dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -663,6 +663,279 @@ template class NaryOpNode<functional::min<double>>;
...
+
+    if constexpr (std::is_same<result_type, bool>::value) {
+        return true;
+    }
+    // there are other cases we could/should handle here.
+    return Array::max();
+}
+
+template <class BinaryOp>
+double PartialReduceNode<BinaryOp>::min() const {
```
arcondello, line 801: This also needs an update along the lines of https://github.com/dwavesystems/dwave-optimization/pull/115

dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -750,8 +1023,7 @@ template <>
 ReduceNode<std::plus<double>>::ReduceNode(ArrayNode* array_ptr) : ReduceNode(array_ptr, 0) {}
 
 template <class BinaryOp>
-ReduceNode<BinaryOp>::ReduceNode(ArrayNode* array_ptr)
-        : init(), array_ptr_(array_ptr) {
+ReduceNode<BinaryOp>::ReduceNode(ArrayNode* array_ptr) : init(), array_ptr_(array_ptr) {
```
arcondello, line 1026: Please don't edit unrelated functions for formatting. It clutters the git history and without https://github.com/dwavesystems/dwave-optimization/issues/113 it's not even deterministic.

tests/cpp/tests/test_array.cpp
```diff
@@ -95,7 +95,8 @@ TEST_CASE("ArrayIterator") {
         }
 
         WHEN("We create a mask over the array and create a masked iterator") {
-            auto mask = std::vector<double>{true, false, false, true, false, false, false, true, false};
+            auto mask =
```
arcondello, line 98: Please don't edit unrelated functions for formatting. It clutters the git history and without https://github.com/dwavesystems/dwave-optimization/issues/113 it's not even deterministic.

dwave/optimization/include/dwave-optimization/array.hpp
alexzucca90: I put them here for exactly that purpose, as I was writing tests when developing. Since the broadcasting shape functions were in this file already, I thought it was consistent. I agree the tests are somewhat simple,  but better over-testing IMHO. If we find bugs, at least we can rule out they're not coming from raveling-unraveling indices.

arcondello: FWIW, IMO including `broad_shape()` in this file was a mistake for the reasons I outline above. But the testing point is reasonable.

dwave/optimization/include/dwave-optimization/array.hpp
```diff
@@ -737,6 +738,27 @@ std::vector<ssize_t> broadcast_shape(const std::span<const ssize_t> lhs,
...
 
+/// For "partial reduction", get the shape of the resulting shape of the array when preforming a
+/// reduction over an axis
+std::vector<ssize_t> partial_reduce_shape(const std::span<const ssize_t> array_shape,
+                                          const ssize_t axis);
+std::vector<ssize_t> partial_reduce_shape(std::initializer_list<ssize_t> input_shape,
+                                          const ssize_t axis);
+
+/// Gets the strides of a n-dimensional array assuming contiguous memory
+std::vector<ssize_t> as_contiguous_strides(const std::span<const ssize_t> shape);
```
arcondello, line 749: This duplicates the functionality of the existing [shape_to_strides()](https://github.com/dwavesystems/dwave-optimization/blob/a4e00cf2c24caec9eeab129a73c773da307af58e/dwave/optimization/include/dwave-optimization/array.hpp#L645) static method.

dwave/optimization/src/nodes/mathematical.cpp
alexzucca90: so here you are saying something like
```
ssize_t p_index =
                start_idx + i *array_ptr_->strides()[this->axis_] / array_ptr_->itemsize();
        assert(p_index < static_cast<ssize_t>(array_ptr_->size(state)));
        double other = array_ptr_->buff(state)[p_index];
```
if I understand correctly?

arcondello: I think I have a snippet. Putting it together now.

alexzucca90: Update this now, let me know if it looks good

alexzucca90:
Ok I tried to make the code ready to support multiple axes in the future. I pushed single axis assumption behind the compilation barrier. Can you check the serialization/deserialization?

dwave/optimization/symbols.pyx
alexzucca90: I think this is now addressed.

dwave/optimization/include/dwave-optimization/array.hpp
alexzucca90: I moved this behind compilation barrier in `mathematical.cpp`.

dwave/optimization/src/array.cpp
```diff
@@ -246,4 +246,34 @@ std::vector<ssize_t> broadcast_shape(std::initializer_list<ssize_t> lhs,
     return broadcast_shape(std::span(lhs), std::span(rhs));
 }
 
+std::vector<ssize_t> unravel_index(const std::span<const ssize_t> strides, ssize_t index) {
+    ssize_t ndim = static_cast<ssize_t>(strides.size());
+    std::vector<ssize_t> indices;
+    indices.reserve(ndim);
+
+    for (auto stride : strides) {
```
arcondello, line 254:
```suggestion
    for (const auto& stride : strides) {
```

dwave/optimization/src/array.cpp
```diff
@@ -246,4 +246,34 @@ std::vector<ssize_t> broadcast_shape(std::initializer_list<ssize_t> lhs,
...
+        indices.push_back(index / (stride / sizeof(double)));
+        index = index % (stride / sizeof(double));
+    }
+    return indices;
+}
+
+ssize_t ravel_multi_index(const std::span<const ssize_t> strides,
+                          const std::span<const ssize_t> indices) {
+    ssize_t ndim = static_cast<ssize_t>(strides.size());
```
arcondello, line 263:
```suggestion
    const ssize_t ndim = static_cast<ssize_t>(strides.size());
```

dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -663,6 +663,322 @@ template class NaryOpNode<functional::min<double>>;
 template class NaryOpNode<std::multiplies<double>>;
 template class NaryOpNode<std::plus<double>>;
 
+// PartialReduceNode *****************************************************************
+std::vector<ssize_t> partial_reduce_shape(const std::span<const ssize_t> input_shape,
+                                          const ssize_t axis) {
+    std::vector<ssize_t> shape;
+    shape.assign(input_shape.begin(), input_shape.end());
```
arcondello, line 670:
```suggestion
    std::vector<ssize_t> shape(input_shape.begin(), input_shape.end());
```

dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -663,6 +663,322 @@ template class NaryOpNode<functional::min<double>>;
...
+    shape.erase(shape.begin() + axis);
+    return shape;
+}
+
+std::vector<ssize_t> as_contiguous_strides(const std::span<const ssize_t> shape) {
+    ssize_t ndim = static_cast<ssize_t>(shape.size());
+
+    assert(ndim >= 0);
```
arcondello, line 678:
```c++
    assert(ndim > 0);
```
right? because we access `strides[ndim - 1]` shortly.

Or, if we want to handle the `ndim == 0` case, we could do
```c++
    assert(ndim >= 0);
    if (ndim < 1) return {};
```

dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -663,6 +663,322 @@ template class NaryOpNode<functional::min<double>>;
...
+        strides[i] = strides[i + 1] * shape[i + 1];
+    }
+    return strides;
+}
+
+/// TODO: support multiple axes
+template <class BinaryOp>
+PartialReduceNode<BinaryOp>::PartialReduceNode(ArrayNode* node_ptr, std::span<const ssize_t> axes,
+                                               double init)
+        : ArrayOutputMixin(partial_reduce_shape(node_ptr->shape(), axes[0])),
```
arcondello, line 692: This will segfault with `axes.size() < 1`.

Unfortunately `std::span::at()` doesn't come in until C++26, otherwise that would be an elegant fix.

I think you'll need to add a
```c++
std::span<const ssize_t> nonempty(std::span<const ssize_t> span) {
    if (span.empty()) throw ...;
    return span;
}
```
function and then use it here
```c++
        : ArrayOutputMixin(partial_reduce_shape(node_ptr->shape(), nonempty(axes)[0])),
```

dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -663,6 +663,322 @@ template class NaryOpNode<functional::min<double>>;
...
+          axes_(make_axes(axes)),
+          parent_strides_c_(as_contiguous_strides(array_ptr_->shape())) {
+    if (array_ptr_->dynamic()) {
+        throw std::invalid_argument("cannot do a partial reduction on a dynamic array");
+    } else if (array_ptr_->size() < 1) {
+        throw std::invalid_argument("cannot do a partial reduction on an empty array");
+    }
+
+    // Validate axes
+    /// TODO: support negative axes
```
arcondello, line 704: We should add an
```c++
axes.size() == 1
```
check here for now. Probably throwing an error rather than as assertion.

dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -663,6 +663,322 @@ template class NaryOpNode<functional::min<double>>;
...
+template <class BinaryOp>
+PartialReduceNode<BinaryOp>::PartialReduceNode(ArrayNode* node_ptr, const ssize_t axis, double init)
+        : PartialReduceNode(node_ptr, {axis}, init) {}
+
+template <>
+PartialReduceNode<std::plus<double>>::PartialReduceNode(ArrayNode* array_ptr, const ssize_t axis)
+        : PartialReduceNode(array_ptr, axis, 0) {}
+
+template <class BinaryOp>
+PartialReduceNode<BinaryOp>::PartialReduceNode(ArrayNode* array_ptr, std::span<const ssize_t> axes)
```
arcondello, line 728: I don't think we want this generic overload, right? It doesn't set up `init` and we already cover the one case that does.

dwave/optimization/model.pyx
```diff
@@ -2049,12 +2049,23 @@ cdef class ArraySymbol(Symbol):
...
+        if axis is not None:
+            if not isinstance(axis, numbers.Integral):
+                raise TypeError("axis of the sum should be an int")
+
+            if not (0 <= axis < self.ndim()):
+                raise ValueError("axis should be 0 <= axis < self.ndim()")
+
+            return PartialSum(self, axis)
+
         return Sum(self)
```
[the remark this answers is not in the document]
alexzucca90: this does not compile, doesn't know what `dwave` is of course..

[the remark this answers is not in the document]
arcondello: Ah, right. You can add `import dwave.optimization.symbols` back at the top of the function.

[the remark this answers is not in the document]
alexzucca90: but then we're importing more, aren't we?

> This pattern works too and doesn't make an unnecessary import.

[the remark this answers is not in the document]
arcondello: the unnecessary bit was adding two symbols to the namespace. The import happens either way.

This is all very much aesthetic stuff. I doubt the performance difference is even noticeable.

dwave/optimization/src/nodes/mathematical.cpp
alexzucca90: so remove that also from the header?

arcondello: So the header file keeps it. But we only implement it for the relevant type overloads.

arcondello (approved):
LGTM

dwave/optimization/include/dwave-optimization/array.hpp
```diff
@@ -23,6 +23,7 @@
 #include <memory>
 #include <numeric>
 #include <optional>
+#include <ranges>
```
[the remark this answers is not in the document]
alexzucca90: I used to, not now removing it

PR #115, opened 2024-09-20
Check for valid indices when indexing arrays

arcondello:


dwave/optimization/src/nodes/indexing.cpp
```diff
@@ -248,22 +248,47 @@ struct AdvancedIndexingNode::IndexParser_ {
...
+                // Test whether we can index with the given array. This is probably a bit
+                // over-strict and may need to be loosened over time.
+                if (!indexing_array_ptr->integral()) {
+                    throw std::invalid_argument("cannot index an array with a non-integral array");
+                }
+                if (indexing_array_ptr->min() < 0) {
+                    throw std::invalid_argument("indexing array cannot allow negative values");
+                }
+                if (idx == 0 && array_ptr->shape()[0] < 0) {
+                    // 100 is a magic number
```
wbernoudy, line 262: This is the maximum graph depth to search for "tightening" the size info?

dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -411,8 +513,38 @@ double NaryOpNode<BinaryOp>::min() const {
...
+        for (const Array* rhs_ptr : operands_ | std::views::drop(1)) {
+            double rhs_low = rhs_ptr->min();
+            double rhs_high = rhs_ptr->max();
+
+            double tmp = std::min({lhs_low * rhs_low, lhs_low * rhs_high, lhs_high * rhs_low,
+                                   lhs_high * rhs_high});
+
+            lhs_high = std::max({lhs_low * rhs_low, lhs_low * rhs_high, lhs_high * rhs_low,
+                                 lhs_high * rhs_high});
+
```
wbernoudy, line 539: I'm not sure this logic always holds, consider for example three operands that have these as the min/max:
```
[(1, 5), (-5, -1), (-1, 0)]
```

Here's how the loop would iterate
```
1. lhs_low, lhs_high = 1, 5
combos = (1 * -5, 1 * -1, 5 * -5, 5 * -1) = (-5, -1, -25, -5)
min(combos), max(combos) = -25, -1

2. lhs_low, lhs_high = -25, -1
combos = (-25 * -1, -25 * 0, -1 * -1, -1 * 0) = (25, 0, 1, 0)
min(combos), max(combos) = 1, 25

final lhs_low, lhs_high = 1, 25
```

So it would return 1, but the actual minimum is -25. Hope I got that right...

dwave/optimization/src/nodes/indexing.cpp
wbernoudy: More like loosening.

arcondello: Yes

dwave/optimization/src/nodes/mathematical.cpp
arcondello: I don't see how to get -25 out of your example. I can get it by using
```
[(1, 5), (-5, -1), (-1, 1)]
```
which does give
```
2. lhs_low, lhs_high = -25, -1
combos = (-25 * -1, -25 * 1, -1 * -1, -1 * 1) = (25, -25, 1, -1)
min(combos), max(combos) = -25, 25

final lhs_low, lhs_high = -25, 25
```

But there may be other counterexamples... Thinking.

wbernoudy: Good catch, I had tried to simplify my example and messed it up in the process. After writing some quick python to play around with this I'm now convinced the logic is sound.

tests/cpp/tests/test_nodes_indexing.cpp
```diff
@@ -874,160 +892,126 @@ TEST_CASE("AdvancedIndexingNode") {
...
-                        arr_ptr->commit(state);
-                        dyn_ptr->commit(state);
-                        i_ptr->commit(state);
-                        j_ptr->commit(state);
-                        k_ptr->commit(state);
-                        adv->commit(state);
+                    AND_WHEN("We commit") {
+                        graph.commit(state, graph.descendants(state, {arr_ptr, dyn_ptr}));
 
-                        arr_ptr->set(state, 47, -1);
```
wbernoudy, line 932: Why are we removing the tests that mutate the main array and check for correctness?

arcondello: We do mutate both, see line 941 above.

arcondello: We do also check for correctness, but mostly rely on the `ArrayValidationNode` for consistency.

wbernoudy: You grow the array, but what about updates standard updates without changing size?

arcondello: Oh, I see. Sure. I can restore that.

tests/cpp/tests/test_nodes_indexing.cpp
```diff
@@ -874,160 +892,126 @@ TEST_CASE("AdvancedIndexingNode") {
...
-                        arr_ptr->commit(state);
-                        dyn_ptr->commit(state);
-                        i_ptr->commit(state);
-                        j_ptr->commit(state);
-                        adv->commit(state);
-                        val->commit(state);
-
-                        std::vector<double> new_expected = expected;
-
-                        arr_ptr->set(state, 47, -1);
```
wbernoudy, line 1008: These kinds of updates (within the captured range and outside of it), after commit, are rather important to test as well. Would rather not see them deleted

arcondello: Added back in in 5b2c1ee

tests/cpp/tests/test_nodes_indexing.cpp
arcondello: Added back in in 5b2c1ee

wbernoudy (approved):

PR #116, opened 2024-09-23
Reorganize C++ mathematical nodes tests

arcondello:
The mathematical nodes were not very consistently organized. So this PR somewhat fixes that. No tests have been added or removed. Purely moves stuff around.

No functional change.

### DIFF
--- a/tests/cpp/tests/test_nodes_mathematical.cpp
+++ b/tests/cpp/tests/test_nodes_mathematical.cpp
@@ -195,7 +195,7 @@ TEMPLATE_TEST_CASE("BinaryOpNode", "", std::equal_to<double>, std::less_equal<do
     }
 }
 
-TEST_CASE("BinaryOpNode") {
+TEST_CASE("BinaryOpNode - LessEqualNode") {
     auto graph = Graph();
 
     GIVEN("A less equal node operating on two continuous arrays") {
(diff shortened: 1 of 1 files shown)


tests/cpp/tests/test_nodes_mathematical.cpp
```diff
@@ -567,59 +456,346 @@ TEMPLATE_TEST_CASE("NaryOpNode", "", functional::max<double>, functional::min<do
...
+        THEN("We cannot construct a reduce node without an initial value") {
+            CHECK_THROWS(graph.emplace_node<MaxNode>(a_ptr));
+            CHECK_THROWS(graph.emplace_node<MinNode>(a_ptr));
+            CHECK(a_ptr->successors().size() == 0);  // no side effects
         }
     }
 }
 
-TEST_CASE("ProdNode") {
+TEST_CASE("ReduceNode - ProdNode") {
```
wbernoudy, line 798: Could add a `[ReduceNode]` tag to these

arcondello: I wanted them visible in the output when there's an error and tags for whatever reason don't show up there. Agree we could add tags, but IMO that's a followup PR for all of the tests - currently they are all untagged.

wbernoudy: Sure! Just noting it would be nice to be able to run all the `ReduceNode` tests together still. Wasn't recommending removing the `ReduceNode` from the main title.

wbernoudy (approved):

PR #118, opened 2024-09-25
Fix whitespace in `intro.rst`

arcondello:
Previously there was a mix of tabs and spaces, but only in a few sections.

Content it unchanged. It renders identically to before.

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -347,8 +347,8 @@ The two tabs below provide the two formulations.
 
         The model in this tab is formulated using compact matrix operations. 
 
-	Instantiate a nonlinear model and add the constant symbols. 
-	
+        Instantiate a nonlinear model and add the constant symbols. 
+
         >>> model = Model()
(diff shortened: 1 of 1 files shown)


docs/intro.rst
```diff
@@ -489,25 +489,25 @@ The two tabs below provide the two formulations.
...
-	14 36
-	
+        >>> model = Model()
+        >>> # Add the constants
+        >>> cost = model.constant(cost_per_day)
+        >>> distances = model.constant(distance_matrix)
+        >>> # Add the decision symbol
+        >>> route = model.list(4)
+        >>> # Optimize the objective
+            >>> model.minimize((cost * distances[route[:-1],route[1:]]).sum())
```
JoelPasvolsky, line 499: This line doesn't look right though it renders okay

JoelPasvolsky (approved):

docs/intro.rst
arcondello: oops, missed that one. Will fix

PR #119, opened 2024-09-25
Feature/return constraint symbol

jlalbers:
Updated the `Model.add_constraint()` method to return the constraint symbol, which can be inspected by the user to determine constraint feasibility. This will allow users to more easily track which constraints are violated if a model state is infeasible compared to using the `Model.iter_constraints()` method. Update summary:

- Add return type to `Model.add_constraint()` method in `model.pyi`.
- Return `value` parameter in `Model.add_constraint()` method in `model.pyx`.
- Update unit tests for the `Model.add_constraint()` method in `tests/test_model.py`.
- Add release note.

No explicit updates to the docs, as the new documentation for the `Model.add_constraint()` method should be built automatically by sphinx with the updated method docstring. However, I can add a blurb about checking constraint feasibility to `docs/intro.rst` if the addition would add value.


docs/intro.rst
```diff
@@ -363,7 +363,7 @@ The two tabs below provide the two formulations.
         capacity.
 
         >>> total_weight = items * weight 
-        >>> model.add_constraint(total_weight.sum() <= capacity)
+        >>> _ = model.add_constraint(total_weight.sum() <= capacity)
```
arcondello, line 366: I am so annoyed by Python that [we cannot just do](https://bugs.python.org/issue32509)
```python
>>> model.add_constraint(total_weight.sum() <= capacity)  # doctest: +ELLIPSIS
...
```

IMO
```python
>>> model.add_constraint(total_weight.sum() <= capacity)  # doctest: +ELLIPSIS
<dwave.optimization.symbols.LessEqual at ...>
```
is clearer.

@JoelPasvolsky, any opinions?

dwave/optimization/model.pyx
```diff
@@ -92,19 +92,36 @@ cdef class Model:
...
 
             >>> from dwave.optimization.model import Model
             >>> model = Model()
             >>> i = model.integer()
             >>> c = model.constant(5)
-            >>> model.add_constraint(i <= c)
+            >>> constraint_sym = model.add_constraint(i <= c)
+
+            The returned constraint symbol can be assigned and evaluated
+                for a model state:
```
arcondello, line 108:
```suggestion
            for a model state:
```

jlalbers:
One interesting thing I noticed is the following behaviour:
```python
...
>>> i = model.integer()
>>> c = model.constant(5)
>>> c_sym = model.add_constraint(i <= c)
>>> c_iter, = model.iter_constraints()
>>> id(c_sym) == id(c_iter)
False
>>> c_sym.__repr__()
<dwave.optimization.symbols.LessEqual at 0x0123...>
>>> c_iter.__repr__()
<dwave.optimization.symbols.LessEqual at 0x0123...>
>>> c_sym.__repr__() == c_iter.__repr__()
True
>>> c_sym.equals(c_iter)
True
```
I think this is because `iter_constraints()` is constructing a new symbol from the pointer in the constraints vector, so there are actually two distinct Python objects pointing to the same constraint. The `equals` method shows that they are still pointing to the same graph node, but don't know if the differing ids will cause any issue.

arcondello:
Yes, there is a many-to-1 relationship between the Python `Symbol` objects, and the underlying node. You can do
```Python
symbol1.id() == symbol2.id()
```
to test whether they are using the same node. This is part of the design, for better or worse.

docs/intro.rst
jlalbers: I'll update this in the docs

arcondello: Here and elsewhere please.

jlalbers: @arcondello updated

arcondello (approved):
awesome! Thanks

Issue #82, opened 2024-08-09
Consider adding a method to check feasibility


arcondello:
We currently have a [`Graph::feasible()`](https://github.com/dwavesystems/dwave-optimization/blob/6c626f52371eea3419015d0efbaa027255dcef10/dwave/optimization/include/dwave-optimization/graph.hpp#L152) method at the C++ level.

We could expose something similar at the Python level

Something like
```Python
class Model:
    ...

    def feasible(state_index: int) -> bool:
        with self.lock():
            return all(constraint.state(state_index) for constraint in self.iter_constraints())
```

arcondello:
Closed by https://github.com/dwavesystems/dwave-optimization/pull/120

PR #120, opened 2024-09-26
Feature/feasibility

jlalbers:
@arcondello adds a `feasible()` method to the `Model` class to make it easier for users to check the feasibility of a model state without manually checking the state of each constraint symbol. Addresses #82 

Current work will build, but throws a segfault when I try to actually call the method in Python.

### DIFF
--- a/dwave/optimization/model.pyx
+++ b/dwave/optimization/model.pyx
@@ -253,7 +253,27 @@ cdef class Model:
 
         Returns:
             Feasibility of the state.
+
+        Examples:
+            This example demonstrates checking the feasibility of a simple model with
+            feasible and infeasible states.
+
+            >>> from dwave.optimization.model import Model
+            >>> model = Model()
+            >>> b = model.binary()
+            >>> model.add_constraint(b) # doctest: +ELLIPSIS
+            <dwave.optimization.BinaryVariable at ...>
+            >>> model.states.resize(2)
+            >>> b.set_state(0, 1) # Feasible
+            >>> b.set_state(1, 0) # Infeasible
+            >>> with model.lock():
+            ...     model.feasible(0)
+            True
+            >>> with model.lock():
+            ...     model.feasible(1)
+            False
         """
+
         cdef Py_ssize_t num_states = self.states.size()
 
         if not -num_states <= index < num_states:
--- a/dwave/optimization/model.pyx
+++ b/dwave/optimization/model.pyx
@@ -254,7 +254,8 @@ cdef class Model:
         Returns:
             Feasibility of the state.
         """
-        cdef int num_states = self.state_size()
+        cdef Py_ssize_t num_states = self.states.size()
+
         if not -num_states <= index < num_states:
             raise ValueError(f"index out of range: {index}")
         elif index < 0:  # allow negative indexing
--- a/dwave/optimization/model.pyx
+++ b/dwave/optimization/model.pyx
@@ -245,6 +245,23 @@ cdef class Model:
         lists = [DisjointList(main, i) for i in range(num_disjoint_lists)]
         return main, lists
 
+    def feasible(self, int index = 0):
+        """Check the feasibility of the state at the input index.
+
+        Args:
+            index: index of the state to check for feasibility.
+
+        Returns:
+            Feasibility of the state.
+        """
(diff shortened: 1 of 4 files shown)


dwave/optimization/model.pyx
```diff
@@ -228,6 +228,23 @@ cdef class Model:
...
+        Returns:
+            Feasibility of the state.
+        """
+        cdef int num_states = self.state_size()
+        if not -num_states <= index < num_states:
+            raise ValueError(f"index out of range: {index}")
+        elif index < 0:  # allow negative indexing
+            index += num_states
+
+        return self._graph.feasible(self.states._states[index])
```
alexzucca90, line 246: I suspect it seg-faults because you need to initialize the state before calling the feasible method. This should be the cause if in your test, when calling it, you set a state and then called the feasible method right after
So you can add some python tests! :)

jlalbers:
There was a bug when checking the index against the state size (used the `state_size` method instead of `states.size`. However, I'm still getting a segfault when calling `feasible` with index 0 using an initialized state, i.e. the following:
```python
...
>>> i = model.integer()
>>> c = model.constant(5)
>>> model.add_constraint(i <= c)
>>> model.feasible(0)
...
ValueError: index 0 is out of range # Expected behavior
>>> model.states.resize(1)
>>> i.set_state(0, 1) # Feasible state
>>> model.feasible(0)
Process finished with exit code 139 (interrupted by signal 11:SIGSEGV)
```
Could it be something to do with the `State` type alias used by Cython being different from the one used in the `Graph` C++ implementation? Could throw a segfault if the type aliases represent different objects and `Graph::feasible` tries to illegally access a member that ends up being at the wrong memory offset for a different object type

dwave/optimization/model.pyx
alexzucca90: Sorry, I think I was unclear earlier. I meant here
```python
self._graph.initialize_state(self.states._states[index])
return self._graph.feasible(self.states._states[index])
```

jlalbers: Ahh gotcha. Sorry, just my unfamiliarity with the codebase. Anything after initializing the state I'd need to clean up?

alexzucca90: I think that should be it.. there are two things worth remembering here. The C++ method `initialize_state()` initialize all nodes in the graph (eagerly) to their value given the decision variables. If you were to check every single constraint from the python side, then it would be lazily from checking each constraint. Worth keeping this in mind in case one deals with models with very deep graphs on the objective side and simple constraints on the decisions (kinda a worst case scenario)

dwave/optimization/model.pyx
```diff
@@ -228,6 +245,24 @@ cdef class Model:
...
+        Returns:
+            Feasibility of the state.
+        """
+        cdef Py_ssize_t num_states = self.states.size()
+
+        if not -num_states <= index < num_states:
+            raise ValueError(f"index out of range: {index}")
+        elif index < 0:  # allow negative indexing
+            index += num_states
+
```
alexzucca90, line 263: maybe we can also return True if number of constraints is 0

dwave/optimization/model.pyx
jlalbers: I'll try doing some performance analysis and see if there's a significant cost with the eager initialization. Also could test with common problem types (i.e. models created by the generator functions) and see if it makes any difference

dwave/optimization/model.pyx
```diff
@@ -245,6 +245,44 @@ cdef class Model:
...
+        """
+
+        cdef Py_ssize_t num_states = self.states.size()
+
+        if not -num_states <= index < num_states:
+            raise ValueError(f"index out of range: {index}")
+        elif index < 0:  # allow negative indexing
+            index += num_states
+
+        return all(sym.state(index) for sym in self.iter_constraints())
```
arcondello, line 284:
```suggestion
        return all(sym.state(index) for sym in self.iter_constraints())
```
index is checked by each symbol's `.state()` method, so we can just be lazy about it.

arcondello (approved):
LGTM, will merge when/if the CI passes

alexzucca90 (approved):
LGTM

PR #123, opened 2024-09-27
Use catch2 version 3.7.1

arcondello:
The latest version supported by meson wrap as of Sept 2024.

Now that meson is handling install/build of catch2 for us. It makes sense to use the latest. We were on quite an old branch.

Also fix some compiler warnings generated by `IntegerNode`.


.circleci/config.yml
```diff
@@ -97,7 +97,7 @@ jobs:
             . env/bin/activate
             python -m build \
               -Ccompile-args="-v" \
-              -Csetup-args="-Dbuildtype=debug" \
+              -Csetup-args="-Db_ndebug=true" \
```
arcondello, line 100: the `bdebug` build type builds the C++ test, which doesn't work with creating a wheel. So we instead turn on debug symbols directly.

tests/cpp/tests/test_array.cpp
```diff
@@ -14,7 +14,7 @@
 
 #include <sstream>
 
-#include "catch2/catch.hpp"
+#include "catch2/catch_all.hpp"
```
wbernoudy, line 17: Isn't the `catch_all.hpp` header significantly bigger than the original `catch2/catch.hpp` header? Is it slower to compile now?

wbernoudy: I see they recommend doing this replacement in the upgrade docs so I'm sure it's fine. Just curious.

arcondello: Ok. I changed to use more specific headers. Though the compilation time changed not at all so :shrug: .

wbernoudy: Disappointing!

wbernoudy (approved):

PR #124, opened 2024-09-27
Implement `SumNode.min()` and `SumNode.max()` for dynamic predecessors

arcondello:
Doesn't need a release note, because it's covered on the release note already added in https://github.com/dwavesystems/dwave-optimization/pull/115

### DIFF
--- a/dwave/optimization/src/nodes/mathematical.cpp
+++ b/dwave/optimization/src/nodes/mathematical.cpp
@@ -907,11 +907,40 @@ double ReduceNode<BinaryOp>::max() const {
         });
     }
     if constexpr (std::is_same<BinaryOp, std::plus<double>>::value) {
-        // the dynamic case. For now let's just fall back to Array's default
-        // implementation because this gets even more complicated
-        if (array_ptr_->dynamic()) return Array::max();
+        const double high = array_ptr_->max();
+        const double init = this->init.value_or(0);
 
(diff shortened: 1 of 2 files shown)


dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -965,11 +994,40 @@ double ReduceNode<BinaryOp>::min() const {
...
+        }
 
-        return this->init.value_or(0) + array_ptr_->size() * array_ptr_->min();
+        // our predecessor array is dynamic. So there are a few more cases
+        // we need to check
+
+        // 100 is a magic number. It's how far back in the predecessor
+        // chain to check to get good bounds on the size for the given array.
+        // This will exit early if it converges.
+        const SizeInfo sizeinfo = array_ptr_->sizeinfo().substitute(100);
```
wbernoudy, line 1015: Should we consider making this a special constant that we control globally, now that you've have to use it a few times? Unless there is a reason that different nodes would want to limit recursion to different depths.

wbernoudy (approved):

dwave/optimization/src/nodes/mathematical.cpp
arcondello: We should for sure. I had the same thought. Will do so in a followup soon.

PR #132, opened 2024-10-04
Add decrement operator for ArrayIterator

wbernoudy:


arcondello:
We should add another static assert [here](https://github.com/dwavesystems/dwave-optimization/blob/a627daf9d5f3007012df9094ba0dd04aa025e6d2/dwave/optimization/include/dwave-optimization/array.hpp#L470) for [bidirectional_iterator](https://en.cppreference.com/w/cpp/iterator/bidirectional_iterator).

dwave/optimization/include/dwave-optimization/array.hpp
```diff
@@ -413,17 +435,40 @@ class ArrayIterator {
...
+                for (std::ptrdiff_t i = 0; i > n; --i) {
+                    offset += unadvance();
+                }
+            } else {
+                for (std::ptrdiff_t i = 0; i < n; ++i) {
+                    offset += advance();
+                }
             }
             return offset;
         }
```
arcondello, line 474: I think we could also just do
```c++
        std::ptrdiff_t advance(std::ptrdiff_t n) {
            std::ptrdiff_t offset = 0;

            for (; n > 0; --n) offset += unadvance();
            for (; n < 0; ++n) offset += advance();

            return offset;
        }
```
likely after the compiler has at it, it won't make much difference. But I guess this is a few fewer instructions/memory locations if it's taken literally.

arcondello (approved):
LGTM!

dwave/optimization/include/dwave-optimization/array.hpp
wbernoudy: Done, though I had to switch them :stuck_out_tongue:

PR #134, opened 2024-10-07
Bugfix/cvrptw generator

jlalbers:
I was reverse-engineering this generator function while developing a slightly different formulation for this problem type and noticed that the values for some of the intermediate symbols were different than expected using an asymmetric time-distance matrix. Let me know if the modifications make sense or if I'm missing something in the original function and the updates are off-base.

### DIFF
--- a/dwave/optimization/generators.py
+++ b/dwave/optimization/generators.py
@@ -525,10 +525,16 @@ def capacitated_vehicle_routing_with_time_windows(demand: numpy.typing.ArrayLike
     time_distance_ext = np.zeros((num_customers + 1, num_customers + 1))
     time_distance_ext[:num_customers, :num_customers] = time_distances_array[1:, 1:]
 
-    time_depo_ext = np.zeros(num_customers + 1)
-    time_depo_ext[:num_customers] = time_distances_array[0, 1:]
+    # Time from depot to destination
+    time_from_depo_ext = np.zeros(num_customers + 1)
+    time_from_depo_ext[:num_customers] = time_distances_array[0, 1:]
 
-    t_depo = model.constant(time_depo_ext)
+    # Time from source to depot
+    time_to_depo_ext = np.zeros(num_customers + 1)
+    time_to_depo_ext[:num_customers] = time_distances_array[1:, 0]
+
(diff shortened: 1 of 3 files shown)


dwave/optimization/generators.py
```diff
@@ -525,10 +525,16 @@ class as the decision variable being optimized, with permutations of its
...
+    time_from_depo_ext = np.zeros(num_customers + 1)
+    time_from_depo_ext[:num_customers] = time_distances_array[0, 1:]
 
-    t_depo = model.constant(time_depo_ext)
+    # Time from source to depot
+    time_to_depo_ext = np.zeros(num_customers + 1)
+    time_to_depo_ext[:num_customers] = time_distances_array[1:, 0]
+
+    t_from_depo = model.constant(time_from_depo_ext)
+    t_to_depo = model.constant(time_to_depo_ext)
```
arcondello, line 537: I suppose you could do
```python
    if np.equal(time_from_depo_ext, time_to_depo_ext).all():
         t_from_depo, t_to_depo = model.constant(time_from_depo_ext)
    else:
        t_from_depo = model.constant(time_from_depo_ext)
        t_to_depo = model.constant(time_to_depo_ext)
```
but that's probably premature optimization.

arcondello (approved):

dwave/optimization/generators.py
jlalbers: Doesn't hurt to add if it's just a couple lines of difference

arcondello:
Also needs a release note.

elham-azd (approved):
LGTM. Good catch, thank you!

releasenotes/notes/fix-cvrptw-generator-044c7d7b5ad1c112.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    Updates the logic in the `capacitated_vehicle_routing_with_time_windows` generator function
```
arcondello, line 4:
```suggestion
    Update the logic in the ``capacitated_vehicle_routing_with_time_windows()`` generator function
```

PR #138, opened 2024-10-11
Fix two bugs in advanced indexing propagation for non-const non-dynamic arrays with dynamic indexers

wbernoudy:
First one was just a simple logic error in which method should be called when reverting the reverse offset map. Second one was basically covered by an old TODO: to ensure all updates to the parent array are reflected in the final advanced indexing output, we need to track the offset indices that were modified (not just the offsets themselves) as new offsets may be added that duplicate previous unchanged offsets.

### DIFF
--- a/dwave/optimization/src/nodes/indexing.cpp
+++ b/dwave/optimization/src/nodes/indexing.cpp
@@ -112,8 +112,8 @@ struct AdvancedIndexingNodeData : NodeStateData {
         for (const auto& update : offsets_diff | std::views::reverse) {
             offsets_[update.index] = update.old;
 
-            if (!update.placed()) delete_from_reverse(update.index, update.value);
-            if (!update.removed()) add_to_reverse(update.index, update.old);
+            if (!update.removed()) delete_from_reverse(update.index, update.value);
+            if (!update.placed()) add_to_reverse(update.index, update.old);
         }
 
         for (auto& [index, old, _] : diff | std::views::reverse) {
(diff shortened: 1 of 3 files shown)


tests/cpp/tests/test_nodes_indexing.cpp
```diff
@@ -1081,38 +1081,113 @@ TEST_CASE("AdvancedIndexingNode") {
                         verify_array_diff({}, {3, 5, 7, 2}, adv->diff(state));
                     }
 
-                    AND_WHEN("We mutate the main array") {
+                    AND_WHEN("We revert the indexing nodes") {
```
arcondello, line 1084: Do we want to toss an `ArrayValidationNode` on `adv` while we're at it?

dwave/optimization/src/nodes/indexing.cpp
```diff
@@ -816,16 +814,15 @@ void AdvancedIndexingNode::propagate(State& state) const {
...
-                    for (const auto& idx : *idxs) {
-                        ssize_t item_stride = subspace_stride_ / itemsize();
-                        ssize_t new_index = idx * item_stride + mapped_index.second;
-                        assert(data[new_index] == update.old);
-                        diff.emplace_back(new_index, update.old, update.value);
-                        data[new_index] = update.value;
-                    }
+            const auto* idxs = node_data->get_offset_idxs(mapped_index.first);
+            if (idxs != nullptr) {
+                for (const auto& idx : *idxs) {
```
arcondello, line 819: NIT: I know this is what was there before. But I find `auto` in this context less helpful than just stating the type (I think `ssize_t`?). Also could add a couple `const` below. Happy enough to leave it alone as well.

arcondello (approved):
LGTM

dwave/optimization/src/nodes/indexing.cpp
wbernoudy: I agree, happy to change it. What are you looking at to add `const` to though?

arcondello: `item_stride`, `new_index` etc.

arcondello:
Also needs a release note :slightly_smiling_face:

tests/cpp/tests/test_nodes_indexing.cpp
wbernoudy: Yeah, might as well. I can remove the `verify_array_diff()` calls as well.

PR #140, opened 2024-10-15
Update `UnaryOpNode` to correctly handle dynamic predecessors

arcondello:
Previously it would accept them but then segfault.

### DIFF
--- a/dwave/optimization/src/nodes/_state.hpp
+++ b/dwave/optimization/src/nodes/_state.hpp
@@ -87,6 +87,14 @@ class ArrayNodeStateData : public NodeStateData {
 
     std::span<const Update> diff() const noexcept { return updates; }
 
+    // Append a new value to the buffer, tracking the addition in the diff
+    // Return whether a change was made.
+    bool emplace_back(double value) {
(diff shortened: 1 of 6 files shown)


dwave/optimization/src/nodes/_state.hpp
```diff
@@ -87,6 +87,14 @@ class ArrayNodeStateData : public NodeStateData {
 
     std::span<const Update> diff() const noexcept { return updates; }
 
+    // Append a new value to the buffer, tracking the addition in the diff
+    // Return whether a change was made.
+    bool emplace_back(double value) {
```
wbernoudy, line 92: This is for future implementations of `emplace_back` (or subclasses) that might do something smarter and only modify the buffer sometimes?

arcondello: The return value you mean? It's for consistency with the others. Honestly it's inherited from another method, we could probably remove them all.... but that would be out of scope for this PR I think.

wbernoudy: Oops, totally missed this is subclassing `NodeStateData`. Got it.

wbernoudy (approved):

PR #142, opened 2024-10-16
Note maximum state size at construction

JoelPasvolsky:


dwave/optimization/model.pyx
```diff
@@ -1614,6 +1614,11 @@ cdef class Symbol:
     def state_size(self):
         """Return an estimated size, in bytes, of the node's state.
 
+        The state size, set at construction of the symbol, accounts for the
+        largest possible size for symbols such as the successor symbols of a
+        :class:`~dwave.optimization.symbols.DisjointLists` symbol, where each
+        can range from zero to all the elements.
+
```
arcondello, line 1621:
```suggestion
        """Return an estimated size, in bytes, of the symbol's state.

        The state size, set at construction of the symbol, accounts for the
        largest possible size for the symbol.
        
        The size of an instantiated state can be access via
        
        >>> symbol.state().size()  # doctest: +SKIP
```
this is the method for all `Symbol`s, not just `DisjointLists`. So we should make it generic.

arcondello (approved):

PR #143, opened 2024-10-16
Split up some large `.hpp`/`.cpp` files

arcondello:
No functional changes.

This will speed up repeating compiling in some cases and overall make things a bit less unwieldy.

The `manipulation.hpp` name was chosen to match NumPy's [array manipulation docs](https://numpy.org/doc/2.0/reference/routines.array-manipulation.html).


arcondello:
> Why not more subdirectories for the tests, e.g. tests/test_nodes/mathematical/binaryop.cpp?

That's a good idea. The reason _was_ because the way the makefile did discovery, but that's not an issue anymore! Will make the change.

arcondello:
Though now I am half tempted to put them in the source tree... maybe later. For now I've added nesting in d6d2b5b

wbernoudy (approved):

PR #146, opened 2024-10-24
Fix the type information for `ArraySymbol.__abs__()`

arcondello:
Previously `mypy` would erroneously say that you cannot do `abs(symbol)` leading to confusion for users.


thisac (approved):
LGTM

PR #150, opened 2024-10-25
Add `compress` kwarg to `Model.to_file()` and `States.to_file()`

arcondello:
By default this maintains the existing behavior but allows us to toggle it via a keyword argument.

See also https://github.com/dwavesystems/dimod/pull/1296

### DIFF
--- a/dwave/optimization/model.pyx
+++ b/dwave/optimization/model.pyx
@@ -427,6 +427,7 @@ cdef class Model:
     def into_file(self, file, *,
                   Py_ssize_t max_num_states = 0,
                   bool only_decision = False,
+                  bool compress = False,
                   ):
         """Serialize the model into an existing file.
 
@@ -442,6 +443,13 @@ cdef class Model:
             only_decision:
                 If ``True``, only decision variables are serialized.
                 If ``False``, all symbols are serialized.
+            compress:
+                If ``True``, the data will be compressed with
+                :py:const:`zipfile.ZIP_DEFLATED`.
+
+        .. versionadded:: 0.4.1
+           The ``compress`` keyword argument compresses the file using
+           :py:const:`zipfile.ZIP_DEFLATED`.
 
         See also:
             :meth:`.from_file`, :meth:`.to_file`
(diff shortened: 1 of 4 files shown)


dwave/optimization/model.pyx
```diff
@@ -427,6 +427,7 @@ cdef class Model:
     def into_file(self, file, *,
                   Py_ssize_t max_num_states = 0,
                   bool only_decision = False,
+                  bool compress = False,
```
randomir, line 430: How about we expose compression level as well?

randomir (approved):
LGTM.

It would be nice to expose some compression level controls as well, or at the very least to specify explicitly what's used (`zlib.Z_DEFAULT_COMPRESSION`/6).

dwave/optimization/model.pyx
arcondello: I thought about that, but erred on consistency with https://github.com/dwavesystems/dimod/pull/1296 and the others for compatibility. No one has requested finer control for those so :shrug:

randomir: Makes sense. I can see pros and cons of both approaches.

randomir: Looks like someone already [did the analysis](https://github.com/dwavesystems/dimod/pull/1296#discussion_r1029893021). :facepalm:

arcondello:
On the one hand, this is a pretty harmless PR, on the other I am underwhelmed with the results. A simple benchmark
```
10000 binary variables in a single array with 25 states, uncompressed
serialization time: 0.006702525999571662s
file size in bytes: 2007953 bytes
10000 binary variables in a single array with 25 states, compressed
serialization time: 0.06300492599984864s
file size in bytes: 83618 bytes
10000 binary variables as 10000 scalars with 25 states, uncompressed
serialization time: 9.95059110299917s
file size in bytes: 74242127 bytes
10000 binary variables in a single array with 25 states, compressed
serialization time: 13.670800387000781s
file size in bytes: 57585889 bytes
```

arcondello:
Closing this as stale for now. It's not a bad idea, but perhaps needs more thinking before we actually merge it.

PR #152, opened 2024-10-28
Test several gcc versions in CI

arcondello:


arcondello:
Ugh, I thought it was https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c12-dont-make-data-members-const-or-references-in-a-copyable-or-movable-type, but apparently not.... still looking.

arcondello:
It's also not https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c43-ensure-that-a-copyable-class-has-a-default-constructor

arcondello:
Ok, finally got it. It had nothing to do with move constructing really. The issue was the the `MaskInfo` owned the fill value which we returned a reference to in some cases. Subsequent moves might cause it to dangle. Not really sure why it was showing up as "uninitialized", but :shrug: .

I'll make a new PR to clean all this up. Some of the copyable/moveable stuff is probably worth keeping as well for performance.

arcondello:
https://github.com/dwavesystems/dwave-optimization/pull/154 is the first followup PR, just adding the gcc tests in CI for now.

arcondello:
https://github.com/dwavesystems/dwave-optimization/pull/155 has the rest of it

PR #153, opened 2024-10-29
Add `ConcatenateNode`

samuelbodin:
Draft PR.
Some initial work for ConcatenateNode. 

Includes
 - New class `ConcatenateNode`
 - New helper/utility function `cartesian_product`
 - 2 test cases for shape, ndim, and buffer initialization


dwave/optimization/include/dwave-optimization/c++23/cartesian_product.hpp
```diff
@@ -0,0 +1,23 @@
...
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#pragma once
+
+#include <vector>
+
+namespace dwave::optimization {
+
+    std::vector<std::vector<ssize_t>> cartesian_product(std::vector<std::vector<ssize_t>> v);
```
arcondello, line 21: I'd put this in [utils.hpp](https://github.com/dwavesystems/dwave-optimization/blob/main/dwave/optimization/include/dwave-optimization/utils.hpp) rather than a new folder. And just note that it's a backport. [Like we do with `unreachable()`](https://github.com/dwavesystems/dwave-optimization/blob/d27a6385925c9fde5fddf011750da28a2b873e8e/dwave/optimization/include/dwave-optimization/utils.hpp#L32).

dwave/optimization/include/dwave-optimization/nodes/indexing.hpp
```diff
@@ -254,6 +254,23 @@ class BasicIndexingNode : public ArrayNode {
     const bool contiguous_;
 };
 
+class ConcatenateNode : public ArrayOutputMixin<ArrayNode> {
```
arcondello, line 257: I'd put this in the new [manipulation.hpp](https://github.com/dwavesystems/dwave-optimization/blob/main/dwave/optimization/include/dwave-optimization/nodes/manipulation.hpp).

dwave/optimization/src/c++23/cartesian_product.cpp
```diff
@@ -0,0 +1,43 @@
...
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#include "dwave-optimization/c++23/cartesian_product.hpp"
+
+#include <numeric>
+#include <iostream>
```
arcondello, line 18: I don't think this gets used?

dwave/optimization/src/c++23/cartesian_product.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+namespace dwave::optimization {
+
+// Based on https://stackoverflow.com/a/31169617
+// Since std::cartesian_product is only specified for c++23 we
+// add our own implementation of it.
+std::vector<std::vector<ssize_t>> cartesian_product(std::vector<std::vector<ssize_t>> v) {
```
arcondello, line 25: It looks like we don't mutate `v`? So probably this should accept a reference so we don't do a copy.

dwave/optimization/src/nodes/indexing.cpp
```diff
@@ -1596,6 +1597,156 @@ std::span<const ssize_t> BasicIndexingNode::shape(const State& state) const {
...
 
+// ConcatenateNode ************************************************************
+
+std::vector<ssize_t> make_concatenate_shape(std::vector<ArrayNode*> array_ptrs, ssize_t axis) {
+    // One or more arrays must be given
+    if (array_ptrs.size() < 1) {
+        throw std::invalid_argument("need at least one array to concatenate");
+    }
+
+    for (auto it = std::next(array_ptrs.begin()); it != array_ptrs.end(); it++) {
```
arcondello, line 1608:
```c++
    for (auto it = std::next(array_ptrs.begin()), stop = array_ptrs.end(); it != stop; it++) {
```
see https://llvm.org/docs/CodingStandards.html#don-t-evaluate-end-every-time-through-a-loop

also applies to a bunch of other places in the code.

dwave/optimization/src/nodes/indexing.cpp
```diff
@@ -1596,6 +1597,156 @@ std::span<const ssize_t> BasicIndexingNode::shape(const State& state) const {
...
+                }
+            }
+        }
+    }
+
+    state[index] = std::make_unique<ArrayNodeStateData>(std::move(values));
+}
+
+void ConcatenateNode::commit(State& state) const {
+
```
arcondello, line 1739: I think this can just be `data_ptr<ArrayNodeStateData>(state)->commit();` That's one of the benefits to using `ArrayNodeStateData`, it handles committing/reverting.

dwave/optimization/src/nodes/indexing.cpp
```diff
@@ -1596,6 +1597,156 @@ std::span<const ssize_t> BasicIndexingNode::shape(const State& state) const {
...
+}
+
+void ConcatenateNode::revert(State& state) const {
+
```
arcondello, line 1743: I think this can just be `data_ptr<ArrayNodeStateData>(state)->revert();` That's one of the benefits to using `ArrayNodeStateData`, it handles committing/reverting.

arcondello (commented):
First quick pass for formatting/organization. Will do a logic pass next

dwave/optimization/include/dwave-optimization/nodes/indexing.hpp
samuelbodin: Oh right, I forgot to move it after rebasing. Will fix!

dwave/optimization/src/c++23/cartesian_product.cpp
samuelbodin: Yeah, reason why I removed the reference was because I thought about having the option to pass rvalues. Can definitely have it as a reference in this case

dwave/optimization/src/nodes/indexing.cpp
samuelbodin: Im surprised issue 1 isn't optimized away by the compiler. 
Issue 2 makes perfect sense, will fix.

dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -18,6 +18,154 @@
...
+
+    // On the concatenation axis we sum the axis dimension sizes
+    for (auto it = std::next(array_ptrs.begin()), stop = array_ptrs.end(); it != stop; it++) {
+       shape[axis] = shape[axis] + (*it)->shape()[axis];
+    }
+
+    return shape;
+}
+
+ConcatenateNode::ConcatenateNode(std::vector<ArrayNode*> array_ptrs, const ssize_t axis)
```
arcondello, line 90: We should use `span` here. It ends up being more flexible by allowing other range types. It also avoids a copy. IMO it also does a better job of signalling to the user/reader what the intent is.
```c++
ConcatenateNode::ConcatenateNode(std::span<const ArrayNode*> array_ptrs, const ssize_t axis)
```
note the `const` is  important! Because it allows the span to be constructed from a rvalue vector.
```c++
// this compiles
std::vector<ssize_t> make_shape(std::span<const ssize_t> shape) {
    return std::vector<ssize_t>(shape.begin(), shape.end());
}

make_shape(std::vector<ssize_t>{0, 1, 2});
```
```c++
// this doesn't
std::vector<ssize_t> make_shape(std::span<ssize_t> shape) {
    return std::vector<ssize_t>(shape.begin(), shape.end());
}

make_shape(std::vector<ssize_t>{0, 1, 2});
```

dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -18,6 +18,154 @@
 
 namespace dwave::optimization {
 
+std::vector<ssize_t> make_concatenate_shape(std::vector<ArrayNode*> array_ptrs, ssize_t axis) {
```
arcondello, line 21: We should use `span` here. It ends up being more flexible by allowing other range types. It also avoids a copy. IMO it also does a better job of signalling to the user/reader what the intent is.
```c++
ConcatenateNode::ConcatenateNode(std::span<const ArrayNode*> array_ptrs, const ssize_t axis)
```

dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -18,6 +18,154 @@
...
+    // One or more arrays must be given
+    if (array_ptrs.size() < 1) {
+        throw std::invalid_argument("need at least one array to concatenate");
+    }
+
+    for (auto it = std::next(array_ptrs.begin()), stop = array_ptrs.end(); it != stop; it++) {
```
arcondello, line 27:
```c++
    for (auto it = std::next(array_ptrs.begin()), stop = array_ptrs.end(); it != stop; ++it) {
```
almost certainly doesn't matter in this case, but [preincrement is a good habit](https://llvm.org/docs/CodingStandards.html#prefer-preincrement). Here and elsewhere.

dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -18,6 +18,154 @@
...
+                            std::to_string((*it)->shape()[i]));
+                }
+            }
+        }
+    }
+
+    // Axis must be in range 0..ndim-1
+    // We can do this check on the first input array since
+    // we at this point know they all have the same shape
+    if (!(0 <= axis && axis < array_ptrs.at(0)->ndim())) {
```
arcondello, line 66: Performance doesn't really matter in this code, but we've already checked that `array_ptrs.size() > 0` so we should just use `[0]`. Or, if we want to be even more explicit, we can use `array_ptrs.front()->ndim()`

dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -18,6 +18,154 @@
...
+                std::to_string(array_ptrs.at(0)->ndim()));
+    }
+
+
+    // The shape of the input arrays, which will be the
+    // same except for possibly on the concatenation axis
+    std::span<const ssize_t> shape0 = array_ptrs.at(0)->shape();
+    std::vector<ssize_t> shape;
+    shape.reserve(shape0.size());
+    std::copy(shape0.begin(), shape0.end(), std::back_inserter(shape));
```
arcondello, line 80: Can just do
```c++
    std::vector<ssize_t> shape(shape0.begin(), shape0.end());
```
because span returns a random access iterator the constructor should be smart enough to to the reserve etc.

dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -18,6 +18,153 @@
...
+    }
+
+    return shape;
+}
+
+ConcatenateNode::ConcatenateNode(std::vector<ArrayNode*> array_ptrs, const ssize_t axis)
+        : ArrayOutputMixin(make_concatenate_shape(array_ptrs, axis)) {
+
+    axis_ = axis;
+    array_ptrs_ = array_ptrs;
```
arcondello, line 94: Why not put these in the initializer list?

dwave/optimization/src/utils.cpp
```diff
@@ -82,4 +82,25 @@ bool is_integer(const double& value) {
...
+    out.reserve(N);
+    for (long long n = 0; n < N; ++n) {
+        lldiv_t q { n, 0 }; // Initial values for quotient and remainder
+        std::vector<ssize_t> u;
+        u.resize(v.size());
+        for (long long i = v.size()-1; 0 <= i; --i) {
+            q = div(q.quot, v[i].size());
+            u[i] = (v[i][q.rem]);
+        }
+        out.push_back(u);
```
arcondello, line 101: The compiler will probably catch it, but we can be sure to avoid a copy with
```c++
        out.emplace_back(std::move(u));
```

arcondello (commented):
Logic for `initialize_state()` looks good to me, next up is propagation!

dwave/optimization/src/nodes/manipulation.cpp
samuelbodin: fixed

dwave/optimization/src/nodes/manipulation.cpp
samuelbodin: fixed

dwave/optimization/src/nodes/manipulation.cpp
samuelbodin: fixed

dwave/optimization/src/nodes/manipulation.cpp
samuelbodin: fixed

dwave/optimization/src/utils.cpp
samuelbodin: fixed

dwave/optimization/src/nodes/manipulation.cpp
samuelbodin: fixed

dwave/optimization/src/nodes/manipulation.cpp
samuelbodin: fixed

dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -18,6 +18,163 @@
...
+    auto ptr = data_ptr<ArrayNodeStateData>(state);
+    ssize_t offset = 0;
+
+    for (auto it = array_ptrs_.begin(), stop = array_ptrs_.end(); it != stop; ++it) {
+        if (std::distance(array_ptrs_.begin(), it) > 0) {
+            offset += (*std::prev(it))->shape()[axis_];
+        }
+
+        for (auto diff : (*it)->diff(state)) {
+            auto indices = unravel_index((*it)->strides(), diff.index);
```
wbernoudy, line 170: I mentioned in our chat that this should be correct, but looking at it again, I'm questioning myself. The reason is that I believe the update index (`diff.index`) is actually a linear index assuming "standard" strides of the originating node (we should really document this better @arcondello). Thus, using the `strides()` of the predecessor may cause this computed multi-index to be wrong if the predecessor has non-standard strides.

To me this adds more weight that we should think about requiring array nodes to provide a `standardized_strides()` or something that returns the strides based only the shape. Otherwise, I think you'll have to compute the "standard" strides for each predecessor to be able to use `unravel_index()` here.

samuelbodin: I see, do you have an example of non standard strides?

arcondello: I think @wbernoudy means contiguous strides rather than standard. I.e. what would the strides be for a contiguous array.

You can make a non-contiguous array easily with a `BasicIndexingNode`. Something like
```c++
graph.emplace_node<BasicIndexingNode>(x_ptr, Slice(std::nullopt, std::nullopt, 2))
```
this is the same as
```python
x[::2]
```
in Python.

arcondello:
Taking a look at [NumPy's implementation](https://github.com/numpy/numpy/blob/fa111b44cfeea4694b38b48f3dbed871a4becc5c/numpy/_core/src/multiarray/multiarraymodule.c#L381), I think it might be simpler to keep a `start`  for each concatenated array. E.g. for
```
[[ 0 1 ]  [[ 4 5 ]  [[ 8 ]
 [ 2 3 ]]  [ 6 7 ]]  [ 9 ]]
```
concatenated on axis 1 the we get
```
strides = [40, 8]  // strides of the resulting array
array0_start = 0
array1_start = 16
array2_start = 32
```
Or, for axis 0
```
[[ 0 1 ]
 [ 2 3 ]]
[[ 0 1 ]
 [ 2 3 ]]
[[ 0 1 ]]
```
```
strides = [16, 8]  // strides of the resulting array
array0_start = 0
array1_start = 32
array2_start = 64
```

Then setting the initial state looks like
```c++
std::vector<ssize_t> strides_into = {40, 8};
std::vector<ssize_t> array_starts = {0, 16, 32};

double* buff;  // allocate to the size of the array we're moving into 

for (ssize_t i = 0; i < 3; ++i) {
    // create a view into our buffer with the same shape as our input array starting at the correct place
    auto view_it = ArrayIterator(buff + array_starts[i], this->ndim(), array_ptrs[i].shape().data(),  this->strides().data());

    // now just copy that data
    std::copy(array_ptrs[i]->begin(state), array_ptrs[i]->end(state), view_it);
}
```

Updates work similarly, but instead of copying you just do `view_it + update.index;`

arcondello:
With the minor issue that we don't actually let you use `ArrayIterator`s to modify, as they all return `const` :laughing:. But we can change that.

dwave/optimization/src/nodes/manipulation.cpp
wbernoudy: Yes, thanks, I should have been saying non-contiguous instead of non-standard.

arcondello:
https://github.com/dwavesystems/dwave-optimization/pull/162 has a mutable iterator implementation.

samuelbodin:
> Taking a look at [NumPy's implementation](https://github.com/numpy/numpy/blob/fa111b44cfeea4694b38b48f3dbed871a4becc5c/numpy/_core/src/multiarray/multiarraymodule.c#L381), I think it might be simpler to keep a `start` for each concatenated array. E.g. for
> 
> ```
> [[ 0 1 ]  [[ 4 5 ]  [[ 8 ]
>  [ 2 3 ]]  [ 6 7 ]]  [ 9 ]]
> ```
> 
> concatenated on axis 1 the we get
> 
> ```
> strides = [40, 8]  // strides of the resulting array
> array0_start = 0
> array1_start = 16
> array2_start = 32
> ```
> 
> Or, for axis 0
> 
> ```
> [[ 0 1 ]
>  [ 2 3 ]]
> [[ 0 1 ]
>  [ 2 3 ]]
> [[ 0 1 ]]
> ```
> 
> ```
> strides = [16, 8]  // strides of the resulting array
> array0_start = 0
> array1_start = 32
> array2_start = 64
> ```
> 
> Then setting the initial state looks like
> 
> ```c++
> std::vector<ssize_t> strides_into = {40, 8};
> std::vector<ssize_t> array_starts = {0, 16, 32};
> 
> double* buff;  // allocate to the size of the array we're moving into 
> 
> for (ssize_t i = 0; i < 3; ++i) {
>     // create a view into our buffer with the same shape as our input array starting at the correct place
>     auto view_it = ArrayIterator(buff + array_starts[i], this->ndim(), array_ptrs[i].shape().data(),  this->strides().data());
> 
>     // now just copy that data
>     std::copy(array_ptrs[i]->begin(state), array_ptrs[i]->end(state), view_it);
> }
> ```
> 
> Updates work similarly, but instead of copying you just do `view_it + update.index;`

Cool! I'll get started on it

Where/how is `strides_into` being used?

arcondello:
Oops, `strides_to` is the same as `this->strides()` in my example above. Typo

arcondello:
This is the same idea as above, but I wrote a snippet that's hopefully a bit clearer for the view idea
```python
import numpy as np

# Get 3 arrays
A = np.asarray([[0, 1], [2, 3]])
B = np.asarray([[4, 5], [6, 7]])
C = np.asarray([[8], [9]])

# Made a concatenated array over axis 1
D = np.concatenate((A, B, C), axis=1)

# Now, recover their views in the D array
A_view = D[:, 0:2]
B_view = D[:, 2:4]
C_view = D[:, 4:5]


# Edit the same entry in B and B_View
B[1, 1] = 100
B_view[1, 1] = 100

print(B)
print(D)
```

dwave/optimization/src/utils.cpp
```diff
@@ -82,4 +82,25 @@ bool is_integer(const double& value) {
...
+// Based on https://stackoverflow.com/a/31169617
+// Since std::cartesian_product is only specified for c++23 we
+// add our own implementation of it.
+std::vector<std::vector<ssize_t>> cartesian_product(std::vector<std::vector<ssize_t>> &v) {
+    std::vector<std::vector<ssize_t>> out;
+    auto product = []( long long a, std::vector<ssize_t>& b ) { return a * b.size(); };
+    const long long N = std::accumulate(v.begin(), v.end(), 1LL, product);
+    out.reserve(N);
+    for (long long n = 0; n < N; ++n) {
+        lldiv_t q { n, 0 }; // Initial values for quotient and remainder
```
arcondello, line 94: irritatingly, the order of `quot` and `rem` is not defined in the standard, see https://en.cppreference.com/w/cpp/numeric/math/div

samuelbodin: Fixed!

dwave/optimization/src/nodes/manipulation.cpp
samuelbodin: Should be fixed now with the use of random access ArrayIterator :)

docs/reference/symbols.rst
```diff
@@ -47,6 +47,7 @@ are inherited by the :ref:`model symbols <symbols_model_symbols>`.
 
     ~ArraySymbol.all
     ~ArraySymbol.any
+    ~ArraySymbol.concatenate
```
arcondello, line 50: I think we want to add it as a function in the `mathematical.py` rather than as a method. Function should have a similar signature to [np.concatenate()](https://numpy.org/doc/stable/reference/generated/numpy.concatenate.html).

dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -18,6 +18,150 @@
...
+                            array_ptrs_[arr_i]->shape().data(),
+                            this->strides().data());
+
+        std::copy(array_ptrs_[arr_i]->begin(state), array_ptrs_[arr_i]->end(state), view_it);
+    }
+
+    state[index] = std::make_unique<ArrayNodeStateData>(std::move(values));
+}
+
+void ConcatenateNode::commit(State& state) const {
```
arcondello, line 137: This is an aesthetic thing, and won't block the PR, but I prefer the methods to be in alphabetical order. I know some of the files don't have that - it's legacy code that will be updated at some later point.

dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -18,6 +18,150 @@
 
 namespace dwave::optimization {
 
+std::vector<ssize_t> make_concatenate_shape(std::span<ArrayNode*> array_ptrs, ssize_t axis) {
+    // One or more arrays must be given
+    if (array_ptrs.size() < 1) {
```
arcondello, line 23: fwiw `NumPy` handles this case
```python
In [2]: np.concatenate(([0, 1], [2, 3]))
Out[2]: array([0, 1, 2, 3])

In [3]: np.concatenate(([0, 1],))
Out[3]: array([0, 1])
```
so either we should support it at the C++ level, or the Python level. Python support could be
```python
def concatenate(array_likes, axis=0):
    if len(array_likes) == 1:
        return array_likes[0]
```
or C++ support would effectively make `ConcatenateNode` a `CopyNode` (https://github.com/dwavesystems/dwave-optimization/issues/16).

dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -18,6 +18,150 @@
...
+
+    // On the concatenation axis we sum the axis dimension sizes
+    for (auto it = std::next(array_ptrs.begin()), stop = array_ptrs.end(); it != stop; ++it) {
+       shape[axis] = shape[axis] + (*it)->shape()[axis];
+    }
+
+    return shape;
+}
+
+ConcatenateNode::ConcatenateNode(std::span<ArrayNode*> array_ptrs, const ssize_t axis)
```
arcondello, line 88: We should add an explicit error somewhere that all arrays are not dynamic. That is `array_ptr->size() >= 0`.

dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -18,6 +18,150 @@
...
+
+    for (ssize_t arr_i = 0, stop = array_ptrs_.size(); arr_i < stop; ++arr_i) {
+        auto view_it = ArrayIterator(
+                            ptr->buff() + array_starts_[arr_i],
+                            this->ndim(),
+                            array_ptrs_[arr_i]->shape().data(),
+                            this->strides().data());
+
+        for (auto diff : array_ptrs_[arr_i]->diff(state)) {
+            auto update_it = view_it + diff.index;
```
arcondello, line 156: A bit of future-proofing would be
```c++
assert(!diff.placed());
assert(!diff.removed());
```
as we don't currently handle that case.

arcondello (commented):
Looks great! Just a few things left.

dwave/optimization/src/nodes/manipulation.cpp
samuelbodin: Not sure I follow, does `array_ptr->size() < 0` mean the array is dynamic?

docs/reference/symbols.rst
samuelbodin: fixed!

dwave/optimization/src/nodes/manipulation.cpp
samuelbodin: Fixed so most of the methods are now ordered alphabetically. 
Is there a way to forward declare the `make_concatenate_shape`? If not it's forced to be defined before the constructors

dwave/optimization/src/nodes/manipulation.cpp
samuelbodin: Now handled in the python, see [[Add Concatenate to mathematical](https://github.com/dwavesystems/dwave-optimization/pull/153/commits/cb1541999094a1b03aebc457823beaa0c8a14488)](https://github.com/dwavesystems/dwave-optimization/pull/153/commits/cb1541999094a1b03aebc457823beaa0c8a14488)

dwave/optimization/src/nodes/manipulation.cpp
samuelbodin: fixed!

dwave/optimization/src/nodes/manipulation.cpp
samuelbodin: Hmm actually not fixed... numpy seems to cover many different cases

```
>>> a = np.arange(10).reshape((2,5))
>>> np.concatenate(a)
array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])

>>> np.concatenate((a))
array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])

>>> np.concatenate((a,))
array([[0, 1, 2, 3, 4],
           [5, 6, 7, 8, 9]])
```

Also works with lists
```
>>> b = np.arange(10).reshape((2,5))
>>> np.concatenate([b])
array([[0, 1, 2, 3, 4],
           [5, 6, 7, 8, 9]])
```

We'll have to resort to some type checking.

```
a = np.array([0,1])
np.concatenate((a))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: zero-dimensional arrays cannot be concatenated
```

dwave/optimization/src/nodes/manipulation.cpp
arcondello: Yes. `array_ptr->size()` will be `-1` if the size of the array is dependent on the state. `array_ptr->size(state)` will always return a non-negative number.

dwave/optimization/src/nodes/manipulation.cpp
arcondello: We don't need to handle every case NumPy does, we just want to be a subset

samuelbodin: Ah okay, then Ill add the one for single array input

samuelbodin: The following is now supported and acts like a copy
```
a = model.constant([0,1])
concatenate((a,)) 
```

Since we require ArraySymbols as predecessors, we should either disallow this or construct a `constant` symbol from it. Wdyt? 
```
concatenate([0,1]) 
```

dwave/optimization/src/nodes/manipulation.cpp
samuelbodin: Fixed! Using `(*ptr)->dynamic()`  [e7213d6](https://github.com/dwavesystems/dwave-optimization/pull/153/commits/e7213d6ab70b55cab35df1273a65babd43e5f069)

dwave/optimization/src/nodes/manipulation.cpp
samuelbodin: Fixed with [7693a54](https://github.com/dwavesystems/dwave-optimization/pull/153/commits/7693a54039726aabb6fa0cb1ec0c3a656de8f53b)

dwave/optimization/include/dwave-optimization/nodes/manipulation.hpp
```diff
@@ -25,6 +25,28 @@
 
 namespace dwave::optimization {
 
+class ConcatenateNode : public ArrayOutputMixin<ArrayNode> {
+ public:
+
+    explicit ConcatenateNode(std::span<ArrayNode*> array_ptrs, ssize_t axis);
```
arcondello, line 31:
```suggestion
 public:
    explicit ConcatenateNode(std::span<ArrayNode*> array_ptrs, ssize_t axis);
```

dwave/optimization/mathematical.py
```diff
@@ -14,11 +14,13 @@
 
 import functools
 import typing
+import collections
```
arcondello, line 17:
```suggestion
import collections
import functools
import typing
```

dwave/optimization/mathematical.py
```diff
@@ -107,6 +110,50 @@ def add(x1: ArraySymbol, x2: ArraySymbol, *xi: ArraySymbol) -> typing.Union[Add,
...
+        ...     model.states.resize(1)
+        ...     print(a_b.state(0))
+        [[0. 1.]
+         [2. 3.]
+         [4. 5.]]
+    """
+    if isinstance(array_likes, ArraySymbol):
+        return array_likes
+
+    if isinstance(array_likes, collections.abc.Iterable) and (0 < len(array_likes)):
```
arcondello, line 147: `Iterable` [might not have a length](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes). You could make is stricter. Probably `Sequence`, that would cover `tuple`/`list`/`np.ndarray`.

dwave/optimization/mathematical.py
```diff
@@ -107,6 +110,50 @@ def add(x1: ArraySymbol, x2: ArraySymbol, *xi: ArraySymbol) -> typing.Union[Add,
...
+        if isinstance(array_likes[0], ArraySymbol):
+            if len(array_likes) == 1:
+                return array_likes[0]
+
+            return Concatenate(tuple(array_likes), axis)
+
+    raise RuntimeError("concatenate takes one or more ArraySymbol as input")
```
arcondello, line 154:
```suggestion
    raise TypeError("concatenate takes one or more ArraySymbol as input")
```
I probably would have gone with `ValueError` myself, but NumPy does a `TypeError` :shrug:

dwave/optimization/symbols.pyx
```diff
@@ -779,6 +781,70 @@ cdef class BinaryVariable(ArraySymbol):
...
+
+    Examples:
+        This example creates a Concatenate symbol.
+
+        >>> from dwave.optimization.model import Model
+        >>> from dwave.optimization.symbols import Concatenate
+        >>> model = Model()
+        >>> a = model.constant([[1,2], [3,4]])
+        >>> b = model.constant([[5,6]])
+        >>> a_b = Concatenate((a,b), axis=0)
```
arcondello, line 795:
```suggestion
        >>> a = model.constant([[1, 2], [3, 4]])
        >>> b = model.constant([[5, 6]])
        >>> a_b = Concatenate((a, b), axis=0)
```
some small pep8 stuff

dwave/optimization/mathematical.py
arcondello, line 147: `Iterable` [might not have a length](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes). You could make is stricter. Probably `Sequence`, that would cover `tuple`/`list`/`np.ndarray`.

```suggestion
    if isinstance(array_likes, collections.abc.Sequence) and (0 < len(array_likes)):
```

arcondello (approved):
LGTM!

wbernoudy (approved):
Fantastic first PR :smile:

PR #154, opened 2024-10-29
Test C++ code with gcc11-14

arcondello:
We want to test the C++ code with several gcc versions. In each case we let the docker image determine the Python version used for meson etc.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -198,19 +198,28 @@ jobs:
       - run: *run-tests
 
 
-  cpp-linux:
-    docker:
-      - image: cimg/python:3.9
+  cpp-gcc:
+    parameters:
+      gcc-version:
+        type: string
 
-    steps: &unix-cpp-tests
+    docker:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -198,19 +198,28 @@ jobs:
...
+      - image: gcc:<< parameters.gcc-version >>
 
-    steps: &unix-cpp-tests
+    steps:
       - checkout
       - run:
+          name: install python requirements
+          command: |
+            apt update
+            apt install python3-pip python3-venv -y
```
wbernoudy, line 215: Perhaps worth specifying python3.11 here?

arcondello: It depends on the docker image we're using. This way it'll inherit what's already there (or so I understand).

wbernoudy: Right, this is only the C++ tests. All good then.

wbernoudy (approved):

PR #155, opened 2024-10-29
Refactor C++ `ArrayIterator` for performance and safety

arcondello:
See also [C.12: Don’t make data members const or references in a copyable or movable type](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c12-dont-make-data-members-const-or-references-in-a-copyable-or-movable-type)


arcondello:
I considered dropping masked iteration altogether - we don't currently use it anywhere. But I do think it'll be useful at some point so I am hesitant to remove it right now.

mdcoury (approved):

wbernoudy (approved):

PR #159, opened 2024-11-05
Feature/modulus

### LINKED ISSUE #156: Consider adding a `modulus` symbol
It would be useful to have a `modulus` symbol for certain scenarios (i.e. identifying weekdays over a multi-week scheduling shift). Recommend mirroring the Python implementation of the modulo operator, i.e.:
```python
15 % 7 # -> 1
-15 % 7 # -> 6
15 % -7 # -> -6
-15 % -7 # -> -1
```

I can start working on a PR for this.

jlalbers:
Added a `Modulus` symbol that can be created with the `%` operator or the `mod()` function in `dwave.optimization.mathematical`.

Closes #156


dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
```diff
@@ -54,6 +55,25 @@ struct min {
...
+        const double y_double = static_cast<double>(y);
+        const double remainder = std::fmod(x_double, y_double);
+        // std::fmod result consistent with numpy for same-sign arguments
+        if ((x_double >= 0) == (y_double >= 0))
+            return remainder;
+        // Make result consistent with numpy for different-sign arguments
+        else
+            return remainder + y_double;
+    }
+};
```
arcondello, line 75:
```python
template <class T>
struct modulus {};

template <>
struct modulus<double> {
    constexpr double operator()(const double& x, const double& y) const {
        // Copy numpy behavior and return 0 for `x % 0`
        if (y == 0) {
            return 0;
        }
        const double remainder = std::fmod(x, y);

        if ((x >= 0) == (y >= 0)) {
            // std::fmod result consistent with numpy for same-sign arguments
            return remainder;
        } else {
             // Make result consistent with numpy for different-sign arguments
            return remainder + y;
        }
    }
};
```
so because we're hardcoding double behaviour, let's do a slightly silly thing and only implement it for `modulus<double>` we can specialize for other types later if we need to. It also slightly simplifies the code.

I also fixed a few brackets for consistency.

jlalbers: Sounds good. It also looks like I need to implement a few of the `BinaryOpNode` functions to fix the gcc tests.

arcondello: Yes, specifically `::min()` , `::max()`, and `::integral()`. Luckily they aren't too complex in this case. Watch the "negative" modulus possibility though.

jlalbers:
@arcondello should be good now, let me know if you spot any issues

tests/test_symbols.py
```diff
@@ -1504,6 +1512,75 @@ def test_state(self):
         np.testing.assert_array_equal(m.state(0), x.state(0))
 
 
+class TestModulus(utils.BinaryOpTests):
```
arcondello, line 1515: Maybe worth testing `0` as well as scalar vs array modulus?

arcondello (approved):
All of my comments are style ones, I think this could be merged as-is. If you do or don't want to make changes, let me know.

tests/test_symbols.py
```diff
@@ -1581,6 +1581,26 @@ def test_float_mod(self):
...
+        ]
+        for lhs, rhs in itertools.product(values, repeat=2):
+            np_result = np.mod(lhs, rhs)
+            lhs_c = model.constant(lhs)
+            rhs_c = model.constant(rhs)
+            result = lhs_c % rhs_c
+
+            with model.lock():
+                model.states.resize(1)
+                print(result.state(0))
```
arcondello, line 1601: left a print statement in.

jlalbers: 😅

Issue #41, opened 2024-07-03
Consider adding node removal


arcondello:
Right now it's not possible to remove nodes from a model at any level. It would be nice to have ways to remove extraneous nodes generated as part of construction, or duplicate nodes, etc.

PR #161, opened 2024-11-05
Add methods to remove unused symbols from a model

arcondello:
Closes https://github.com/dwavesystems/dwave-optimization/issues/41

### DIFF
--- a/dwave/optimization/include/dwave-optimization/graph.hpp
+++ b/dwave/optimization/include/dwave-optimization/graph.hpp
@@ -248,7 +248,7 @@ class Node {
     friend void Graph::reset_topological_sort();
     template <class NodeType, class... Args>
     friend NodeType* Graph::emplace_node(Args&&... args);
-    friend ssize_t Graph::remove_unused_nodes(bool);
+    friend ssize_t Graph::remove_unused_nodes(bool ignore_listeners);
 
  protected:
     // For use by non-dynamic node constructors.
@@ -295,6 +295,9 @@ class Node {
         return static_cast<const StateData*>(state[index].get());
     }
 
+    // Whether this node type is elegible to be removed from the model
+    virtual bool removable() const { return true; }
+
  private:
     ssize_t topological_index_ = -1;  // negative is unset
 
@@ -334,6 +337,10 @@ NodeType* Graph::emplace_node(Args&&... args) {
 }
 
 class ArrayNode: public Array, public virtual Node {};
-class DecisionNode: public Decision, public virtual Node {};
+class DecisionNode: public Decision, public virtual Node {
+ protected:
+    // In general we do not allow decisions to be removed from models.
+    bool removable() const override { return false; }
+};
 
 }  // namespace dwave::optimization
--- a/dwave/optimization/include/dwave-optimization/graph.hpp
+++ b/dwave/optimization/include/dwave-optimization/graph.hpp
@@ -158,6 +158,20 @@ class Graph {
     std::span<DecisionNode* const> decisions() noexcept { return decisions_; }
     std::span<const DecisionNode* const> decisions() const noexcept { return decisions_; }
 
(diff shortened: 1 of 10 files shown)


dwave/optimization/include/dwave-optimization/graph.hpp
```diff
@@ -158,6 +158,20 @@ class Graph {
     std::span<DecisionNode* const> decisions() noexcept { return decisions_; }
     std::span<const DecisionNode* const> decisions() const noexcept { return decisions_; }
 
+    // Remove unused nodes from the graph.
+    //
+    // This method will reset the topological sort if there is one.
+    //
+    // A node is considered unused if all of the following are true:
+    // * It is not a decision.
```
wbernoudy, line 166: This is currently a constraint because of the difficulty of removing some types of decisions, correct?

dwave/optimization/include/dwave-optimization/graph.hpp
```diff
@@ -234,6 +248,7 @@ class Node {
     friend void Graph::reset_topological_sort();
     template <class NodeType, class... Args>
     friend NodeType* Graph::emplace_node(Args&&... args);
+    friend ssize_t Graph::remove_unused_nodes(bool);
```
wbernoudy, line 251: IMO it's nice to add the names of the arguments in declaration even though it's not necessary

dwave/optimization/src/graph.cpp
```diff
@@ -298,6 +298,71 @@ void Graph::propose(State& state, std::vector<const Node*> sources,
...
+    const ssize_t num_nodes_before = this->num_nodes();
+
+    // We generally want to avoid touching decisions
+    const ssize_t num_decisions = this->num_decisions();
+
+    // Mark the nodes that we plan to keep regardless of their number of successors
+    // For performance we store a signalling value in the topological_index of the
+    // nodes. We need to take some care to not override the topological index of
+    // any decisions.
+    constexpr ssize_t keep = -2;
```
wbernoudy, line 316: This is set to `-2` because `-1` already has a meaning of "unsorted", correct? But at this point you've topologically sorted so it should be guaranteed all nodes have an index >= 0.

arcondello: That's right. I just need a negative number. I chose `-2` because `-1` already has meaning as unsorted. Should never matter but :shrug:

dwave/optimization/include/dwave-optimization/graph.hpp
arcondello: In the context of a `friend` declaration it seems even less relevant, but can do.

wbernoudy: Fair enough!

dwave/optimization/include/dwave-optimization/graph.hpp
arcondello: It's not actually difficult to implement but there are a few reasons:
1) I can see how symbols might be accidentally added to the model, but for decisions I think there is a high chance they are there deliberately anyway. I.e. i doubt we'd be removing decisions often.
2) This way if one was to use this method as part of a presolve step, the serialized states of the model after presolve would still be relevant to the original model.

wbernoudy: Good point, I didn't think about the presolve angle.

wbernoudy:
What about the case of an unused `DisjointListNode` (a successor of the `DisjointListsNode`)? Won't this remove it, and cause issues for `DisjointListNode` later on?

arcondello:
Ah! Good catch. Yes. Need to think about how to handle that.

arcondello:
Ok, added a mechanism for nodes to opt out of removal. Probably good for decisions as well (for safety).

wbernoudy (approved):

PR #162, opened 2024-11-06
Refactor ArrayIterator for random access and clearer const/non-const

arcondello:
Almost a complete refactor of `ArrayIterator`. Highlights:
* `const` and non-`const` iterators.
* Moved the implementation to be a nested class inside `Array`.
* Made it a random access iterator.

### DIFF
--- a/dwave/optimization/include/dwave-optimization/array.hpp
+++ b/dwave/optimization/include/dwave-optimization/array.hpp
@@ -237,335 +237,6 @@ struct Update {
     double value;   // The new/current value.
 };
 
-namespace {
-
-// This is a generic iterator for arrays.
-// It handles contiguous, strided, and masked arrays with one type.
-template<bool IsConst>
-class ArrayIteratorImpl_ {
- public:
-    using difference_type = std::ptrdiff_t;
-    using value_type = std::conditional<IsConst, const double, double>::type;
-    using pointer = value_type*;
-    using reference = value_type&;
-
-    ArrayIteratorImpl_() = default;
-
-    ArrayIteratorImpl_(const ArrayIteratorImpl_& other) noexcept
-            : ptr_(other.ptr_),
-              mask_((other.mask_) ? std::make_unique<MaskInfo>(*other.mask_) : nullptr),
-              shape_((other.shape_) ? std::make_unique<ShapeInfo>(*other.shape_) : nullptr) {}
-
-    ArrayIteratorImpl_(ArrayIteratorImpl_&& other) = default;
-
-    ~ArrayIteratorImpl_() = default;
-
-    // Create a contiguous iterator pointing to ptr
-    explicit ArrayIteratorImpl_(value_type* ptr) noexcept
-            : ptr_(ptr), mask_(nullptr), shape_(nullptr) {}
-
-    // Create a masked iterator with a fill value. Will return the value pointed at by *fill_ptr
-    // when *mask_ptr evaluates to true.
-    ArrayIteratorImpl_(value_type* data_ptr, const value_type* mask_ptr,
-                       value_type* fill_ptr) noexcept
-            : ptr_(data_ptr),
-              mask_(std::make_unique<MaskInfo>(mask_ptr, fill_ptr)),
-              shape_(nullptr) {}
-
-    // shape and strides must outlive the iterator!
-    ArrayIteratorImpl_(value_type* ptr, ssize_t ndim, const ssize_t* shape, const ssize_t* strides)
-            : ptr_(ptr),
-              mask_(nullptr),
-              shape_((ndim >= 1) ? std::make_unique<ShapeInfo>(ndim, shape, strides) : nullptr) {}
-    ArrayIteratorImpl_(value_type* ptr, std::span<const ssize_t> shape,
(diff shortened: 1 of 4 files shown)


arcondello:
It still needs a release note and maybe one more editing pass for me. But I think it's in a reviewable state.

dwave/optimization/include/dwave-optimization/array.hpp
```diff
@@ -237,338 +237,473 @@ struct Update {
...
+        friend ArrayIteratorImpl_ operator+(difference_type lhs, ArrayIteratorImpl_ rhs) {
+            return rhs += lhs;
+        }
+        friend ArrayIteratorImpl_ operator-(ArrayIteratorImpl_ lhs, difference_type rhs) {
+            return lhs -= rhs;
+        }
 
-                offset -= strides[dim] * shape[dim];  // zero out this dim
-                loc[dim - 1] = 0;
+        friend difference_type operator-(const ArrayIteratorImpl_& a, const ArrayIteratorImpl_& b) {
```
arcondello, line 415: I contemplated moving this (and a few other similar methods) into the `.cpp` file. But to my great surprise the compiler claims that it did indeed get inlined. So because it's performance-critical I decided to keep it in the header. Probably doesn't actually matter but :shrug: .

dwave/optimization/include/dwave-optimization/array.hpp
```diff
@@ -237,338 +237,473 @@ struct Update {
...
+                difference_type offset = 0;  // the number of bytes we need to move
+
+                // working from right-to-left, figure out how many steps in each
+                // axis. We handle axis 0 as a special case
+                {
+                    // We'll be using std::div() over ssize_t, so we'll store our
+                    // current location in the struct returned by it.
+                    // Unfortunately std::div() is not templated, but overloaded,
+                    // so we use decltype instead.
+                    decltype(std::div(ssize_t(), ssize_t())) qr{ .quot = n, .rem = 0 };
```
wbernoudy, line 632: Given how annoying it is to use `std::div` for this, would it be cleaner to just use `/` and `%` and a temporary variable? Is it that you need consistency for a negative remainder? Seems like you're already accounting for that.

tests/cpp/test_array.cpp
```diff
@@ -15,25 +15,144 @@
...
     CHECK(SizeInfo(5) == SizeInfo(5));
     CHECK(SizeInfo(5) == 5);
 }
 
-TEST_CASE("ArrayIterator") {
+TEST_CASE("ArrayIterator and ConstArrayIterator") {
+    using ArrayIterator = typename Array::iterator;
+    using ConstArrayIterator = typename Array::const_iterator;
+
+    static_assert(std::is_nothrow_constructible<ArrayIterator>::value);
```
wbernoudy, line 40: Should we be putting these kinds of `static_asserts` in the test files as opposed to the cpp files from now on?

wbernoudy (approved):

dwave/optimization/include/dwave-optimization/array.hpp
```diff
@@ -237,338 +237,473 @@ struct Update {
...
+    //
+    // `IsConst` toggles whether the iterator is an output iterator or not.
+    template <bool IsConst>
+    class ArrayIteratorImpl_ {
+     public:
+        using difference_type = std::ptrdiff_t;
+        using value_type = std::conditional<IsConst, const double, double>::type;
+        using pointer = value_type*;
+        using reference = value_type&;
+
```
samuelbodin, line 279: Both `pointer` and `reference` are unused in this class as far as I can tell, but Im guessing they are intended to be used from outside the class for convenience?

samuelbodin (approved):

tests/cpp/test_array.cpp
arcondello: I don't think it much matters. The reason I moved them was because while I was in the middle of development and not all of the asserts were satisfied, I was getting the same compiler error repeated over and over each time the header was included by another file. By putting them in a `.cpp` file, a failed `static_assert` is only transmitted to me by the compiler once. So IMO putting them in the test files makes sense, but it's mostly for convenience reasons.

dwave/optimization/include/dwave-optimization/array.hpp
arcondello: Yeah, this is a holdover from early iterator interface requirements. At one point C++ wanted iterators to have all of these types declared in their namespace. Though they also wanted [iterator_traits](https://en.cppreference.com/w/cpp/iterator/iterator_traits) which I haven't used.

In practice, most of the iterator implementations I've seen around still put them in, so I am inclined to leave them as harmless, if a bit redundant.

tests/cpp/test_array.cpp
wbernoudy: Makes sense to me. Was just curious if we should follow this convention moving forward.

arcondello: Eh, I am ambivalent. I probably will start keeping `static_asserts()` in the `.cpp` files unless I have a reason not to. But not feeling so solid on that yet that I want to claim it's best practice.

PR #163, opened 2024-11-08
Fix Python warnings and add CI check for future ones

arcondello:


dwave/optimization/generators.py
```diff
@@ -460,7 +460,6 @@ class as the decision variable being optimized, with permutations of its
         sublist representing various itineraries for each vehicle.
     """
 
-    time_distances = np.asarray(time_distances)
```
arcondello, line 463: This line is redundant to an eventual `_requires()` call, and in an early implementation of this PR was causing issues. I kept this fix in even though I didn't eventually need it.

thisac (approved):

PR #165, opened 2024-11-13
Add SquareRoot symbol

mdcoury:
Initial add for square_root node, no unit-testing yet... want some feedback on the approach because I don't recall it being so clunky to add a new node when I did this before.


dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
```diff
@@ -58,29 +58,37 @@ struct min {
 template <class T>
 struct modulus {};
 
-template <>
```
alexzucca90, line 61: One thing that @arcondello is gonna mention is to not format unmodified code.

dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -1776,6 +1784,10 @@ double UnaryOpNode<functional::square<double>>::min() const {
     }
     return min * min;
 }
+template <>
+double UnaryOpNode<functional::square_root<double>>::min() const {
+    return std::sqrt(array_ptr_->min());
```
alexzucca90, line 1789: As I mentioned elsewhere, the only thing to decide is how to handle negative numbers... Are we even going to accept the predecessor to have
```cpp
array_ptr_->min() < 0
```
?

mdcoury: Up to you guys.  Like I mentioned in the other thread, I'm planning on letting std::sqrt determine the behviour.  If you want something different, just let me know.

arcondello: We want to disallow it. We're most interested in following NumPy while not allowing `nan`. So we need to require `array_ptr->min() >= 0`.

dwave/optimization/model.pyx
```diff
@@ -2033,6 +2034,14 @@ cdef class ArraySymbol(Symbol):
 
         return self.array_ptr.size()
 
+    def sqrt(self):
```
arcondello, line 2037: NumPy array's don't support `.sqrt()` so neither should we. Instead this should be a function in `mathematical.py` to match [np.sqrt](https://numpy.org/doc/stable/reference/generated/numpy.sqrt.html).

dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -1692,6 +1692,10 @@ template <>
 bool UnaryOpNode<functional::square<double>>::integral() const {
     return array_ptr_->integral();
 }
+template <>
+bool UnaryOpNode<functional::square_root<double>>::integral() const {
+    return array_ptr_->integral();
```
arcondello, line 1697: This is saying that if `a` is an integral than so shall `sqrt(a)`, which is not true. We should return `false`.

dwave/optimization/src/nodes/mathematical.cpp
mdcoury: What's your behaviour expectation?

arcondello: We should check `predecessor->min() >= 0` on construction and raise an ``std::invalid_argument` error if no. That means the the predecessor will always output a non-negative number and the default `std::sqrt` behavior will be correct.

arcondello:
```c++
template <class UnaryOp>
UnaryOpNode<UnaryOp>::UnaryOpNode(ArrayNode* node_ptr)
        : ArrayOutputMixin(node_ptr->shape()), array_ptr_(node_ptr) {
    if constexpr (std::is_same<UnaryOp, functional::square_root<double>>::value) {
        // raise error message here.
    }
    add_predecessor(node_ptr);
}
```

mdcoury:
`@[arcondello](https://github.com/arcondello) arcondello changed the title Add new nodes Add SquareRoot symbol [18 hours ago](https://github.com/dwavesystems/dwave-optimization/pull/165#event-15290629930)`

There are two nodes being added.  I was going to put them both in this PR: do you want them put into separate PRs?

arcondello:
> do you want them put into separate PRs?

They don't need to be, you can change the name again. But "Add new nodes" is not very meaningful, and I often look at the PR titles when writing release notes etc, so if you add another node you can just call it out in the name.

dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
arcondello: To be clear, I am going to be a stickler and ask that these changes be undone. Please don't format code unrelated to the PR.

mdcoury: I changed VScode to strip trailing whitespace because that's what NLSolver requires.  It's going be probably mean that I disable that and either manually add back in the whitespace or throw away this PR and create a new one.  Ultimately, it's your call but honestly it seems ridiculous to worry about trailing whitespace and not something that I've run across is 25 years of software development.

arcondello: It is ultimately a tradeoff between making PRs easy to digest vs tooling in IDEs. Either direction has pros and cons.

tests/test_symbols.py
```diff
@@ -2014,6 +2030,45 @@ class TestSquare(utils.UnaryOpTests):
     def op(self, x):
         return x ** 2
 
+    def num_symbols(self):
+        return 2
+
+
+class TestSquareRoot(utils.SymbolTests):
```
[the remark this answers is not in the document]
mdcoury: @arcondello Doing so seems to lead to a cascade of failures... perhaps I'm missing something, though?

tests/test_symbols.py
```diff
@@ -248,6 +249,9 @@ class UnaryOpTests(SymbolTests):
         def op(self, x):
             pass
 
+        def num_symbols(self):
```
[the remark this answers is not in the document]
mdcoury: I found this necessary in order to as TestSquareRoot implementing UnaryOps... It's unnecessary if I'm implementing it as a SymbolTest...

PR #166, opened 2024-11-14
Add AccumulateZip node/symbol

wbernoudy:
Adds `InputNode` which subclasses `ConstantNode` but allows assigning of the state after initialization.

This is to support the new `NaryReduceNode`, which does an n-ary reduce operation on an arbitrary number of inputs/predecessors. The node is constructed by providing an expression, formulated as a separate `Graph`, and initial values for the reduction operation.

The current design accepts any `Graph` but then throws exceptions if it contains unsupported nodes. It would be best if we can pass along helpful error messages to users at the python level. Currently this works by encoding the error message and a `Node*` in a simple JSON string that is then parsed from Python, where we can grab the corresponding symbol.


wbernoudy:
I tried out some more complicated things to support custom exception handling in Cython. I put it in a separate branch for now as I'm not sure we'll go that direction. Separate PR here to try to make it a bit easier to review https://github.com/wbernoudy/dwave-optimization/pull/1

dwave/optimization/include/dwave-optimization/nodes/constants.hpp
```diff
@@ -118,16 +129,54 @@ class ConstantNode : public ArrayOutputMixin<ArrayNode> {
...
+// InputNode acts like a placeholder or store of data very similar to ConstantNode,
+// with the key different being that its contents *may* change in between propagations.
+// However, it is not a decision variable--instead its use cases are acting as an "input"
+// for "models as functions", or for placeholders in large models where (otherwise constant)
+// data changes infrequently (e.g. a scheduling problem with a preference matrix).
+//
+// Initial values can be provided at node construction that will be used on state initialization.
+// Otherwise, the node will be initialized with all zeros.
+// TODO: finalize the constructors
+class InputNode : public ConstantNode {
```
arcondello, line 144: It is not clear to me what benefit this gets from subclassing `ConstantNode`? It holds a contiguous buffer it seems so it could be a lot simpler - constant does a lot of other things. I was imagining something like
```c++
class InputNode : ArrayOutputMixin<ArrayNode> {

    void initialize_state(State& state) {
        throw std::logic_error("input nodes must have their state initialized explicitly");
    }
    void initialize_state(State& state, std::span<double> values);
}
```
This means that one must do something like
```c++
auto state = graph.empty(state);
input_ptr->initialize_state(state, ...);
graph.initialize_state(state);  // finish everything else
```
obviously for the "outer" graph we would want to check before upload that there are no `Input` symbols, or maybe we just initialize them to `0` by default.

dwave/optimization/include/dwave-optimization/nodes/lambda.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include "dwave-optimization/array.hpp"
+#include "dwave-optimization/graph.hpp"
+#include "dwave-optimization/nodes/constants.hpp"
+
+namespace dwave::optimization {
+
+class NaryReduceNode : public ArrayOutputMixin<ArrayNode> {
+ public:
+    // Runtime constructor that can be used from Cython/Python
+    NaryReduceNode(Graph&& expression, const std::vector<InputNode*>& inputs,
```
arcondello, line 28: I wonder if it makes more sense to not have the user provide inputs etc explicitly, but rather to inspect the `expression`. Maybe `set_objetive()` sets the output? And we can traverse the nodes looking for `InputNode`

dwave/optimization/include/dwave-optimization/utils.hpp
```diff
@@ -166,4 +167,23 @@ void deduplicate_diff(std::vector<Update>& diff);
 // Return whether the given double encodes an integer.
 bool is_integer(const double& value);
 
+// Forward declaration
+class ArrayNode;
```
arcondello, line 171: I don't see where this is used.

arcondello (commented):
Mostly looked at design rather than line-by-line review.

dwave/optimization/include/dwave-optimization/nodes/constants.hpp
wbernoudy: My thinking was that

1. We need the `InputNode` to have a default value so that you _don't_ have to initialize it explicitly.
2. Given 1, might as well take advantage of the constructors of `ConstantNode`

If we can relax 1., then I agree, this would be much simpler! I think my original reasoning for that requirement was flawed anyway.

dwave/optimization/src/array.cpp
```diff
@@ -196,6 +198,41 @@ bool array_shape_equal(const Array& lhs, const Array& rhs) {
     return array_shape_equal(&lhs, &rhs);
 }
 
+bool array_shape_equal(const std::span<const Array* const> array_ptrs) {
```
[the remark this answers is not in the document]
wbernoudy: It does avoid duplicating the number of DFS, but I'm happy to call that premature optimization and switch to just calling the pairwise one.

dwave/optimization/include/dwave-optimization/nodes/lambda.hpp
wbernoudy: I think inputs should be provided explicitly because they need to be mapped to the operands/predecessors. Otherwise we are relying on the implicit ordering of when the nodes are added to the expression model, and it becomes less clear which one is the "special" input that takes the previous value.

But I still find this interface confusing... was thinking about taking a map between inputs/operands as well.

dwave/optimization/include/dwave-optimization/utils.hpp
wbernoudy: It's not, I had used it in a previous version of `is_variant`. Will remove.

dwave/optimization/src/array.cpp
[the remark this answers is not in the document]
arcondello: I'm inclined to switch to the simpler implementation for now. But could be convinced either way.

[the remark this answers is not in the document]
wbernoudy: I'll switch it.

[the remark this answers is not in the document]
arcondello: Thinking slightly more, perhaps the best of both worlds would be to have the pairwise one call the n-length one.

dwave/optimization/include/dwave-optimization/nodes/lambda.hpp
arcondello: At the Python level this could perhaps be addressed by not adding an `input()` method, but having a subclass of `Model` that requires `num_inputs` at construction-time. Then we rely on the order? We might have to treat inputs as decision variables in that they get topologically ordered immediately in that case... Needs thought.

wbernoudy: I like that. The `ExpressionModel` or whatever (subclass of `Model`) could disallow adding decisions.

Perhaps we could also somehow pull a list of supported nodes from C++ and then disallow adding unsupported symbols to `ExpressionModel`. This will go a long way to prevent constructing unsupported expressions, as I don't think we will have any supported nodes that can have non-scalar output given all scalar inputs.

wbernoudy: And I'll note I expect the standard flow when creating expressions is just to create a list of `Input`s anyway and use that. So `exp = ExpressionModel(7)`, `exp.inputs[0] + exp.inputs[1]` feels natural to me.

dwave/optimization/model.py
```diff
@@ -40,7 +40,61 @@
...
+    def input(self, lower_bound: float, upper_bound: float, integral: bool):
+        """TODO"""
+        # avoid circular import
+        from dwave.optimization.symbols import Input
+        # Shape is always scalar for now
+        return Input(self, lower_bound, upper_bound, integral, shape=tuple())
+
+    def set_output(self, value: ArraySymbol):
+        """TODO"""
+        self.output = value
```
arcondello, line 74: Probably we want to set this as the objective for the `_Graph`? So `super().minimize(value)`. Even better:
1) Rename `_Graph.minimize()` -> `_Graph._set_objective()`
2) Add `Model.minimize()` wrapping it
3) Call it from here.

dwave/optimization/src/nodes/constants.cpp
```diff
@@ -18,6 +18,7 @@
 #include <ranges>
 
 #include "dwave-optimization/utils.hpp"
+#include "_state.hpp"
```
arcondello, line 21: We don't use this?

dwave/optimization/src/nodes/lambda.cpp
```diff
@@ -0,0 +1,300 @@
+// Copyright 2023 D-Wave Systems Inc.
```
arcondello, line 1: 2024 D-Wave Inc.

dwave/optimization/src/nodes/lambda.cpp
```diff
@@ -0,0 +1,300 @@
...
+
+    double min_val = std::numeric_limits<double>::infinity();
+    double max_val = -std::numeric_limits<double>::infinity();
+
+    static double dummy = 0;
+    bool all_is_integral = true;
+    for (const double& v : new_values) {
+        min_val = std::min(min_val, v);
+        max_val = std::max(min_val, v);
+        all_is_integral &= (std::modf(v, &dummy) == 0.0);
```
arcondello, line 70: https://github.com/dwavesystems/dwave-optimization/blob/0eb1ccc996270d0d39a041eb5228346de64193f0/dwave/optimization/include/dwave-optimization/utils.hpp#L167

dwave/optimization/src/nodes/lambda.cpp
```diff
@@ -0,0 +1,300 @@
...
+    }
+    if (max_val > max()) {
+        throw std::invalid_argument("new data contains a value smaller than the min");
+    }
+    if (integral() && !all_is_integral) {
+        throw std::invalid_argument("new data contains a non-integral value");
+    }
+
+    data_ptr<ArrayNodeStateData>(state)->assign(new_values);
+}
```
arcondello, line 84:
```C++
void InputNode::assign(State& state, std::span<const double> new_values) const {
    if (static_cast<ssize_t>(new_values.size()) != this->size()) {
        throw std::invalid_argument("size of new values must match");
    }

    if (!new_values.empty())
        if (std::ranges::min(new_values) < min()) {
            throw std::invalid_argument("new data contains a value smaller than the min");
        }
        if (std::ranges::max(new_values) > max()) {
            throw std::invalid_argument("new data contains a value smaller than the min");
        }
        if (integral() && std::ranges::all_of(new_values, is_integer)) {
            throw std::invalid_argument("new data contains a non-integral value");
        }
    }

    data_ptr<ArrayNodeStateData>(state)->assign(new_values);
}
```
could also use [ranges::minmax()](https://en.cppreference.com/w/cpp/algorithm/ranges/minmax) I suppose

dwave/optimization/include/dwave-optimization/array.hpp
```diff
@@ -935,6 +935,10 @@ std::ostream& operator<<(std::ostream& os, const Array::View& view);
 bool array_shape_equal(const Array* lhs_ptr, const Array* rhs_ptr);
 bool array_shape_equal(const Array& lhs, const Array& rhs);
 
+// Test whether multiple arrays all have the same shape.
+bool array_shape_equal(const std::span<const Array* const> array_ptrs);
+bool array_shape_equal(const std::vector<const Array*>& array_ptrs);
```
arcondello, line 940: Isn't this overload redundant? Do you mean to say
```c++
bool array_shape_equal(const std::vector<const Array*> array_ptrs);
```
or
```c++
bool array_shape_equal(const std::vector<const Array*>&& array_ptrs);
```
?

I am guessing we can just drop it.

dwave/optimization/include/dwave-optimization/nodes/lambda.hpp
```diff
@@ -0,0 +1,109 @@
...
+class InputNode : public ArrayOutputMixin<ArrayNode> {
+ public:
+    explicit InputNode(std::span<const ssize_t> shape, double min, double max, bool integral)
+            : ArrayOutputMixin(shape), min_(min), max_(max), integral_(integral) {};
+
+    explicit InputNode(std::initializer_list<ssize_t> shape, double min, double max, bool integral)
+            : ArrayOutputMixin(shape), min_(min), max_(max), integral_(integral) {};
+
+    explicit InputNode()
+            : InputNode({}, -std::numeric_limits<double>::infinity(),
```
arcondello, line 41:
```c++
            : InputNode({}, std::numeric_limits<double>::lowest(),
```

dwave/optimization/include/dwave-optimization/nodes/lambda.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+    double const* buff(const State&) const override;
+
+    std::span<const Update> diff(const State& state) const noexcept override;
+
+    void propagate(State& state) const noexcept override {};
+    void commit(State& state) const noexcept override;
+    void revert(State& state) const noexcept override;
+
+    void assign(State& state, const std::vector<double>& new_values) const;
```
arcondello, line 65: Isn't this one redundant?

dwave/optimization/src/array.cpp
```diff
@@ -162,36 +165,47 @@ std::string shape_to_string(const std::span<const ssize_t> shape) {
...
+    while (first_size.array_ptr != nullptr && first_size.array_ptr != first_ptr) {
+        first_ptr = first_size.array_ptr;
+        first_size = first_size.substitute();
     }
-    while (rhs_size.array_ptr != rhs_ptr) {
-        rhs_ptr = rhs_size.array_ptr;
-        rhs_size = rhs_size.substitute();
-        if (lhs_size == rhs_size) return true;
+
+    for (const Array* array_ptr : array_ptrs | std::views::take(1)) {
```
arcondello, line 182: Do you meant `std::views::drop(1)`?

dwave/optimization/src/nodes/lambda.cpp
```diff
@@ -0,0 +1,300 @@
...
+            throw std::invalid_argument(
+                    R"({"message": "Expression should only contain scalars", "node_ptr": )" +
+                    std::to_string((uintptr_t)(void*)node_ptr.get()) + "}");
+        }
+    }
+
+    return expression;
+}
+
+auto get_operands_shape(const std::vector<InputNode*>& inputs,
```
arcondello, line 142: Why not `std::span`?

dwave/optimization/src/array.cpp
wbernoudy: Yep, not sure how I missed that.

dwave/optimization/include/dwave-optimization/array.hpp
wbernoudy: You're right, probably leftover from when I was trying to get the rvalue version to work.

dwave/optimization/model.py
wbernoudy: Ended up doing exactly this.

dwave/optimization/_model.pyx
```diff
@@ -259,8 +261,10 @@ cdef class _Graph:
         """
         if not self.is_locked():
             # lock for the duration of the method
-            with self.lock():
-                return self.into_file(file, max_num_states=max_num_states, only_decision=only_decision)
+            self.lock()
+            self.into_file(file, max_num_states=max_num_states, only_decision=only_decision)
+            self.unlock()
+            return
```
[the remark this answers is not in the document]
wbernoudy: The context manager version of `.lock()` is defined only for `Model` currently. I could also add it to `Expression` or `_Graph`.

dwave/optimization/libcpp/graph.pxd
```diff
@@ -38,22 +39,33 @@ cdef extern from "dwave-optimization/graph.hpp" namespace "dwave::optimization"
...
 # Sometimes Cython isn't able to reason about pointers as template inputs, so
 # we make a few aliases for convenience
 ctypedef Node* NodePtr
 ctypedef ArrayNode* ArrayNodePtr
 ctypedef DecisionNode* DecisionNodePtr
 
+# This seems to be necessary to allow Cython to iterate over the returned
+# span from `inputs()` directly. Otherwise it tries to cast it to a non-const
+# version of span before iterating, which the C++ compiler will complain about.
+ctypedef InputNode* const constInputNodePtr
```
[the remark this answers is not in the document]
wbernoudy: I am still very confused by it as well. But If I change the usage below to `const InputNodePtr` like the other methods that return a span to node pointers, I get
``` Cannot assign type 'InputNodePtr' (alias of 'InputNode *') to 'InputNode *'```

File: dwave/optimization/_model.pyx (hunk 15 above)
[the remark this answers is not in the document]
Comment (arcondello) [reply dwave/optimization/_model.pyx:267]: Ah, excellent point! In which case we should at least do it in a
```
self.lock()
try:
    self.into_file(...)
finally:
    self.unlock()
```

File: dwave/optimization/libcpp/graph.pxd (hunk 16 above)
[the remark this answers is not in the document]
Comment (arcondello) [reply dwave/optimization/libcpp/graph.pxd:54]: That's a head scratcher for sure. Certainly nothing worth blocking this over.

File: dwave/optimization/_model.pyx (hunk 15 above)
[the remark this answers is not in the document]
Comment (wbernoudy) [reply dwave/optimization/_model.pyx:267]: Yep, that's better!

File: dwave/optimization/include/dwave-optimization/graph.hpp [hunk 17]
```diff
@@ -73,6 +74,8 @@ class Graph {
  public:
     Graph();
     ~Graph();
+    Graph(Graph&&);
+    Graph& operator=(Graph&& other) = default;
```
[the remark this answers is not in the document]
arcondello: Nevermind! See https://github.com/dwavesystems/dwave-optimization/pull/184

dwave/optimization/libcpp/graph.pxd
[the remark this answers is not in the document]
arcondello: I now understand what's happening here. Cython is getting confused between the `InputNode` in `libcpp.graph` and the `InputNode` in `libcpp.nodes`. 

I made a PR to this branch changing it, see https://github.com/wbernoudy/dwave-optimization/pull/3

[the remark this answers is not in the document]
wbernoudy: Ah, nice find and thanks for the fix.

dwave/optimization/symbols.pyx
```diff
@@ -2327,6 +2403,176 @@ cdef class NaryMultiply(ArraySymbol):
...
+        >>> expr.set_output(xi + previous)
+        >>> cumulative_sum_x = NaryReduce(expr, (x,))
+        >>> type(cumulative_sum_x)
+        <class 'dwave.optimization.symbols.NaryReduce'>
+    """
+
+    def __init__(
+        self,
+        expression: Expression,
+        operands: Collection[ArraySymbol],
```
arcondello, line 2447: These types don't do anything, so I'd rather remove them, and the `from typing import Collection`.

dwave/optimization/_model.pyi
```diff
@@ -53,6 +53,7 @@ class _Graph:
     def is_locked(self) -> bool: ...
     def iter_constraints(self) -> collections.abc.Iterator[ArraySymbol]: ...
     def iter_decisions(self) -> collections.abc.Iterator[Symbol]: ...
+    def iter_inputs(self) -> collections.abc.Iterator[Symbol]: ...
```
arcondello, line 56: I think it's OK to use `ArraySymbol` here.

dwave/optimization/include/dwave-optimization/nodes/inputs.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+    void assign(State& state, std::span<const double> new_values) const;
+
+    double const* buff(const State&) const override;
+    void commit(State& state) const noexcept override;
+    std::span<const Update> diff(const State& state) const noexcept override;
+
+    bool integral() const override { return integral_; };
+
+    void initialize_state(State& state) const override {
```
arcondello, line 52: Could mark this as [noreturn](https://en.cppreference.com/w/cpp/language/attributes/noreturn). Though not really sure how helpful that ends up being.

dwave/optimization/symbols.pyx
```diff
@@ -17,6 +17,7 @@
 import collections.abc
 import json
 import numbers
+import tempfile
```
arcondello, line 20: I don't think this is used?

tests/cpp/nodes/test_constants.cpp
```diff
@@ -15,6 +15,7 @@
 #include "catch2/catch_test_macros.hpp"
 #include "dwave-optimization/graph.hpp"
 #include "dwave-optimization/nodes/constants.hpp"
+#include "dwave-optimization/nodes/testing.hpp"
```
arcondello, line 18: Not used

wbernoudy: Removed these, found a few more I could remove as well.

arcondello (approved):

dwave/optimization/symbols.pyx
```diff
@@ -2325,6 +2399,173 @@ cdef class NaryMultiply(ArraySymbol):
...
+    computed operation, and computes a new value at the next output index.
+
+    This takes inspiration from
+    `numpy.ufunc.reduce <https://numpy.org/doc/2.1/reference/generated/numpy.ufunc.reduce.html>`_ 
+    but is different in that the reduction operation can take an arbitrary
+    number of arguments, instead of always two (hence the "n-ary" in
+    `NaryReduce`).
+
+    Args:
+        expression:
```
arcondello, line 2416: I just noticed that [functools.reduce()](https://docs.python.org/3/library/functools.html#functools.reduce) and [std::accumulate()](https://en.cppreference.com/w/cpp/algorithm/accumulate) both use the left-most argument for the init/accumulated value. So I think we should follow that convention by making it the first input.

dwave/optimization/src/nodes/lambda.cpp
```diff
@@ -0,0 +1,324 @@
...
+class AccumulateZipNodeData : public ArrayNodeStateData {
+ public:
+    explicit AccumulateZipNodeData(std::vector<double>&& values,
+                                std::vector<Array::const_iterator>&& iterators, State&& state)
+            : ArrayNodeStateData(std::move(values)),
+              iterators(std::move(iterators)),
+              register_(std::move(state)) {}
+
+    // used to avoid reallocating memory for predecessor iterators every propagation
+    std::vector<Array::const_iterator> iterators;
```
arcondello, line 34: It's not in general safe to re-use iterators between propagations. Nodes might reallocate their state memory (e.g. if they're storing them in a `vector`). Because we require fixed-size predecessors, it's probably safe in practice, but not sure we should rely on that behavior.

wbernoudy: They are not re-used between propagations, this is just a bit of premature optimization to avoid reallocating this vector each time

dwave/optimization/expression.py
```diff
@@ -0,0 +1,103 @@
...
+from dwave.optimization.model import Model
+
+
+__all__ = ["expression"]
+
+
+class Expression:
+    """An expression that can be used as an input to other symbols.
+
+    .. versionadded:: 0.6.3
```
arcondello, line 31: 0.6.4

dwave/optimization/src/nodes/lambda.cpp
```diff
@@ -0,0 +1,338 @@
+// Copyright 2024 D-Wave Inc.
```
arcondello, line 1:
```
// Copyright 2024 D-Wave
```
also 2024, it's been so long!

Some other files as well.

arcondello (commented):
Couple minor aesthetic bits

PR #170, opened 2024-11-18
Update README's Contributing section with Release Note subsection

arcondello:
This way I can stop [linking dimod](https://github.com/dwavesystems/dwave-optimization/pull/169#issuecomment-2483516712).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -120,3 +120,23 @@ has guidelines for contributing to Ocean packages.
 
 ``dwave-optimization`` includes some formatting customization in the
 `.clang-format <.clang-format>`_ and `setup.cfg <setup.cfg>`_ files.
+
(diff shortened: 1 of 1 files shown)


JoelPasvolsky (approved):
Needs to also be added to [SDK](https://docs.ocean.dwavesys.com/en/stable/docs_optimization/sdk_index.html)

(Note that I'm looking at switching the RST RNs to links to the RNs in source to prevent warnings on duplicates.)

JoelPasvolsky:
> Not all Ocean packages in the SDK

I meant to add it under dwave-optimization as we do for [dimod](https://docs.ocean.dwavesys.com/en/stable/docs_dimod/release_notes.html)

PR #175, opened 2024-11-26
Add release note for Concatenate symbol

arcondello:
https://github.com/dwavesystems/dwave-optimization/pull/153 was missing a release note. My bad for not catching it in the review.


samuelbodin:
> #153 was missing a release note. My bad for not catching it in the review.

All good, I missed it too :P

samuelbodin (approved):

PR #176, opened 2024-12-02
Move `States` into a new `dwave/optimization/states.pyx`

arcondello:
Also `Model` no longer holds onto a Cython-level reference to the `States`, but rather holds it as a generic Python object. We weren't using that access for anything performant we care about, and this is a weaker coupling which will help with circular dependencies and Python-ization of some of the current Cython code.

The overall buildtime hasn't changed any because it's dominated by the `Model` and the symbols.

This is the first in a sequence of changes meant to separate out C++ `Graph` management from the `Model` itself. See also https://github.com/dwavesystems/dwave-optimization/pull/174.


dwave/optimization/states.pxd
```diff
@@ -0,0 +1,44 @@
...
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+from libcpp.vector cimport vector
+
+from dwave.optimization.libcpp.state cimport State as cppState
+from dwave.optimization.model cimport Model
+
```
wbernoudy, line 19: Two newlines

dwave/optimization/states.pyx
```diff
@@ -0,0 +1,331 @@
+# Copyright 2024 D-Wave
```
wbernoudy, line 1: Is `# distutils: language = c++` still recommended for cython files?

dwave/optimization/states.pxd
```diff
@@ -0,0 +1,44 @@
...
+    # Accessors should check the length of the state when accessing!
+    cdef vector[cppState] _states
+
+    # The number of views into the states that exist. The model cannot
+    # be unlocked while there are unsafe views
+    cdef public Py_ssize_t _view_count
+
+    # Object that contains or will contain the information needed to construct states
+    cdef readonly object _future
+    cdef readonly object _result_hook
```
wbernoudy, line 44: Missing `StateView` declaration?

dwave/optimization/states.pyx
arcondello: `distutils` itself is long since deprecated. And ever since we switched to `meson` build system these headers are ignored. I intent to clear them out of the other files as I go.

dwave/optimization/states.pxd
arcondello: Deliberately omitted. The only reason to put it in the `.pxd` file is to allow it to be `cimport`ed. We don't do that anywhere and I can't think of a a particular benefit a user would get from being able to do so. So by omitting it we have less surface area to worry about.

wbernoudy (approved):
Assuming `States`, `StateView` and tests were just copied over, LGTM

PR #179, opened 2024-12-03
Make `Model` a Python class by moving Cython parts into a new `_Graph` class

arcondello:
See also https://github.com/dwavesystems/dwave-optimization/pull/174

Some highlights:
* Most of the pure-Python methods have been moved to `Model` rather that `_Graph`
* `_Graph` doesn't hold onto `States`, `Model` does
* Several methods (e.g. `.lock`, `.unlock`, `.minimize`) have "basic" features implemented in Cython but some bits are implemented in Python.
* Keep the `cimport`-ability of `Model` via some aliasing tricks (need to contemplate how much I like this one a little more)


arcondello:
A few things left to do:

- [ ] Better document the alias thing if we want to keep it
- [ ] Add `typing` to `model.py`
- [ ] Another pass over the docstrings for clarity

arcondello:
Upon reflection, I think it makes sense to keep `Symbol.model` as `model` rather than say renaming it `Symbol._graph`. Even if we were to add [an Expression class](https://github.com/dwavesystems/dwave-optimization/pull/166) calling expressions "model" is IMO not such a bad thing. If it later becomes confusing we can always do the rename in a different PR.

dwave/optimization/model.py
```diff
@@ -0,0 +1,499 @@
...
+            This example creates a :math:`1 \times 20`-sized binary symbol.
+
+            >>> from dwave.optimization.model import Model
+            >>> model = Model()
+            >>> x = model.binary((1,20))
+        """
+        from dwave.optimization.symbols import BinaryVariable  # avoid circular import
+        return BinaryVariable(self, shape)
+
+    def constant(self, array_like: numpy.typing.ArrayLike) -> Constant:
```
wbernoudy, line 133: This and maybe `to_networkx` are the only methods I believe that we'll want to have on the future `Expression` class that we would duplicate. Considering how short this is and that `to_networkx` is not necessary for now (if ever) seems fine to me.

dwave/optimization/_graph.pyx
```diff
@@ -49,44 +36,24 @@ from dwave.optimization.states cimport States
...
-    Examples:
-        This example creates a model for a
-        :class:`flow-shop-scheduling <dwave.optimization.generators.flow_shop_scheduling>`
-        problem with two jobs on three machines.
-
-        >>> from dwave.optimization.generators import flow_shop_scheduling
-        ...
-        >>> processing_times = [[10, 5, 7], [20, 10, 15]]
-        >>> model = flow_shop_scheduling(processing_times=processing_times)
+    It is not intended for a user to use ``_Graph`` directly. Rather classes
```
wbernoudy, line 45: Rather,

wbernoudy (approved):

PR #181, opened 2024-12-10
Refactor `Update`'s comparison methods slightly

arcondello:


dwave/optimization/src/utils.cpp
```diff
@@ -23,7 +23,7 @@ namespace dwave::optimization {
 void deduplicate_diff(std::vector<Update>& diff) {
     if (diff.empty()) return;
 
-    std::stable_sort(diff.begin(), diff.end());
+    std::ranges::stable_sort(diff);
```
arcondello, line 26: No functional change from this, just using a slightly more modern method.

wbernoudy (approved):

PR #183, opened 2024-12-11
Flatten the docs for Model, State, and mathematical functions

arcondello:
The main benefit is that new methods etc are automatically incorporated into the docs.

Similar to https://github.com/dwavesystems/dwave-optimization/pull/73, except for the other classes.


arcondello:
Ah! Already seeing the benefit. It found a broken docstring that wasn't working before. Will fix.

randomir (approved):
LGTM!

IMO, all our docs should be generated like this. With zero manual overhead. Otherwise we forget to include/update new methods/classes, more often than not.

PR #184, opened 2024-12-11
Fixup C++ Graph constructors and assignement operators

arcondello:
Use better practices. I.e.
* Mark the relevant methods as `noexcept`. See [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c66-make-move-operations-noexcept).
* Explicitly disallow copy construction/assigment. This never worked so better to be explicit about it.
* Implement the [rule of five](https://en.cppreference.com/w/cpp/language/rule_of_three).
* Add a few explicit tests.
* Move the definitions into the header. See also [this stackoverflow thread](https://stackoverflow.com/questions/62580224/should-functions-declared-with-default-only-go-in-the-header-file).

### DIFF
--- a/tests/cpp/test_graph.cpp
+++ b/tests/cpp/test_graph.cpp
@@ -174,6 +174,69 @@ TEST_CASE("Topological Sort", "[topological_sort]") {
     // test edge cases.
 }
 
+TEST_CASE("Graph constructors, assignment operators, and swapping") {
+    static_assert(std::is_nothrow_default_constructible<Graph>::value);
+    static_assert(!std::is_copy_constructible<Graph>::value);
+    static_assert(!std::is_copy_assignable<Graph>::value);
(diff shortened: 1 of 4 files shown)


tests/cpp/test_graph.cpp
```diff
@@ -174,6 +174,69 @@ TEST_CASE("Topological Sort", "[topological_sort]") {
...
+            THEN("The nodes have all be moved over") {
+                // these tests are far from complete, but we're using default
+                // constructors/operators so this is intended to be a sanity
+                // check
+                CHECK(other.nodes().size() == 3);
+                CHECK(graph.nodes().size() == 0);
+            }
+        }
+
+        WHEN("We cosntruct another graph using the move assignment operator") {
```
wbernoudy, line 204: construct*

wbernoudy (approved):

PR #186, opened 2024-12-12
Refactor some tests to use ArrayValidationNode

arcondello:
Replace uses of `verify_array_diff` with `ArrayValidationNode`.


tests/cpp/nodes/indexing/test_advanced.cpp
```diff
@@ -120,8 +121,6 @@ TEST_CASE("AdvancedIndexingNode") {
                     CHECK(B_ptr->size(state) == s_ptr->size(state));
 
                     CHECK(std::ranges::equal(B_ptr->view(state), std::vector{4}));
-
-                    verify_array_diff({}, {4}, B_ptr->diff(state));
```
wbernoudy, line 124: Don't we need to call propagate on the `ArrayValidationNode` to ensure the diff is valid? Maybe we can switch a lot of these tests to using `graph->propagate()` so we don't have to propagate all the nodes individually...

arcondello: Oh! Good catch. Will fix everywhere.

arcondello: Ok, should be fixed. Also made the tests a bit more future proof by just propagating/commiting/reverting all descendants.

tests/cpp/nodes/test_collections.cpp
```diff
@@ -463,6 +450,8 @@ TEST_CASE("SetNode") {
 
         auto ptr = graph.emplace_node<SetNode>(num_elements);
 
+        graph.emplace_node<ArrayValidationNode>(ptr);
```
wbernoudy, line 453: I think you're still missing propagate for this one

arcondello: Good catch. Ok, this time I think I got them all.

arcondello: ah, still an error. One sec.

arcondello: ok, fixed. This time for real.

wbernoudy (approved):

PR #187, opened 2024-12-19
Add some convenience methods to `Array::View`

arcondello:
As well as some other minor refactors and testing.

I wrote these while working on some experiments and thought they might be more generally useful.


alexzucca90 (approved):
LGTM

Issue #188, opened 2024-12-19
Consider dropping ``Node::update()`` method


arcondello:
Currently the `Node` design includes a `Node::update()` method. This method is called by the node's predecessors if they have updates they want to propagate to the node. In principal this allows nodes to more efficiently check their predecessors for updates or to eagerly update their state. However, as of writing (Dec 2024) no nodes actually use this information, and many nodes don't actually call their successor's update.

With this in mind, we should consider dropping the notion entirely. The downside is nodes will need to always check every predecessor for updates on each propagation. And it means that all nodes will update their state lazily rather than eagerly. The upside is one less virtual method call/test. Which in the case of many scalars likely has a noticeable performance difference.

wbernoudy:
I'm for dropping `update()` entirely. I think it makes the most sense for nodes that could eagerly update their state, but we still don't have any great motivating use-cases for that. In terms of efficiently checking predecessors for updates, each successor just doing `predecessor->diff().size() == 0` (we could even add a simpler method) if they want seems to have the same computational cost and is just as clear.

PR #189, opened 2025-01-02
Mark Decision::update() and ConstantNode::update() as noreturn

arcondello:
A slight tidying of some `::update()` methods. IMO it makes sense to do this as an intermediate step while still contemplating https://github.com/dwavesystems/dwave-optimization/issues/188.


wbernoudy (approved):
For my future self: may need to similar changes to `InputNode` in #166

PR #194, opened 2025-01-06
Fix for issue 190

samuelbodin:
`concatenate()` should handle a numpy object array of symbols

### DIFF
--- a/tests/test_symbols.py
+++ b/tests/test_symbols.py
@@ -778,6 +778,39 @@ class TestConcatenate(utils.SymbolTests):
         model.lock()
         yield c
 
+    def test_simple_concatenate(self):
+        model = Model()
+        with self.subTest("Concatenate ndarray of constants returns ndarray"):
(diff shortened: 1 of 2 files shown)


arcondello:
Can you add a unittest or two?

tests/test_symbols.py
```diff
@@ -778,6 +778,39 @@ def generate_symbols(self):
         model.lock()
         yield c
 
+    def test_simple_concatenate(self):
+        model = Model()
+        with self.subTest("Concatenate ndarray of constants returns ndarray"):
```
arcondello, line 783: I don't think we need/want this one. No need to raise an error or anything, but I wouldn't say this is defined behavior.

arcondello (approved):
I'd drop one of the tests, otherwise looks good to me

tests/test_symbols.py
samuelbodin: Removed it

PR #196, opened 2025-01-07
Fix for issue 191

samuelbodin:
concatenate() should handle scalars

### DIFF
--- a/dwave/optimization/mathematical.py
+++ b/dwave/optimization/mathematical.py
@@ -143,6 +143,21 @@ def concatenate(array_likes : typing.Union[collections.abc.Iterable, ArraySymbol
         [[0. 1.]
          [2. 3.]
          [4. 5.]]
+
+        This example concatenates two sequences of scalars by first stacking them.
+
+        >>> from dwave.optimization import Model, stack, concatenate
+        ...
+        >>> model = Model()
+        >>> a = stack([model.constant(1), model.constant(2)])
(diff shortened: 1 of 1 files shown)


dwave/optimization/mathematical.py
```diff
@@ -52,6 +52,7 @@
     "minimum",
     "mod",
     "multiply",
+    "stack",
     "sqrt",
     "where",
```
arcondello, line 57:
```python
    "sqrt",
    "stack",
    "where",
```

dwave/optimization/mathematical.py
```diff
@@ -516,6 +532,40 @@ def multiply(x1: ArraySymbol, x2: ArraySymbol, *xi: ArraySymbol,
     raise RuntimeError("implementated by the op() decorator")
 
 
+def stack(xi: collections.abc.Iterable):
```
arcondello, line 535: I think this can apply for any number of dimensions, not just 0d. If we want to simplify, we could implement `vstack` instead, so it's always concatenating along the 1st axis. But I am guessing the `stack` isn't so bad.

Also `-> ArraySymbol` for the output typing.

arcondello (commented):
Needs tests and a release note.

dwave/optimization/mathematical.py
samuelbodin: Ah good points, will fix

PR #198, opened 2025-01-09
Add release notes to docs

arcondello:


docs/release_notes.rst
```diff
@@ -0,0 +1,5 @@
+.. release-notes:: Release Notes
+    :collapse-pre-releases:
```
JoelPasvolsky, line 2: I don't see that this option is doing anything:
image

docs/release_notes.rst
```diff
@@ -0,0 +1,5 @@
...
+    :ignore-notes:
+        fix-to_networkx-a78f0f669cc9638c.yaml
+    :reporoot: ../
```
JoelPasvolsky, line 5: Is this needed? The `sphinx-build` command is run from the parent directory

JoelPasvolsky (approved):
For the SDK I would like to just give pointers to the GitHub RNs, they're formatted nicely now, and otherwise the reuse of the same tags in all the submodules causes endless warnings.

docs/release_notes.rst
arcondello: It's not a pre-release, it's a release candidate :shrug:

docs/release_notes.rst
arcondello: This is relative to the `docs/` directory, the repo root is one outside of that.

PR #201, opened 2025-01-16
Add ability to insert values into an array symbol

arcondello:
Following [numpy.put()](https://numpy.org/doc/2.1/reference/generated/numpy.put.html). Though we return a new array rather than mutating in-place.

Closes https://github.com/dwavesystems/dwave-optimization/issues/202

### DIFF
--- a/dwave/optimization/mathematical.py
+++ b/dwave/optimization/mathematical.py
@@ -536,7 +536,16 @@ def put(array: ArraySymbol, indices: ArraySymbol, values: ArraySymbol) -> Put:
 
     Args:
         array: Base array. Must be not be a dynamic array nor a scalar.
-        indices: The indices in the flattened base array to be replaced.
+        indices:
+            The indices in the flattened base array to be replaced.
+
+            .. warning::
+                If ``indices`` has duplicate values, it is undefined which
+                value from ``values`` will be propagated.
+                This will likely hurt the performance of the model.
+                Care should be taken to ensure that ``indices`` does not
+                contain duplicates.
+
         values: Values to place in ``array`` at ``indices``.
 
     Examples:
--- a/dwave/optimization/mathematical.py
+++ b/dwave/optimization/mathematical.py
@@ -541,7 +541,8 @@ def put(array: ArraySymbol, indices: ArraySymbol, values: ArraySymbol) -> Put:
 
             .. warning::
                 If ``indices`` has duplicate values, it is undefined which
-                value from ``values`` will be propagated.
+                of the possible corresponding values from ``values`` will
+                be propagated.
                 This will likely hurt the performance of the model.
                 Care should be taken to ensure that ``indices`` does not
                 contain duplicates.
--- a/dwave/optimization/mathematical.py
+++ b/dwave/optimization/mathematical.py
@@ -33,6 +33,7 @@ from dwave.optimization.symbols import (
     NaryMultiply,
     Not,
     Or,
+    Put,
     SquareRoot,
     Where,
     Xor,
@@ -52,6 +53,7 @@ __all__ = [
     "minimum",
     "mod",
     "multiply",
+    "put",
     "sqrt",
     "stack",
     "where",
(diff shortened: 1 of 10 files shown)


dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -170,6 +172,247 @@ void ConcatenateNode::revert(State& state) const {
...
+        mask_diff_.clear();
+    }
+
+    std::unique_ptr<NodeStateData> copy() const override {
+        return std::make_unique<PutNodeState>(*this);
+    }
+
+    // decrement the mask count at index and set the state to array_value if
+    // the output is no longer masked
+    void decrement_mask(ssize_t index, double array_value) {
```
wbernoudy, line 196: I don't think the logic here handles all cases. For example, if we remove an index that was duplicated, this will decrement the mask counter (correct) but will still use the original corresponding value, which may not be the same as corresponding value for the duplicate index.

I tried to write a test case that checks for this, this one is currently failing and (I think) it should pass:
```cpp
    GIVEN("A length 6 array of integers, and two integer arrays for indexing and values") {
        auto graph = Graph();

        auto a_ptr = graph.emplace_node<IntegerNode>(std::initializer_list<ssize_t>{6});
        auto ind_ptr = graph.emplace_node<IntegerNode>(std::initializer_list<ssize_t>{2}, 0, 5);
        auto val_ptr = graph.emplace_node<IntegerNode>(std::initializer_list<ssize_t>{2});
        auto mask_ptr = graph.emplace_node<DynamicArrayTestingNode>(std::initializer_list<ssize_t>{-1}, 0, 1, true);

        auto put_ptr = graph.emplace_node<PutNode>(
                a_ptr,
                graph.emplace_node<AdvancedIndexingNode>(ind_ptr, mask_ptr),
                graph.emplace_node<AdvancedIndexingNode>(val_ptr, mask_ptr)
                );

        graph.emplace_node<ArrayValidationNode>(put_ptr);

        auto state = graph.empty_state();
        a_ptr->initialize_state(state, {0, 1, 2, 3, 4, 5});
        graph.initialize_state(state);

        WHEN("The indices contain two duplicates with different corresponding values") {
            ind_ptr->set_value(state, 0, 0);
            ind_ptr->set_value(state, 1, 0);

            val_ptr->set_value(state, 0, 10);
            val_ptr->set_value(state, 1, 11);

            mask_ptr->grow(state, {0});
            mask_ptr->grow(state, {1});

            // This is now effectively put([0, 1, 2, 3, 4, 5], [0, 0], [10, 11])

            graph.propagate(state, graph.descendants(state, {val_ptr, ind_ptr, mask_ptr}));
            graph.commit(state, graph.descendants(state, {val_ptr, ind_ptr, mask_ptr}));

            THEN("State is correct") {
                CHECK(std::ranges::equal(put_ptr->view(state), std::vector{11, 1, 2, 3, 4, 5}));

                AND_WHEN("We shrink the mask, shrinking both effective indices and values") {
                    mask_ptr->shrink(state);

                    // This is now effectively put([0, 1, 2, 3, 4, 5], [0], [10])

                    graph.propagate(state, graph.descendants(state, {val_ptr, ind_ptr, mask_ptr}));
                    graph.commit(state, graph.descendants(state, {val_ptr, ind_ptr, mask_ptr}));

                    THEN("First index goes to the other provided value") {
                        // Currently this will fail as it will still output [11, 1, 2, 3, 4, 5]
                        CHECK(std::ranges::equal(put_ptr->view(state), std::vector{10, 1, 2, 3, 4, 5}));
                    }
                }
            }
        }
    }
```

dwave/optimization/mathematical.py
```diff
@@ -519,6 +521,57 @@ def multiply(x1: ArraySymbol, x2: ArraySymbol, *xi: ArraySymbol,
     raise RuntimeError("implementated by the op() decorator")
 
 
+def put(array: ArraySymbol, indices: ArraySymbol, values: ArraySymbol) -> Put:
+    r"""Replace the specified elements in an array with given values.
+
```
wbernoudy, line 526: Probably worth having a note here about expected behavior in the case of duplicate values, even though numpy does not specify this in their docs.

dwave/optimization/src/nodes/manipulation.cpp
arcondello: You're right. Good catch. Will fix.

arcondello: Should be fixed by https://github.com/dwavesystems/dwave-optimization/pull/201/commits/c1f2015a1db8e6420bd1f9d2e708c7c2dd084c3b

dwave/optimization/mathematical.py
arcondello: Should be fixed by https://github.com/dwavesystems/dwave-optimization/pull/201/commits/88134d6571a8e5e8d8161701b98647f779cb8ea3

dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -170,6 +172,247 @@ void ConcatenateNode::revert(State& state) const {
     data_ptr<ArrayNodeStateData>(state)->revert();
 }
 
+// A PutNode needs to track its buffer as well as a mask of which elements in the
+// original array are currently.
```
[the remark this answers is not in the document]
arcondello: Should be fixed by https://github.com/dwavesystems/dwave-optimization/pull/201/commits/b274f0e79076f39d1c167da8e7552bbacff66b03

dwave/optimization/mathematical.py
```diff
@@ -536,7 +536,16 @@ def put(array, indices, values):
 
     Args:
         array: Base array. Must be not be a dynamic array nor a scalar.
-        indices: The indices in the flattened base array to be replaced.
+        indices:
+            The indices in the flattened base array to be replaced.
+
+            .. warning::
+                If ``indices`` has duplicate values, it is undefined which
+                value from ``values`` will be propagated.
```
wbernoudy, line 544:
```suggestion
                of the possible corresponding values from ``values`` will
                be propagated.
```

dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -385,6 +427,10 @@ void PutNode::propagate(State& state) const {
                 }
             }
         }
+
+        if (ptr->num_ambiguities()) {
```
wbernoudy, line 431: Nit but having `num_ambiguities()` and checking it again here feels redundant when you're already checking if `ambiguities_` is empty in `resolve_ambiguities()`.

arcondello: It's because actually constructing the two `views()` actually results in a nontrivial amount of jumping around in memory. I am trying to avoid creating unnecessary objects.

wbernoudy: Ah, I should have read the call here more closely, makes sense.

wbernoudy (approved):

tests/cpp/nodes/test_manipulation.cpp
```diff
@@ -281,6 +281,267 @@ TEST_CASE("ConcatenateNode") {
...
+
+            THEN("Reverting results in no change") {
+                graph.revert(state, graph.descendants(state, {v_ptr}));
+                CHECK(std::ranges::equal(put_ptr->view(state), std::vector{0, 30, 2, 20, 4, 5}));
+                CHECK(std::ranges::equal(put_ptr->mask(state), std::vector{0, 2, 0, 1, 0, 0}));
+            }
+        }
+    }
+
+    GIVEN("A length 6 array of integers, and two integer arrays for indexing and values") {
```
wbernoudy, line 490: What clean code, who wrote this section? :stuck_out_tongue:

arcondello: So beautiful, I even [gave credit](https://github.com/dwavesystems/dwave-optimization/pull/201/commits/c1f2015a1db8e6420bd1f9d2e708c7c2dd084c3b)!

PR #208, opened 2025-01-17
Add philosophy section to the docs as well as navigation

arcondello:


JoelPasvolsky (commented):
This is good. My only thoughts so far are these:
1. A less generic header instead of "Features". Perhaps "Design Principals"?
2. an explicit link to the new page; for example, "`dwave-optimization` and the hybrid nonlinear solver incorporate features and :ref:`design principals <optimization_philosophy>` from each of the following areas:"
3. In the context of a description, the text in the boxes should be descriptive too (e.g., "Exploiting quantum-mechanical effects not available to classical compute" and "Solving optimization problems with linear program solvers")

arcondello:
Agree with the first two. For (3) it's a box we want people to click, so IMO the imperative makes that clearer.

JoelPasvolsky:
> Agree with the first two. For (3) it's a box we want people to click, so IMO the imperative makes that clearer.

I wouldn't understand it that way, if I see "Learn the basics of solving optimization problems with linear program solvers" I'd expect that clicking would move me away from learning about `dwave-optimization` to some LP tutorial. As a new users, I'm here to learn to use the NL solver. Perhaps we can use an icon or something to clarify that clicking will take them to a knowledge resource on the underlying design ideas.

arcondello:
The icons look funny on this PR but it looks fine in the sdk 
image

docs/philosophy.rst
```diff
@@ -0,0 +1,183 @@
...
+`nonlinear <https://en.wikipedia.org/wiki/Nonlinear_programming>`_ relationships
+among variables, constraints, and the objective function.
+
+.. _optimization_philosophy_combinatorial_variables:
+
+Lists, Sets, and other combinatorial variables
+==============================================
+
+While many optimization problems can be expressed using only scalar variable
+types (continuous, integer, and binary), it is often convenient to use other
```
JoelPasvolsky, line 56:
```suggestion
types (continuous, integer, and binary), it is often beneficial to use
```

docs/philosophy.rst
```diff
@@ -0,0 +1,183 @@
...
+combinatorial structures.
+
+In the following example, we will contrast a classic mixed integer linear programming
```
JoelPasvolsky, line 59:
```suggestion
The following example contrasts a classic mixed integer linear programming (MILP)
```

You use the MILP abbreviation below

docs/philosophy.rst
```diff
@@ -0,0 +1,183 @@
...
+formulation using scalar variables with one that uses a list variable.
+
+Example - Traveling Salesperson
```
JoelPasvolsky, line 62:
```suggestion
Example: Traveling Salesperson
```

JoelPasvolsky, line 62:
```suggestion
Traveling-Salesperson Example
```

This option even preserves your hyphen!

docs/philosophy.rst
```diff
@@ -0,0 +1,183 @@
...
+-------------------------------
+
+The goal of renowned
```
JoelPasvolsky, line 65:
```suggestion
The goal of the renowned
```

docs/philosophy.rst
```diff
@@ -0,0 +1,183 @@
...
+`traveling salesperson problem <https://en.wikipedia.org/wiki/Travelling_salesman_problem>`_
+is, for a given a list of cities and distances between each pair of cities, to
```
JoelPasvolsky, line 67:
```suggestion
(TSP) is, for a given a list of cities and distances between each pair of cities, to
```

You use the abbreviation below

docs/philosophy.rst
```diff
@@ -0,0 +1,183 @@
...
+find the shortest possible route that visits each city exactly once and returns
+to the city of origin.
+
+If we wish to restrict ourselves to scalar variables,
+a common mixed-integer linear programming formulation for solving a TSP with
+:math:`N` cities is to use :math:`N^2` binary variables and :math:`2N`
+constraints.
```
JoelPasvolsky, line 74:
```suggestion
When restricted to scalar variables, a common MILP formulation for solving a TSP
with :math:`N` cities uses :math:`N^2` binary variables and :math:`2N`
constraints.
```

docs/philosophy.rst
```diff
@@ -0,0 +1,183 @@
...
+
+.. dropdown:: MILP formulation details
+
+    Let
+
+    .. math::
+
```
JoelPasvolsky, line 81:
```suggestion

```

Trailing spaces

docs/philosophy.rst
```diff
@@ -0,0 +1,183 @@
...
+constraints.
+
+.. dropdown:: MILP formulation details
+
+    Let
+
+    .. math::
+        
+        x_{it} =
+            \begin{cases}
```
JoelPasvolsky, line 83:
```suggestion
            \begin{cases}
```

Trailing space

docs/philosophy.rst
```diff
@@ -0,0 +1,183 @@
...
+is, for a given a list of cities and distances between each pair of cities, to
+find the shortest possible route that visits each city exactly once and returns
+to the city of origin.
+
+If we wish to restrict ourselves to scalar variables,
+a common mixed-integer linear programming formulation for solving a TSP with
+:math:`N` cities is to use :math:`N^2` binary variables and :math:`2N`
+constraints.
+
+.. dropdown:: MILP formulation details
```
JoelPasvolsky, line 76: I ❤️ this use of the dropdown

docs/philosophy.rst
```diff
@@ -0,0 +1,183 @@
...
+        \sum_i x_{it} = 1 \text{  } \forall t \\
+        \sum_t x_{it} = 1 \text{  } \forall i
+
+However, the solution to this problem is defined by an ordered list of cities.
+For example, a three city traveling salesperson problem has
+`(0, 1, 2)`, `(0, 2, 1)`, `(1, 0, 2)`, `(1, 2, 0)`, `(2, 0, 1)`, and `(2, 1, 0)`
+as possible solutions.
+Therefore, rather than thinking about scalars, we can define a *list* variable as
+any permutation of the integers from :math:`[0..N)`.
+Another advantage of this approach is that the problem is unconstrained!
```
JoelPasvolsky, line 109:
```suggestion
Note that the solution to this problem is defined by an ordered list of cities.
For example, a three-city TSP has these possible solutions: `(0, 1, 2)`,
`(0, 2, 1)`, `(1, 0, 2)`, `(1, 2, 0)`, `(2, 0, 1)`, and `(2, 1, 0)`.
Therefore, rather than working with scalars, you can define a *list* variable as
any permutation of the integers from :math:`[0..N)`, significantly reducing the
number of problem variables. Another advantage of this approach is that the
problem is unconstrained!
```

See the alternative below (links to `list()`)

JoelPasvolsky, line 109:
```suggestion
Note that the solution to this problem is defined by an ordered list of cities.
For example, a three-city TSP has these possible solutions: `(0, 1, 2)`,
`(0, 2, 1)`, `(1, 0, 2)`, `(1, 2, 0)`, `(2, 0, 1)`, and `(2, 1, 0)`.
Therefore, rather than working with scalars, you can define a
:meth:`~dwave.optimization.model.Model.list` variable as any permutation of the
integers from :math:`[0..N)`, significantly reducing the number of problem
variables. Another advantage of this approach is that the problem is
unconstrained!
```

Same as the alternative but with a hyperlink to `list()`

docs/philosophy.rst
```diff
@@ -0,0 +1,183 @@
...
+
+.. dropdown:: Formulation with a list variable details
```
JoelPasvolsky, line 111:
```suggestion
.. dropdown:: Formulation with a list variable
```

JoelPasvolsky (commented):
Lovely! Thank you.

docs/philosophy.rst
```diff
@@ -0,0 +1,183 @@
...
+`linear programming <https://en.wikipedia.org/wiki/Linear_programming>`_
+solvers and techniques.
+
+Users familiar with linear programming techniques will find many concepts in
+`dwave-optimization` familiar: objective functions; continuous, linear, and binary variables;
+constraints; and feasible regions are all concepts found in linear programming.
+
+However, unlike linear solvers, `dwave-optimization` allows
+`nonlinear <https://en.wikipedia.org/wiki/Nonlinear_programming>`_ relationships
+among variables, constraints, and the objective function.
```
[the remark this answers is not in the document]
arcondello: You can have a nonlinear relationship among the constraints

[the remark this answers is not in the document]
JoelPasvolsky: Good point. Still need to revise so as not to say "relationships among ... the objective function" (plural vs singular)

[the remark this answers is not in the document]
arcondello:
```rst
However, unlike linear solvers, `dwave-optimization` allows
`nonlinear <https://en.wikipedia.org/wiki/Nonlinear_programming>`_ relationships
among variables, constraints, and intermediate values such as the objective value.
```

docs/philosophy.rst
```diff
@@ -0,0 +1,183 @@
...
+as possible solutions.
+Therefore, rather than thinking about scalars, we can define a *list* variable as
+any permutation of the integers from :math:`[0..N)`.
+Another advantage of this approach is that the problem is unconstrained!
+
+.. dropdown:: Formulation with a list variable details
+
+    Let `x` be a list variable of length `N` with `x_t` giving
+    the city the salesperson visits at time `t`. Let's define `x_N = x_0` for
+    convenience.
```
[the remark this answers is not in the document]
arcondello: One tick is math, no? But sure I can be explicit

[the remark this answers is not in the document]
arcondello: It is implicitly in the sum. 
```
\sum_t d_{x_t, x_{t+1}}
```
doesn't explicitly include the `\sum_t^N` but a careful reader would notice that `x_{t+1}` would go "out of bounds"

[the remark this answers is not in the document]
JoelPasvolsky: image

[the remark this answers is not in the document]
JoelPasvolsky: Maybe make it explicit to clarify what the "Let's define... " refers to?

docs/philosophy.rst
```diff
@@ -0,0 +1,183 @@
...
+
+    a = [...]
+    b = [...]
+    value = sum(u * b for u, v in zip(a, b))
+
+But it is more readable and more performant to express it using array operations
+
+.. code-block:: python
+
+    a = np.asarray([...])
```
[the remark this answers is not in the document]
arcondello: I've dropped all of the imports in these examples for clarity. IMO if the user doesn't recognize the `np.` convention then this section is not for them :slightly_smiling_face:

docs/philosophy.rst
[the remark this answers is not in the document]
arcondello: I'll move it to make it clearer.

JoelPasvolsky (approved):
Looks good!

docs/philosophy.rst
```diff
@@ -0,0 +1,182 @@
+.. _optimization_philosophy:
+
+==========
+Philosophy
+==========
+
+`dwave-optimization` and
+the `Leap <https://cloud.dwavesys.com/leap/>`_\ |TM| service's quantum-classical
```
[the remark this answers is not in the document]
arcondello: :shrug:

PR #209, opened 2025-01-17
Add rint node

elham-azd:


dwave/optimization/mathematical.py
```diff
@@ -583,6 +585,27 @@ def put(array, indices, values):
     """
     return Put(array, indices, values)
 
+def rint(x: ArraySymbol) -> RoundInt:
+    r"""Return an element-wise rint on the given symbol.
```
arcondello, line 589: Don't need the leading `r` in this case. Also, could use a bit more spacing. `sphinx` sometimes gets confused without it.

arcondello (approved):
There are also a few other places of small clarifications in the docs and alphabetizations. But otherwise looks good to me!

PR #218, opened 2025-01-30
Improve QOL when reshaping array symbols

arcondello:
Adds flattening array, reshaping non-contiguous arrays, the `Copy` symbol (see https://github.com/dwavesystems/dwave-optimization/issues/16), and allows reshaped arrays to be used as indexers.


tests/cpp/nodes/test_manipulation.cpp
```diff
@@ -278,6 +278,110 @@ TEST_CASE("ConcatenateNode") {
...
+            CHECK(c_ptr->max() == y_ptr->max());
+            CHECK(c_ptr->min() == y_ptr->min());
+            CHECK(c_ptr->integral() == y_ptr->integral());
+        }
+
+        WHEN("We mutate the state of x") {
+            x_ptr->grow(state);
+            graph.propagate(state, graph.descendants(state, {x_ptr}));
+
+            THEN("c has the same shape as y and the same values") {
```
samuelbodin, line 362: Is the shape check captured implicitly by the state check? Here and in commit/revert below

dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -442,8 +486,34 @@ void PutNode::propagate(State& state) const {
...
+// We do that inference here.
+std::vector<ssize_t> infer_reshape(Array* array_ptr, std::vector<ssize_t>&& shape) {
+    // if the base array is dynamic, we might allow the first dimension to be negative
+    // 1. So let's defer to the various constructors.
+    if (array_ptr->dynamic()) return shape;
+
+    // Check if there are any -1s, and if not fallback to other input checking.
+    auto it = std::ranges::find(shape, -1);
+    if (it == shape.end()) return shape;
+
```
wbernoudy, line 499:
```suggestion

// Ensure there is at most one -1
if (++it != shape.end()) {
    // Same error message as NumPy
    throw std::invalid_argument("can only specify one unknown dimension");
}
```

arcondello: This actually gets checked by the `ArrayOutputMixin` because that doesn't allow any `-1`. So I am deferring the error to there. The error message is slightly less explicit, but IMO it's OK.

wbernoudy: I think it checks that it's not dynamic, i.e. the first index is not negative, but that won't catch the case of multiple -1s...

wbernoudy: Actually, I don't think it gets checked at all. Just returns `DYNAMIC_SIZE` if the first index is negative. So then the `this->dynamic()` check below would return true and throw an error if there are an even number of -1s in the shape, but not if there are an odd number

arcondello: Let me add a test case

arcondello: 4442ba1 adds a test case, it does correctly throw.

arcondello: It's because it results in a negative or otherwise non-matching shape. So it fails on the check for consistency between `Reshape`'s shape and its predecessor's shape.

arcondello: Ok, I found an edge case, `ReshapeNode(&A, {12, -1, -1}`. Will fix.

arcondello: Fixed by 6439503

wbernoudy (approved):

PR #219, opened 2025-01-30
Add missing methods integral, max, min for ConcatenateNode

samuelbodin:


dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -77,6 +77,12 @@ void ConcatenateNode::initialize_state(State& state) const {
     state[index] = std::make_unique<ArrayNodeStateData>(std::move(values));
 }
 
+bool ConcatenateNode::integral() const {
+    return std::accumulate(std::next(array_ptrs_.begin()), array_ptrs_.end(),
```
arcondello, line 81: Might be simpler and more explicit to do
```c++
std::ranges::all_of(array_ptrs_, [](ArrayNode* ptr){ return ptr->integral(); });
```

dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -136,6 +142,18 @@ std::vector<ssize_t> make_concatenate_shape(std::span<ArrayNode*> array_ptrs, ss
     return shape;
 }
 
+double ConcatenateNode::max() const {
+    return std::accumulate(std::next(array_ptrs_.begin()), array_ptrs_.end(),
```
arcondello, line 146: I think we can use https://en.cppreference.com/w/cpp/algorithm/ranges/max
```c++
std::ranges::max(array_ptrs_, {}, [](ArrayNode* ptr){ return ptr->max(); });
```
similar for min

arcondello (commented):
Also needs a release note

dwave/optimization/src/nodes/manipulation.cpp
samuelbodin: Nice!

dwave/optimization/src/nodes/manipulation.cpp
samuelbodin: How do you get the value from the range? Using this on its own doesn't work

samuelbodin: Ok this seems to work

```
return std::ranges::max(
              array_ptrs_ | std::views::transform([](ArrayNode* ptr) { return ptr->max(); }));
```

arcondello (approved):

dwave/optimization/src/nodes/manipulation.cpp
arcondello: Oh weird, I was sure that would work. Glad you found something.

PR #221, opened 2025-01-31
Simple simplex implementation

wbernoudy:
First pass so we can at least start to see what a good interface will look like. <s>Appears to be correct based on the three tiny tests I wrote :smile: </s>

Tested on the [netlib lb instances](https://github.com/scipy/scipy/tree/06712cfa85033d445408f9ee898e4954375a0aa1/benchmarks/benchmarks/linprog_benchmark_files) including the infeasible problems. Results match `scipy.optimize.linprog(method="simplex")`. Reasonably confident that it is correct.

### DIFF
--- a/dwave/optimization/src/simplex.cpp
+++ b/dwave/optimization/src/simplex.cpp
@@ -1,4 +1,4 @@
-// Copyright 2025 D-Wave Systems Inc.
+// Copyright 2025 D-Wave
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
--- a/dwave/optimization/src/simplex.cpp
+++ b/dwave/optimization/src/simplex.cpp
@@ -196,7 +196,7 @@ SolveResult _solve_simplex(Matrix& T, ssize_t n, std::vector<ssize_t>& basis,
 Matrix construct_T(std::span<const double> c, double c0, const Matrix& A,
                    std::span<const double> b) {
     Matrix T(A.n() + 2, A.m() + A.n() + 1);
-    // Copy in A to T[:n, :m] and b.T to A[:, -1]
+    // Copy in A to T[:n, :m] and b.T to T[:, -1]
     for (ssize_t i = 0; i < A.n(); i++) {
         double sign = b[i] < 0 ? -1.0 : 1.0;
         T(i, -1) = sign * b[i];
@@ -214,7 +214,7 @@ Matrix construct_T(std::span<const double> c, double c0, const Matrix& A,
     for (ssize_t i = 0; i < A.m(); i++) {
         T(A.n(), i) = c[i];
     }
-    T(A.n(), T.m() - 1) = c0;
+    T(A.n(), -1) = c0;
 
     // Row pseudo objective
     for (ssize_t j = 0; j < A.m(); j++) {
@@ -223,7 +223,7 @@ Matrix construct_T(std::span<const double> c, double c0, const Matrix& A,
         }
     }
     for (ssize_t i = 0; i < A.n(); i++) {
-        T(A.n() + 1, T.m() - 1) -= T(i, T.m() - 1);
+        T(-1, -1) -= T(i, -1);
     }
 
     return T;
@@ -310,7 +310,7 @@ void check_LP_sizes(std::span<const double> c, std::span<const double> b_lb,
 }
 
 /// Translate the general LP form to the simple:
-///     minimize(c @ x) subject to A @ x == b
+///     minimize(c @ x) subject to A @ x == b, x >= 0
 LP translate_LP_to_simple(std::span<const double> c, std::span<const double> b_lb,
                           std::span<const double> A_data, std::span<const double> b_ub,
                           std::span<const double> A_eq_data, std::span<const double> b_eq,
--- a/dwave/optimization/src/simplex.cpp
+++ b/dwave/optimization/src/simplex.cpp
@@ -300,7 +300,7 @@ void check_LP_sizes(std::span<const double> c, std::span<const double> b_lb,
                     std::span<const double> A_data, std::span<const double> b_ub,
                     std::span<const double> A_eq_data, std::span<const double> b_eq,
                     std::span<const double> lb, std::span<const double> ub) {
-    const ssize_t num_vars = c.size();
+    [[maybe_unused]] const ssize_t num_vars = c.size();
 
     assert(b_lb.size() == b_ub.size() && "b_lb length does not match b_ub length");
     assert(num_vars * b_lb.size() == A_data.size() && "A_data wrong size");
--- a/dwave/optimization/src/simplex.cpp
+++ b/dwave/optimization/src/simplex.cpp
@@ -324,17 +324,16 @@ LP translate_LP_to_simple(std::span<const double> c, std::span<const double> b_l
 
     static double inf = std::numeric_limits<double>::infinity();
 
-    ssize_t lower_bounded_constraint_count = 0;
-    ssize_t upper_bounded_constraint_count = 0;
+    ssize_t A_constraint_count = 0;
     for (ssize_t i = 0; i < A.n(); i++) {
         // Unbounded constraint, can ignore
         if (b_lb[i] == -inf && b_ub[i] == inf) continue;
 
         if (b_lb[i] != -inf) {
-            lower_bounded_constraint_count++;
+            A_constraint_count++;
         }
         if (b_ub[i] != inf) {
-            upper_bounded_constraint_count++;
+            A_constraint_count++;
         }
     }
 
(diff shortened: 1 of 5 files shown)


dwave/optimization/src/simplex.hpp
```diff
@@ -0,0 +1,54 @@
...
+struct SolveResult {
+    SolveResult(SolverStatus solver_status)
+            : status(solver_status.status), nit(solver_status.nit){};
+
+    solve_status_t status;
+    ssize_t nit;
+    std::vector<double> solution;
+};
+
+SolveResult linprog(const std::vector<double>& c, const std::vector<double>& b_lb,
```
arcondello, line 49: I think `span` helps to make it clear that we only intend to use it as a view and to not resize. I know that's implicit to `const` but stylistically I think it's clearer.

tests/cpp/test_simplex.cpp
```diff
@@ -0,0 +1,107 @@
+// Copyright 2025 D-Wave Inc.
```
arcondello, line 1:
```suggestion
// Copyright 2025 D-Wave
```
Nice job on making all three headers different! :laughing:

dwave/optimization/src/simplex.cpp
```diff
@@ -0,0 +1,493 @@
...
+            : n_(n), m_(m), buffer_(std::move(data)) {
+        if (n_ * m_ != static_cast<ssize_t>(buffer_.size())) {
+            throw std::invalid_argument("shape does not match provided data");
+        }
+    }
+
+    Matrix(const std::vector<double>& data, ssize_t n, ssize_t m)
+            : Matrix(std::vector<double>(data), n, m) {}
+
+    double& at(ssize_t i, ssize_t j) {
```
arcondello, line 42: You're using `at()` because `operator[]()` can't take two arguments? I don't love `.at()` specifically because in the stdlib that implies it might raise, but here it only asserts.

I think doing what Eigen does and making these `double& operator()(ssize_t n, ssize_t m)` makes sense to me.

dwave/optimization/src/simplex.cpp
```diff
@@ -0,0 +1,493 @@
...
+    ssize_t n_;
+    ssize_t m_;
+
+    std::vector<double> buffer_;
+};
+
+ssize_t _pivot_col(Matrix& T, double tol, bool bland) {
+    double min_val = std::numeric_limits<double>::infinity();
+    ssize_t col = -1;
+    for (ssize_t j = 0; j < T.m() - 1; j++) {
```
arcondello, line 83:
```c++
    for (ssize_t j = 0, stop = T.m() - 1; j < stop; ++j) {
```

dwave/optimization/src/simplex.cpp
```diff
@@ -0,0 +1,493 @@
...
+            // Bland's rule: return the first index with negative value
+            if (bland) return j;
+
+            if (val < min_val) {
+                min_val = val;
+                col = j;
+            }
+        }
+    }
+    return col;
```
arcondello, line 95: Maybe toss an `assert(T.m() * T.n() == 0 || col >= 0)` here? Or, if we don't want to allow empty `Matrix`, then assert that the matrix has nonzero elements.

Applies to `_pivot_row()` as well.

arcondello (commented):
A couple stylistic/debugging comments. The algorithm looks correct to me. Though I am far from an expert on simplex.

dwave/optimization/src/simplex.cpp
wbernoudy: These functions return `-1` to mean that no row/column was found. I'm fine to change this to something like using `std::optional` to make it clearer though.

dwave/optimization/src/simplex.hpp
```diff
@@ -0,0 +1,54 @@
...
+enum solve_status_t {
+    UNSET,
+    SUCCESS,
+    HIT_ITERATION_LIMIT,
+    FAILURE_NO_FEASIBLE_START,
+    FAILURE_UNBOUNDED,
+    FAILURE_SINGULAR_MATRIX,
+};
+
+struct SolverStatus {
```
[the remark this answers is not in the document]
wbernoudy: I exposed these so that we can test the method easily. And yes, I think both things make sense.

tests/cpp/test_simplex.cpp
wbernoudy: Gotta keep reviewers on their toes.

dwave/optimization/src/simplex.cpp
arcondello: No, I think `-1` make sense. But I would definitely mention it in a comment to the intent is clear.

dwave/optimization/src/simplex.cpp
samuelbodin: Another way is to overload `[]` to return an object of a subclass that also overloads `[]`. But  you get `[i][j]` and not `[i,j]`.

dwave/optimization/src/simplex.cpp
```diff
@@ -0,0 +1,493 @@
...
+
+ private:
+    ssize_t n_;
+    ssize_t m_;
+
+    std::vector<double> buffer_;
+};
+
+ssize_t _pivot_col(Matrix& T, double tol, bool bland) {
+    double min_val = std::numeric_limits<double>::infinity();
```
samuelbodin, line 81: Would `std::numeric_limits<double>::max()` work as well? Curious to know the semantic difference between the two

dwave/optimization/src/simplex.hpp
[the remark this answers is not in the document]
wbernoudy: Actually, on 1., just going to leave it as bare `std::vector` as the empty solution is only passed around a handful of times so the overhead of vector is trivial and not worth the annoyance of `std::optional`.

dwave/optimization/src/simplex.cpp
wbernoudy: That is another good route, and I might end up doing something like that if we need row/column objects/iterators. But I think for now the simplicity of `operator()` is preferable.

dwave/optimization/src/simplex.cpp
wbernoudy: I believe `max()` will specifically return the largest _finite_ value, e.g. for doubles it's 1.18973e+4932. That should basically be fine here as well, other than the very extreme case that all the values are equal to 1.18973e+4932, which now that I think of it is impossible anyway given the logic of the for loop.

dwave/optimization/src/simplex.cpp
```diff
@@ -0,0 +1,563 @@
...
+            }
+        }
+    }
+    return status;
+}
+
+Matrix construct_T(std::span<const double> c, double c0, const Matrix& A,
+                   std::span<const double> b) {
+    Matrix T(A.n() + 2, A.m() + A.n() + 1);
+    // Copy in A to T[:n, :m] and b.T to A[:, -1]
```
alexzucca90, line 199: typo: .. `and b.T to T[:, -1]`

dwave/optimization/src/simplex.cpp
```diff
@@ -0,0 +1,563 @@
...
+}
+
+void _apply_pivot(Matrix& T, std::vector<ssize_t>& basis, ssize_t pivrow, ssize_t pivcol,
+                  double tolerance) {
+    basis[pivrow] = pivcol;
+
+    double pivval = T(pivrow, pivcol);
+
+    for (ssize_t j = 0; j < T.m(); j++) {
+        T(pivrow, j) /= pivval;
```
mdcoury, line 123: If T.m() can be large, might be best to store the reciprocal of pivval outside the loop and multiply by it.

dwave/optimization/src/simplex.cpp
```diff
@@ -0,0 +1,563 @@
...
+            solution[basis[i]] = T(i, -1);
+        }
+    }
+    result.set_partial_solution(std::move(solution));
+
+    return result;
+}
+
+struct LP {
+    LP(std::vector<double>&& c, double c0, Matrix&& A, std::vector<double>&& b)
```
mdcoury, line 290: Learning opportunity for me: why pass in the 'reference to a reference' here?  I'm not following how that works with, what looks to me like, the vector's copy-constructor...

arcondello: This is an [Rvalue reference](https://en.cppreference.com/w/cpp/language/reference). Basically it's saying that it must take ownership of the vector.

tests/cpp/test_simplex.cpp
```diff
@@ -0,0 +1,195 @@
...
+#include <catch2/matchers/catch_matchers_floating_point.hpp>
+
+#include "catch2/catch_test_macros.hpp"
+#include "simplex.hpp"
+
+namespace dwave::optimization {
+
+using Catch::Matchers::WithinRel;
+
+TEST_CASE("LP solver (simplex)", "[simplex]") {
```
samuelbodin, line 25: What is the `[simplex]` parameter in the `TEST_CASE` expression? Is it like a tag?

dwave/optimization/src/simplex.cpp
wbernoudy: Actually, given that this is more of a reference implementation and not focused on performance in any way, I'm inclined to leave it as division. The other upside of using actual division is that the floating point error remains smaller, which may help with numerical stability of the overall algorithm.

arcondello (approved):

PR #222, opened 2025-02-05
Implement `axis` kwarg for `ArraySymbol.prod()`

arcondello:


dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -847,35 +1006,99 @@ PartialReduceNode<BinaryOp>::PartialReduceNode(ArrayNode* array_ptr, const ssize
...
+                                              RunningProduct, double>::type;
+
+    std::vector<accumulator_type> values(size(state));
+
+    const ssize_t axis = axes_[0];
+
+    for (ssize_t index = 0, stop = size(state); index < stop; ++index) {
+        assert(index < static_cast<ssize_t>(values.size()));
+
+        /// We wish to fill `index` of the partial reduction, given the state of the parent.
```
arcondello, line 1041: This is just the old `reduce()` function moved inside the loop. Because we do some templated type fiddling. But the logic is all the same with the exception of the `constexpr if` later on.

dwave/optimization/_model.pyx
```diff
@@ -1338,13 +1338,26 @@ cdef class ArraySymbol(Symbol):
...
         """
-        from dwave.optimization.symbols import Prod  # avoid circular import
-        return Prod(self)
+        import dwave.optimization.symbols
+
+        if axis is not None:
+            if not isinstance(axis, numbers.Integral):
+                raise TypeError("axis of the prod should be an int")
+
+            if not (0 <= axis < self.ndim()):
```
wbernoudy, line 1355: Is there a reason we wouldn't allow `axis=-1` on these?

dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -847,35 +1006,99 @@ PartialReduceNode<BinaryOp>::PartialReduceNode(ArrayNode* array_ptr, const ssize
...
+            assert(array_ptr_->size(state) >= 1);
+            init = *begin;
+            ++begin;
+        }
 
-    for (ssize_t i = 0, stop = size(state); i < stop; ++i) {
-        assert(i < static_cast<ssize_t>(values.size()));
-        values[i] = reduce(state, i);
+        if constexpr (std::same_as<BinaryOp, std::multiplies<double>>) {
+            // we use accumulate rather than reduce do we have an easier lambda
```
wbernoudy, line 1088:
```suggestion
            // we use accumulate rather than reduce so we have an easier lambda
```

dwave/optimization/src/nodes/mathematical.cpp
wbernoudy: I guess the `reduce()` function wasn't used anywhere but `initialize_state()`?

arcondello: Yeah exactly.

wbernoudy (approved):
LGTM

arcondello:
Lol, oops, 
image
ah well, it's staying.

PR #224, opened 2025-02-06
Use catch2 v3.8.0 for C++ testing

arcondello:
The upshot is that the newer `Catch2`s let you do
```c++
CHECK_THAT(lhs, RangeEquals({0, 1, 2}));
```
rather than
```c++
CHECK(std::ranges::equal(lhs, std::vector{0, 1, 2}));
```
this is a bit more succinct, and it actually gives a useful error message when they don't match!

```../tests/cpp/nodes/mathematical/test_binaryop.cpp:228: FAILED:
  CHECK_THAT( le_ptr->shape(), RangeEquals({5, 7}) )
with expansion:
  { 5 } elements are { 5, 7 }
```

This PR has no functional changes, but IMO makes for much cleaner test code.

I am also sure I missed a few, but not trying to be 100% complete.

### DIFF
--- a/subprojects/catch2.wrap
+++ b/subprojects/catch2.wrap
@@ -1,10 +1,10 @@
 [wrap-file]
-directory = Catch2-3.7.1
-source_url = https://github.com/catchorg/Catch2/archive/v3.7.1.tar.gz
-source_filename = Catch2-3.7.1.tar.gz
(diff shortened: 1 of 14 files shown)


wbernoudy (approved):
Awesome

PR #227, opened 2025-02-07
Add symbol_substitution param to from_file

wbernoudy:
Just testing this out currently


dwave/optimization/_model.pyx
```diff
@@ -110,6 +110,7 @@ cdef class _Graph:
     @classmethod
     def from_file(cls, file, *,
                   check_header = True,
+                  symbol_substitution = None,
```
arcondello, line 113: nit: `substitute`

PR #228, opened 2025-02-11
Improve the documentation of `Model.state_size()` and similar methods

arcondello:


dwave/optimization/_model.pyx
```diff
@@ -93,17 +93,39 @@ cdef class _Graph:
         return value
 
     def decision_state_size(self):
-        r"""An estimated size, in bytes, of the model's decision states.
+        """Return an estimate of the size, in bytes, of a model's decision states.
```
JoelPasvolsky, line 96:
```suggestion
        r"""Return an estimate of the size, in bytes, of a model's decision states.
```

Currently you get:
image

With this you'll get:
image

dwave/optimization/_model.pyx
```diff
@@ -93,17 +93,39 @@ cdef class _Graph:
...
+
+        For more details, see :meth:`.state_size()`.
+        This method differs by only counting the state of the decision variables.
```
JoelPasvolsky, line 99:
```suggestion
        This method differs by counting the state of only the decision variables.
```

The rational here is that when you read "This method differs by only" you expect the diff ("not counting non-decision variables" but you get what the updated method does.

dwave/optimization/_model.pyx
```diff
@@ -619,19 +641,46 @@ cdef class _Graph:
...
+            :meth:`Symbol.state_size()` An estimate of the size of a symbol's
+            state.
+
+            :meth:`ArraySymbol.state_size()` An estimate of the size of an array
+            symbol's state.
+
+            :meth:`Model.decision_state_size()` An estimate of the size of a
+            model's decision states.
+
+            :ref:`properties_solver_properties` The properties of the
```
JoelPasvolsky, line 680: Minor (ignore if you wish): Without a colon after the "[Solver Properties](https://docs.dwavesys.com/docs/latest/c_solver_properties.html)" this reads a bit strange. You can change to something like 

```
The  :ref:`properties  documentation \<properties_solver_properties\>` for the
```

(where the backslashes are due to github rendering) or add a colon

dwave/optimization/_model.pyx
```diff
@@ -619,19 +641,46 @@ cdef class _Graph:
...
+            `Leap <https://cloud.dwavesys.com/leap/>`_ service's
+            quantum-classical hybrid nonlinear solver. Including limits on
+            the maximum state of a model.
```
JoelPasvolsky, line 683:
```suggestion
            the maximum state size of a model.
```

JoelPasvolsky (commented):
LGTM, thanks, just a missing raw directive

dwave/optimization/_model.pyx
arcondello: For this one I am incline to ignore. It keeps the pattern of the previous ones and IMO is clear enough in context.

JoelPasvolsky (approved):
Thanks!

Issue #182, opened 2024-12-11
Consider adding a depth limit and/or caching to `Array::min()`, `Array::max()`, etc


arcondello:
Right now calling `::max()` on an array symbol can trigger a dfs. It would be usful to have some way of improving that. Either memoizing, e.g.
```c++
    virtual double min(std::unordered_map<const Array*, double> memo) const {
        auto it = memo.find(this);
        if (it != memo.end()) return it->second;
        auto min_ = this->min();
        memo.emplace_hint(it, this, min_);
        return min_;
    }
```
or setting a depth limit
```c++
    virtual double min(ssize_t depth = -2) const {
        if (depth == -1) return inf;
        ...
    }
```
or both or something else.

arcondello:
For the `memo` approach, we either need to add a second overload for all `::min()`, `::max()`, and `::integral()` methods, or we end up doing something like
```c++
virtual double min(std::optional<std::reference_wrapper<std::unordered_map<const Array*, double>>> memo = std::nullopt) {
    ...
}
```
which gives us a similar API to the `memo` argument to [Python's deepcopy](https://docs.python.org/3/library/copy.html#object.__deepcopy__) but gosh that's an unreadable signature. Or we wait until C++26 when the committee is next considering `optional<T&>`... Or we keep thinking.

arcondello:
Another issue is that many nodes actually use their predecessor's `::max()` method, so we need the cache to hold both the min and max.

PR #229, opened 2025-02-12
Add `Array::minmax()` method

arcondello:
Changes the ways that C++ ``Array`` subclasses propagate their minimum and maximum possible values. Nodes now should implement a single ``::minmax()`` method. That method is then called from ``Array::min()`` and ``Array::max()``.

The ``::minmax()`` function also optionally accepts a ``cache`` argument. It can be passed in using the `cache_type` type alias. E.g.,
```c++
Array::cache_type<std::pair<double, double>> cache;
const auto [low, high] = node_ptr->minmax(cache);
```

Closes https://github.com/dwavesystems/dwave-optimization/issues/182

### DIFF
--- a/dwave/optimization/include/dwave-optimization/array.hpp
+++ b/dwave/optimization/include/dwave-optimization/array.hpp
@@ -895,6 +895,7 @@ class Array {
     // If the ``cache.has_value()`` returns ``false``, then the cache is ignored
     // entirely.
     template <class T, class Func>
+    requires (std::same_as<std::invoke_result_t<Func>, T>)
     T memoize(optional_cache_type<T> cache, Func&& func) const {
         // If there is no cache, then just evaluate the function
         if (!cache.has_value()) return func();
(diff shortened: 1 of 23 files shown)


dwave/optimization/include/dwave-optimization/array.hpp
```diff
@@ -880,6 +890,35 @@ class Array {
         return true;
     }
 
+    // Return a cached value if available, and otherwise return the value returned
+    // by ``func()``.
+    // If the ``cache.has_value()`` returns ``false``, then the cache is ignored
+    // entirely.
+    template <class T, class Func>
+    requires (std::same_as<std::invoke_result_t<Func>, T>)
```
alexzucca90, line 898: beautiful

alexzucca90 (approved):

dwave/optimization/include/dwave-optimization/array.hpp
```diff
@@ -880,6 +890,35 @@ class Array {
         return true;
     }
 
+    // Return a cached value if available, and otherwise return the value returned
+    // by ``func()``.
+    // If the ``cache.has_value()`` returns ``false``, then the cache is ignored
+    // entirely.
+    template <class T, class Func>
```
alexzucca90, line 897: decided against
```cpp
  template <class T, std::invocable Func>
```
?

arcondello: It's implicit to `std::invoke_result_t` so decided to leave it out.

PR #230, opened 2025-02-13
Flatten Symbol and ArraySymbol docs

arcondello:
See also: https://github.com/dwavesystems/dwave-optimization/pull/73, https://github.com/dwavesystems/dwave-optimization/pull/183. This should be the last of that series.

### DIFF
--- a/docs/reference/symbols.rst
+++ b/docs/reference/symbols.rst
@@ -4,64 +4,40 @@
 Symbols
 =======
 
+.. currentmodule:: dwave.optimization.model
+
 Symbols are a model's decision variables, intermediate variables, constants, 
 and mathematical operations.
 
 See :ref:`Symbols <intro_optimization_symbols>` for an introduction to
 working with symbols.
 
-Base Classes
-============
-
-.. currentmodule:: dwave.optimization.model
+All symbols listed in the :ref:`Model Symbols<symbols_model_symbols>`
+section below inherit from the :class:`Symbol` class and, for most mathematical
+symbols, the :class:`ArraySymbol` class.
 
 .. _symbols_base_symbols:
 
+Symbol
+======
+
+All symbols inherit from the :class:`Symbol` class and therefore inherit its
+methods.
+
 .. autoclass:: Symbol
+    :members:
+    :member-order: bysource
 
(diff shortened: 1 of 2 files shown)


docs/reference/symbols.rst
```diff
@@ -10,58 +10,31 @@ and mathematical operations.
...
-    ~ArraySymbol.maybe_equals
-    ~ArraySymbol.min
-    ~ArraySymbol.ndim
-    ~ArraySymbol.prod
-    ~ArraySymbol.reshape
-    ~ArraySymbol.sum
-    ~ArraySymbol.shape
-    ~ArraySymbol.size
-    ~ArraySymbol.state
-    ~ArraySymbol.state_size
```
JoelPasvolsky, line 63: Unrelated to this PR (it was [this one](https://github.com/dwavesystems/dwave-optimization/pull/228)): missing colon

image

Feel free to leave this for a separate PR

JoelPasvolsky (approved):
This is great!

My one concern is that, for someone new coming to this page, seeing at the top the three Symbol, ArraySymbol, and Model Symbols, it won't be immediately clear what each of these are. 

A short line that explains that the first two are the classes and methods inherited by the model symbols would go a long way. 

Would it be clearer to  put Model Symbols at the top? It is the easiest to grasp.

docs/reference/symbols.rst
arcondello: Yeah, let's fix that separately.

arcondello:
> A short line that explains that the first two are the classes and methods inherited by the model symbols would go a long way.

Makes sense, though I am not 100% sure I understand where you want it? If you give my some text and tell me where I can add it.

>  Would it be clearer to put Model Symbols at the top? It is the easiest to grasp.

IMO the first two are important context for the model symbols, so I'd keep the order as-is.

docs/reference/symbols.rst
JoelPasvolsky: https://github.com/dwavesystems/dwave-optimization/pull/231

JoelPasvolsky:
> If you give me some text and tell me where, then I can add it.

Something like this:
image

PR #231, opened 2025-02-13
Fix missing colon

JoelPasvolsky:


dwave/optimization/_model.pyx
```diff
@@ -1561,7 +1561,7 @@ cdef class ArraySymbol(Symbol):
             This example returns the size of an integer symbol.
             In this example, the symbol encodes a :math:`5\times4` of integers,
             each represented by a :math`8` byte float.
```
arcondello, line 1563: You missed this one (I too was making a PR but you got there first :smile: )

arcondello (approved):

dwave/optimization/_model.pyx
JoelPasvolsky: well you were first in missing it 😁 
Thanks!

PR #232, opened 2025-02-18
Add `ARange` symbol and support for slicing by integers

arcondello:


dwave/optimization/src/nodes/creation.cpp
```diff
@@ -0,0 +1,309 @@
...
+    if (start_old != start_new || step_old != step_new || ptr->size() == 0) {
+        ptr->assign(arange(state, start_, stop_, step_));
+        if (ptr->diff().size()) Node::propagate(state);
+        return;
+    }
+
+    // if the stop has also not changed, then nothing to do
+    if (stop_old == stop_new) return;
+
+    // only the stop has changed, and it has actually changed
```
wbernoudy, line 256: Not sure what this comment means

dwave/optimization/src/nodes/creation.cpp
```diff
@@ -0,0 +1,309 @@
...
+            }
+        } else {
+            // we shrank
+            while (buffer.size() && buffer.back() >= stop_new) {
+                ptr->pop_back();
+            }
+        }
+    } else if (step < 0) {
+        if (stop_old < stop_new) {
+            // we shank
```
wbernoudy, line 277: Typo

wbernoudy (approved):
LGTM

dwave/optimization/include/dwave-optimization/nodes/creation.hpp
```diff
@@ -0,0 +1,129 @@
...
+    /// @copydoc Array::size()
+    ssize_t size(const State& state) const override;
+
+    /// @copydoc Array::sizeinfo()
+    SizeInfo sizeinfo() const override;
+
+    /// @copydoc Array::size_diff()
+    ssize_t size_diff(const State& state) const override;
+
+    array_or_int start() const { return start_; }
```
arcondello, line 119: Needs docs

dwave/optimization/src/nodes/creation.cpp
arcondello: Referring to the check two lines above. Agree it's not clear. I will reword it.

dwave/optimization/src/nodes/creation.cpp
arcondello: In hindsight this is very funny

dwave/optimization/mathematical.py
```diff
@@ -117,6 +119,38 @@ def add(x1: ArraySymbol, x2: ArraySymbol, *xi: ArraySymbol) -> typing.Union[Add,
...
+    """
+
+    if stop is None and step is None:
+        # then start is actually stop
+        start, stop = None, start
+
+    if start is None:
+        start = 0
+    if stop is None:
+        raise ValueError  # todo: error message
```
arcondello, line 147: Oops, I need to actually write this error message

dwave/optimization/src/nodes/creation.cpp
```diff
@@ -217,12 +222,12 @@ std::pair<double, double> ARangeNode::minmax(
         // The direction that we're stepping determines the min/max
         if (step_low > 0) {
             const double low = start_low;
-            const double high = stop_high;
+            const double high = stop_high - step_low;
```
wbernoudy, line 225: Why did you add the `- step_low` to this? If I have
```
start_low, start_high = 0, 0
stop_low, stop_high = 5, 5
step_low, step_high = 2, 4
```
Then the ARangeNode `high` should be 4, right? But this would set it to `stop_high - step_low = 3`.

dwave/optimization/src/nodes/creation.cpp
boothby: probably good arbitrary reacts aren't supported in github, but :hocho:

dwave/optimization/src/nodes/creation.cpp
arcondello: You're right. It was an attempt to account for the non-inclusive stop, but it doesn't cover all cases. Will fix.

wbernoudy (approved):
LGTM round 2 :+1:

Issue #213, opened 2025-01-22
Consider adding a testing module


arcondello:
Could include:
* Importable versions of the symbols tests currently found here: https://github.com/dwavesystems/dwave-optimization/blob/5d4002768ae893dd0a304d541c38a48289def311/tests/test_symbols.py#L43-L292
* Generator for small pathological models. Something like https://github.com/dwavesystems/dimod/blob/ac4dcbb7424db6b6a14448ec84f0f04bbbd8b80f/dimod/testing/sampler.py#L76
* Generator for a model with all symbols. Useful for smoke tests.

wbernoudy:
This will also allow us to keep the symbol tests in a separate module without the [annoying workaround here](https://github.com/dwavesystems/dwave-optimization/blob/6befd52a7700c062aea1a7bfb3276a89401707fd/tests/test_symbols.py#L43)

PR #233, opened 2025-02-19
Move symbol unit tests to separate file

wbernoudy:
This moves the `utils` "namespace" from `tests/test_symbols.py` <s>and exposes it in a `dwave.optimization.testing` module.</s> to `tests/utils.py`. No functional change.

Also did some non-functional tidyings in the symbol tests.

<s>Progress on #213</s>


arcondello:
The tidying is an improvement, but I am hesitant to make this pattern publicly available. I think the tests overall need a pretty big refactor before we can include them in the package.

tests/test_symbols.py
```diff
@@ -39,260 +37,16 @@
     stack,
 )
 
+# Try to import utils normally. If it fails, assume that we are running the
+# tests from the inside the `tests/` directory to avoid importing
+# dwave.optimization locally.
+try:
+    from .utils import symbol_testing_utils
```
arcondello, line 44: if we do
```python
    from .utils import symbol_testing_utils as utils
```
we won't need to change the other lines. Avoiding some merge conflicts

tests/utils.py
```diff
@@ -0,0 +1,263 @@
...
+
+import numpy as np
+
+import dwave.optimization
+from dwave.optimization import (
+    Model,
+)
+
+
+class symbol_testing_utils:
```
arcondello, line 13: This was to create a "namespace" in the other file. Now that it's in its own I think we can move everything up a level.

wbernoudy: I tried that originally, got the same error, and so thought it still didn't work for some reason so I left it in the "namespace". But I just tried to remove the namespace and now it works. :shrug:

arcondello (approved):
LGTM!

PR #234, opened 2025-02-19
Add Expit symbol

elham-azd:


arcondello:
Current CI error is because we don't install `scipy` in CI. Fix would be to add `scipy` to `tests/requirements.txt`

tests/test_symbols.py
```diff
@@ -1277,6 +1278,34 @@ def test_unsupported_floor_division(self):
             a // b
 
 
+class TestExpit(utils.SymbolTests):
+    rng = np.random.default_rng(1)
```
arcondello, line 1282: Looks like this isn't used elsewhere?

elham-azd: Right. Forgot to remove it. Thanks!

dwave/optimization/symbols.pyx
```diff
@@ -1670,6 +1672,41 @@ cdef class Equal(ArraySymbol):
...
+        variable to a model.
+
+        >>> from dwave.optimization.model import Model
+        >>> from dwave.optimization.mathematical import expit
+        >>> model = Model()
+        >>> i = model.constant(1.6)
+        >>> ii = expit(i)
+        >>> type(ii)
+        <class 'dwave.optimization.symbols.Expit'>
+    """
```
arcondello, line 1689:
```suggestion
    """Takes the values of a symbol and returns the corresponding logistic sigmoid (expit).

    See Also:
        :func:`~dwave.optimization.mathematical.expit`: equivalent function.
        
    .. versionadded:: 0.5.2
    """
```
What you had was consistent with other symbols in this file, so apologies for changing the pattern out from under you.

We want to discourage folks from using the symbol classes directly, so I think in general we just want to redirect users to the function/method that constructs it.

We also want to add the `..versionadded` when possible. The rule-of-thumb for that is to just increment the patch number. We'll go back an edit them if we end up changing the minor or major.

dwave/optimization/mathematical.py
```diff
@@ -199,6 +201,32 @@ def divide(x1: ArraySymbol, x2: ArraySymbol) -> Divide:
...
+        >>> x = model.constant(1.0)
+        >>> expit_x = expit(x)
+        >>> model.states.resize(1)
+        >>> with model.lock():
+        ...     print(expit_x.state())
+        0.7310585786300049
+
+    See Also:
+        :class:`~dwave.optimization.symbols.Expit`: equivalent symbol.
+    """
```
arcondello, line 227:
```suggestion
    See Also:
        :class:`~dwave.optimization.symbols.Expit`: equivalent symbol.

    .. versionadded:: 0.5.2
    """
```

arcondello (commented):
Two quick documentation changes, otherwise looks great to me!

dwave/optimization/symbols.pyx
elham-azd: That makes sense. Done. Thanks!

dwave/optimization/mathematical.py
elham-azd: Done. Thanks!

arcondello (approved):
LGTM!

Issue #223, opened 2025-02-05
Consider more compressable serialization formats


arcondello:
Models with a large number of scalar decision variables result in very large serialized files. We should consider format and/or serialization changes to help. Some of these are backwards compatibility breaking.

arcondello:
As a preliminary experiment, I tried a few simple approaches on a few pathological models. All experiments were run using the following script:
```python
import time

import numpy as np

from dwave.optimization import Model
from dwave.optimization.mathematical import add, multiply


def heavy(n: int, num_states: int = 100, seed: int = 42) -> Model:
    """Create a "heavy" model with n integer variables in a single symbol."""
    model = Model()

    x = model.integer(n, lower_bound=-1000, upper_bound=1000)

    model.minimize(-x.sum())
    model.add_constraint(x.prod() <= model.constant(1000))

    states = np.random.default_rng(seed).integers(-1000, 1000, size=(num_states, n))

    model.states.resize(num_states)
    for i, state in enumerate(states):
        x.set_state(i, state)

    return model


def tall(n: int, num_states: int = 100, seed: int = 42) -> Model:
    """Create a "tall" model with one scalar integer used in n operations."""
    model = Model()

    x = model.integer(lower_bound=-1000, upper_bound=1000)
    c = model.constant(0)

    out = x
    for _ in range(n):
        out = out + c

    model.minimize(-out)
    model.add_constraint(x <= c)

    states = np.random.default_rng(seed).integers(-1000, 1000, size=(num_states,))

    model.states.resize(num_states)
    for i, state in enumerate(states):
        x.set_state(i, state)

    return model


def wide(n: int, num_states: int = 100, seed: int = 42) -> Model:
    """Create a "wide" model with n integer variables each as a scalar."""
    model = Model()

    xs = [model.integer(lower_bound=-1000, upper_bound=1000) for _ in range(n)]

    model.minimize(-add(*xs))
    model.add_constraint(multiply(*xs) <= model.constant(1000))

    states = np.random.default_rng(seed).integers(-1000, 1000, size=(num_states, n))

    model.states.resize(num_states)
    for i, state in enumerate(states):
        for j, val in enumerate(state):
            xs[j].set_state(i, val)

    return model


# for each model, measure how long it takes to serialize the model and the states,
# and how large the serialized file is
for name, n, func in [("heavy", 100_000, heavy), ("tall", 100_000, tall), ("wide", 500, wide)]:
    print(f"**** generator: {name}({n})")
[...]

PR #235, opened 2025-02-21
Tidy serialization code and tests

arcondello:
The first of a series of serialization PRs to support v1. This one is mostly focused on tidying the existing code to make later PRs easier to review.

The only functional change is the addition of a `version` keyword argument for `to_file()`. I also did some test reorganization which will help the followup PRs.

I'll add a release note once the `version` keyword argument actually does something.

See also https://github.com/dwavesystems/dwave-optimization/issues/223

### DIFF
--- a/dwave/optimization/_model.pyx
+++ b/dwave/optimization/_model.pyx
@@ -39,6 +39,15 @@ from dwave.optimization.symbols cimport symbol_from_ptr
 __all__ = []
 
 
+DEFAULT_SERIALIZATION_VERSION = (0, 1)
+"""A 2-tuple encoding the default serialization format used for serializing models."""
+
+KNOWN_SERIALIZATION_VERSIONS = (
+    (0, 1),
+)
+"""A tuple of 2-tuples listing all serialization versions supported."""
+
+
 cdef class _Graph:
     """A ``_Graph`` is a class that manages a C++ ``dwave::optimization::Graph``.
 
@@ -156,11 +165,16 @@ cdef class _Graph:
 
         read_prefix = file.read(len(prefix))
         if read_prefix != prefix:
-            raise ValueError("unknown file type, expected magic string "
-                             f"{prefix!r} but got {read_prefix!r} "
+            raise ValueError("Unknown file type. Expected magic string "
+                             f"{prefix!r} but recieved {read_prefix!r} "
                              "instead")
 
         version = tuple(file.read(2))
+
+        if version not in KNOWN_SERIALIZATION_VERSIONS:
+            raise ValueError("Unknown serialization format. Expected one of "
+                             f"{KNOWN_SERIALIZATION_VERSIONS} but recieved {version} "
+                             "instead. Upgrading your dwave-optimization version may help.")
 
         if check_header:
             # we'll need the header values to check later
@@ -258,6 +272,7 @@ cdef class _Graph:
     def into_file(self, file, *,
                   Py_ssize_t max_num_states = 0,
                   bool only_decision = False,
+                  object version = None,
                   ):
         """Serialize the model into an existing file.
 
@@ -282,7 +297,12 @@ cdef class _Graph:
         if not self.is_locked():
             # lock for the duration of the method
             with self.lock():
-                return self.into_file(file, max_num_states=max_num_states, only_decision=only_decision)
+                return self.into_file(
+                    file,
+                    max_num_states=max_num_states,
+                    only_decision=only_decision,
+                    version=version,
+                    )
 
         if isinstance(file, str):
             with open(file, "wb") as f:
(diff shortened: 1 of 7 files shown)


tests/test_model.py
```diff
@@ -345,7 +346,87 @@ def test_remove_unused_symbols(self):
             self.assertEqual(num_removed, 1)
             self.assertEqual(model.num_symbols(), 9)
 
-    def test_serialization(self):
+    def test_state_size(self):
```
arcondello, line 349: Git got confused about which tests were changing. These are all unchanged, I simply moved the serialization tests to a different test class, and added one more test. So should be no functional change.

dwave/optimization/_model.pyx
```diff
@@ -156,12 +165,17 @@ cdef class _Graph:
 
         read_prefix = file.read(len(prefix))
         if read_prefix != prefix:
-            raise ValueError("unknown file type, expected magic string "
-                             f"{prefix!r} but got {read_prefix!r} "
+            raise ValueError("Unknown file type. Expected magic string "
+                             f"{prefix!r} but recieved {read_prefix!r} "
```
randomir, line 169: typo

dwave/optimization/_model.pyx
```diff
@@ -156,12 +165,17 @@ cdef class _Graph:
...
                              "instead")
 
         version = tuple(file.read(2))
 
+        if version not in KNOWN_SERIALIZATION_VERSIONS:
+            raise ValueError("Unknown serialization format. Expected one of "
+                             f"{KNOWN_SERIALIZATION_VERSIONS} but recieved {version} "
```
randomir, line 176: typo, here and elsewhere

dwave/optimization/states.pyx
```diff
@@ -200,7 +200,9 @@ cdef class States:
             self._states[i].resize(model.num_nodes())
             model._graph.initialize_state(self._states[i])
 
-    def into_file(self, file):
+    def into_file(self, file, *,
+                  version = None,
```
randomir, line 204: missing types?

tests/test_model.py
```diff
@@ -389,7 +470,28 @@ def test_serialization_by_filename(self):
         for n0, n1 in zip(model.iter_symbols(), new.iter_symbols()):
             self.assertIs(type(n0), type(n1))
 
-    def test_serialization_max_num_states(self):
+    def test_invalid_version(self):
+        from dwave.optimization._model import DEFAULT_SERIALIZATION_VERSION
```
randomir, line 474: perhaps expose this constant in a public module?

dwave/optimization/_model.pyx
```diff
@@ -258,6 +272,7 @@ cdef class _Graph:
     def into_file(self, file, *,
                   Py_ssize_t max_num_states = 0,
                   bool only_decision = False,
+                  object version = None,
```
randomir, line 275: docs missing below

tests/test_model.py
```diff
@@ -389,7 +470,28 @@ def test_serialization_by_filename(self):
...
+
+        model = Model()
+        with io.BytesIO() as f:
+            model.into_file(f)
```
randomir, line 478: want to test the path where you explicitly set `version` as well?

randomir (approved):
LGTM!

dwave/optimization/states.pyx
arcondello: The typing in Cython files doesn't get propagated, it goes in the `.pyi` file instead. Annoyingly.

tests/test_model.py
arcondello: I think we will in future PRs. But for this PR, because we only support one version anyway, I'd like to keep it private.

dwave/optimization/_model.pyx
arcondello: I added a short docstring. Once we expose `DEFAULT_SERIALIZATION_VERSION` I'll expand it to reference it.

tests/test_model.py
arcondello: Good call, and adding that test led me to another bug fix: https://github.com/arcondello/dwave-optimization/blob/2a4bca442b20200d0ee2937e683f217491b76b92/dwave/optimization/model.py#L418-L422

dwave/optimization/model.py
```diff
@@ -412,7 +412,15 @@ def to_file(self, **kwargs) -> typing.BinaryIO:
...
         file = tempfile.TemporaryFile(mode="w+b")
-        self.into_file(file, **kwargs)
+
+        # into_file can raise an exception, in which case we close off the
+        # tempfile before returning
+        try:
+            self.into_file(file, **kwargs)
+        except Exception as exception:
+            file.close()
+            raise exception
```
randomir, line 422: How about:
```suggestion
        try:
            self.into_file(file, **kwargs)
        finally:
            file.close()
```

randomir: actually, I see the file is used below.. so ignore this :)

randomir: but could at least simplify like;
```suggestion
        try:
            self.into_file(file, **kwargs)
        except:
            file.close()
            raise
```

arcondello: Ah! TIL. Though I feel like that syntax is a violation of the explicit is better than implicit.

arcondello: Also interesting: https://peps.python.org/pep-0760/

randomir: Yeah, there's an argument that can be made you shouldn't catch `BaseException`, so could do:
```
except Exception:
    raise
```

But I don't completely agree that bare except is implicit here. It's pretty explicit you want to close the file on _any_ error and re-raise it then. (The use case over which that PEP was rejected, I guess)

PR #241, opened 2025-02-25
Serialization v1.0

arcondello:
Note this PR does not increment the default serialization version. We'll do that in 0.7.0.

In a followup PR that I am already working on, I will add CI testing backwards and forwards in time checking that models serialized by package version 0.5.2 are loadable from 0.6.0+ and vice versa.

The docs need another pass or two for clarity, but better than the `# TODO` that was there before :slightly_smiling_face:

### DIFF
--- a/dwave/optimization/states.pyx
+++ b/dwave/optimization/states.pyx
@@ -168,6 +168,8 @@ cdef class States:
     def _from_zipfile(self, zf, *, version):
         """Given a zipfile containing a serialized model, attempt to extract the
         state(s) associated with any node types that might save them.
+
+        ``zf`` must be a :class:`zipfile.ZipFile`.
         """
         self.resolve()
 
@@ -195,7 +197,6 @@ cdef class States:
 
             symbol._states_from_zipfile(
                 zf,
-                directory=f"nodes/{symbol.topological_index()}/",
                 num_states=num_states,
                 version=version,
             )
@@ -229,6 +230,7 @@ cdef class States:
             self._states[i].resize(model.num_nodes())
             model._graph.initialize_state(self._states[i])
 
+    @_file_object_arg("wb")  # translate str/bytes file inputs into file objects
     def into_file(self, file, *,
                   version = None,
                   ):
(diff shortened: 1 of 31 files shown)


dwave/optimization/states.pyx
```diff
@@ -216,16 +247,112 @@ cdef class States:
...
+            Following the header, the remaining data is encoded as a zip file.
+            All arrays are saved using the NumPy serialization format, see
+            :func:`numpy.save()`.
+
+            The information in the header is also saved in a json-formatted
+            file ``info.json``.
+
+            The serialization version is saved in a file ``version.txt``.
+
+            The states have the following structure.
```
arcondello, line 264: This should note that only the states of decisions and other non-deterministic symbols is saved. I will fix.

randomir: the code below for v1.0 uses `only_decision=False`

tests/test_states.py
randomir, line 1: I would add serialization tests for model and states that verify loading/failing of mismatched version (0.1 in 1.0 and 1.0 in 0.1).

randomir (approved):
LGTM

dwave/optimization/states.pyx
```diff
@@ -216,16 +247,112 @@ cdef class States:
...
+        """
+        if num_states <= 0:
+            # nothing so save, so shortcut
+            return
+
+        model = self._model()  # get a ref-counted model
+
+        for symbol in model.iter_symbols():
+            if symbol._deterministic_state():
+                continue
```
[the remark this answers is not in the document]
arcondello: In many models saving all states would increase the state size by orders of magnitude. The way it's set up currently the state of intermediate variables is calculated lazily rather than eagerly.

tests/test_states.py
arcondello: Totally agree. That's in the next PR

dwave/optimization/states.pyx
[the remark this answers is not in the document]
randomir: Inspecting states of large models is a use case I was thinking about. Perhaps you don't want to construct the huge model just to check states if debugging?

dwave/optimization/states.pyx
arcondello: Yes, because it's using the `if symbol.deterministic_state(): continue` logic instead. Before we just hard-limited it to decisions which gave the same thing if you don't have non-determistic intermediate nodes (which we currently do not, but want to add).

PR #243, opened 2025-02-25
Move Model/States serialization tests into their own test file

arcondello:
This is to support the (in progress) https://github.com/dwavesystems/dwave-optimization/pull/241

No functional change, I haven't edited the tests at all. But moving them as part of #241 made it much harder to review.

### DIFF
--- a/tests/test_model.py
+++ b/tests/test_model.py
@@ -14,8 +14,6 @@
 
 import io
 import operator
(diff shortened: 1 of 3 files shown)


arcondello:
Actually, on second thought, I don't want to structure this way. Possible that I'll have a third thought.

PR #247, opened 2025-03-03
Add some utility decorators for Model/States methods

arcondello:
These are used in https://github.com/dwavesystems/dwave-optimization/pull/241, but are not really serialization-specific. So to make that PR easier to review IMO it makes sense to break them out.

The existing testing covers it pretty well, hence no new tests.

### DIFF
--- a/dwave/optimization/states.pyx
+++ b/dwave/optimization/states.pyx
@@ -19,6 +19,7 @@ from libcpp.utility cimport move
 from dwave.optimization.libcpp.array cimport Array as cppArray
 from dwave.optimization.model cimport ArraySymbol, _Graph
 from dwave.optimization.model import Model
+from dwave.optimization.utilities import _file_object_arg
 
 __all__ = ["States"]
 
@@ -137,6 +138,7 @@ cdef class States:
         self._states.swap(states)
         return move(states)
 
+    @_file_object_arg("rb")  # translate str/bytes file inputs into file objects
     def from_file(self, file, *, replace = True, check_header = True):
         """Construct states from the given file.
 
@@ -200,6 +202,7 @@ cdef class States:
             self._states[i].resize(model.num_nodes())
             model._graph.initialize_state(self._states[i])
 
+    @_file_object_arg("wb")  # translate str/bytes file inputs into file objects
     def into_file(self, file, *,
                   version = None,
                   ):
--- a/dwave/optimization/utilities.py
+++ b/dwave/optimization/utilities.py
@@ -1,0 +1,54 @@
+# Copyright 2025 D-Wave
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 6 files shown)


dwave/optimization/states.pyx
```diff
@@ -137,6 +138,7 @@ cdef class States:
         self._states.swap(states)
         return move(states)
 
+    @_file_object_arg("rb")  # translate str/bytes file inputs into file objects
```
arcondello, line 141: These are redundant for now, but will be used in https://github.com/dwavesystems/dwave-optimization/pull/241

dwave/optimization/utilities.py
```diff
@@ -0,0 +1,53 @@
...
+            if isinstance(file, str):
+                with open(file, mode) as fobj:
+                    return method(cls_or_self, fobj, *args, **kwargs)
+            else:
+                return method(cls_or_self, file, *args, **kwargs)
+        return _method
+    return decorator
+
+
+def _lock(method):
```
arcondello, line 45: Oops, needs a docstring. On it.

dwave/optimization/utilities.py
```diff
@@ -0,0 +1,54 @@
...
+    generalize if we need to.
+
+    Before adding this method we were frequently forgetting to pass all other
+    args/kwargs through which was a source of bugs.
+    """
+    def decorator(method):
+        @functools.wraps(method)
+        def _method(cls_or_self, file, *args, **kwargs):
+            if isinstance(file, str):
+                with open(file, mode) as fobj:
```
randomir, line 37: `open` accepts not only `str` for file, but also `bytes` and a `os.PathLike`(i.e. `pathlib.Path` instance). Neither of these two are a `str` instance.

dwave/optimization/utilities.py
```diff
@@ -0,0 +1,54 @@
...
+    This method assumes that the file argument is the first one. We could
+    generalize if we need to.
+
+    Before adding this method we were frequently forgetting to pass all other
+    args/kwargs through which was a source of bugs.
+    """
+    def decorator(method):
+        @functools.wraps(method)
+        def _method(cls_or_self, file, *args, **kwargs):
+            if isinstance(file, str):
```
randomir, line 36: I would consider inverting the test here to test for required `file` properties (like readable, writeable, binary), instead just `str` instance, as that would be a bit more flexible (to accept not only `bytes`, `Path`s, but also generators, etc).

Also, you could do ```_file_object_arg(read: bool = False, write: bool: False, binary: bool = True)``` for explicitness.

But I also appreciate the appeal of simplicity of the implemented approach.

dwave/optimization/utilities.py
```diff
@@ -0,0 +1,54 @@
...
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+import functools
+
+__all__ = []
+
+
+def _file_object_arg(mode):
```
randomir, line 20: Add type annotations here and below?

randomir (approved):
LGTM

dwave/optimization/utilities.py
arcondello: Good call, `os.PathLike` especially should be covered. Will fix and add tests.

dwave/optimization/utilities.py
arcondello: Yes, especially if we're also doing `os.PathLike` etc inverting makes sense.

Re the explicit input arguments, because this is private I am inclined to go with just `mode`. If we eventually make it public then we can consider more explicit arguments.

PR #248, opened 2025-03-03
Implement separate private methods for managing the header when (de)serializing

arcondello:
No functional change.

A reorganization that is used by https://github.com/dwavesystems/dwave-optimization/pull/241

### DIFF
--- a/dwave/optimization/_model.pyx
+++ b/dwave/optimization/_model.pyx
@@ -161,25 +161,7 @@ cdef class _Graph:
             with open(file, "rb") as f:
                 return cls.from_file(f)
 
-        prefix = b"DWNL"
-
-        read_prefix = file.read(len(prefix))
-        if read_prefix != prefix:
-            raise ValueError("Unknown file type. Expected magic string "
-                             f"{prefix!r} but received {read_prefix!r} "
-                             "instead")
-
-        version = tuple(file.read(2))
-
-        if version not in KNOWN_SERIALIZATION_VERSIONS:
-            raise ValueError("Unknown serialization format. Expected one of "
-                             f"{KNOWN_SERIALIZATION_VERSIONS} but received {version} "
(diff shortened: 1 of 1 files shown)


dwave/optimization/_model.pyx
```diff
@@ -251,6 +233,37 @@ cdef class _Graph:
...
+
+        Advances the stream position to the end of the header.
+
+        Returns:
+            The serialization version as a 2-tuple.
+
+            The header data as a dict.
+        """
+        prefix = b"DWNL"
+
```
randomir, line 248: I'd assert at the beginning the minimal length to be `len(prefix) + 2 + 4`, or fail with `ValueError`.

dwave/optimization/_model.pyx
```diff
@@ -251,6 +233,37 @@ cdef class _Graph:
...
+        version = tuple(file.read(2))
+
+        if version not in KNOWN_SERIALIZATION_VERSIONS:
+            raise ValueError("Unknown serialization format. Expected one of "
+                             f"{KNOWN_SERIALIZATION_VERSIONS} but received {version} "
+                             "instead. Upgrading your dwave-optimization version may help.")
+
+        header_len = struct.unpack('<I', file.read(4))[0]
+        header_data = json.loads(file.read(header_len).decode('ascii'))
```
randomir, line 263: Perhaps catch JSON decoding errors / insufficient data, wrapping them with some kind of `InvalidFileFormat` exception?

Rationale: this is a helper function that returns (version, data) from a file, so handling errors locally along the way would be nice.

randomir (approved):
LGTM

arcondello (commented):
In both cases the user will get an error, albeit an unhelpful one. But for now I am inclined to be lazy rather than eager with errors/checking. We haven't seen any users running into errors. I do think an error checking pass later for serialization on a whole makes sense.

PR #253, opened 2025-03-06
Add LPNodeBase ABC and LPSolutionNode

wbernoudy:
LPNode is no longer an ArrayNode, allowing the user to pick which of the outputs (solution, feasibility, objective) are needed in the model.

### DIFF
--- a/dwave/optimization/include/dwave-optimization/nodes/lp.hpp
+++ b/dwave/optimization/include/dwave-optimization/nodes/lp.hpp
@@ -22,13 +22,13 @@
 
 namespace dwave::optimization {
 
-class LPNode;
+class LPNodeBase;
 
 /// A logical node that propagates whether or not its predecessor LP is feasible.
 /// dev note: maybe call OptimalNode or SuccessNode? to match scipy
 class FeasibleNode : public ScalarOutputMixin<ArrayNode> {
  public:
-    explicit FeasibleNode(LPNode* lp_ptr);
+    explicit FeasibleNode(LPNodeBase* lp_ptr);
 
     /// @copydoc Array::buff()
     double const* buff(const State& state) const override;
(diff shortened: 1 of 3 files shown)


dwave/optimization/include/dwave-optimization/nodes/lp.hpp
```diff
@@ -56,7 +56,32 @@ class FeasibleNode : public ScalarOutputMixin<ArrayNode> {
...
+    static const double infinity();
+
+    /// Return the objective value. Undefined when the LP is not feasible.
+    virtual double objective_value(const State& state) const = 0;
+
+    virtual std::span<const double> solution(const State& state) const = 0;
+
+    virtual std::span<const ssize_t> variables_shape() const = 0;
+
+    virtual std::pair<double, double> _minmax() const = 0;
```
arcondello, line 84: Why the leading underscore?

dwave/optimization/src/nodes/lp.cpp
```diff
@@ -34,15 +34,18 @@ struct LPData {
...
+    explicit LPNodeData(std::vector<double>&& solution, bool is_feasible,
                         double objective_value) noexcept
-            : ArrayNodeStateData(std::move(values)),
+            : solution(std::move(solution)),
               is_feasible(is_feasible),
               old_is_feasible(is_feasible),
               objective_value(objective_value),
               old_objective_value(objective_value) {}
 
+    std::vector<double> solution;
```
arcondello, line 46: I think we do want the `LPSolutionNode` to be statefull in this case, and to hold the solution.

dwave/optimization/src/nodes/lp.cpp
```diff
@@ -181,17 +183,7 @@ LPNode::LPNode(ArrayNode* c_ptr, ArrayNode* b_lb_ptr, ArrayNode* A_ptr, ArrayNod
...
-void LPNode::commit(State& state) const { return data_ptr<LPNodeData>(state)->commit(); }
-
-const double LPNode::default_lower_bound() { return 0.0; }
-
-const double LPNode::default_upper_bound() { return LP_INFINITY; }
-
-std::span<const Update> LPNode::diff(const State& state) const {
-    return data_ptr<LPNodeData>(state)->diff();
-}
+void LPNode::commit(State& state) const {};  // return data_ptr<LPNodeData>(state)->commit(); }
```
arcondello, line 186: dead code?

dwave/optimization/include/dwave-optimization/nodes/lp.hpp
wbernoudy: I wanted to differentiate from the standard `minmax` but I'm realizing `variables_minmax()` is more appropriate.

dwave/optimization/src/nodes/lp.cpp
wbernoudy: Refactored to eliminate this extra state. Had to remove one check in the unit tests that relied on the `LPNode::feasible()` being up to date which it no longer is guaranteed to be.

arcondello (approved):

PR #255, opened 2025-03-06
LP symbols

wbernoudy:
Still need docstrings <s>and serialization</s>

### DIFF
--- a/dwave/optimization/include/dwave-optimization/nodes/lp.hpp
+++ b/dwave/optimization/include/dwave-optimization/nodes/lp.hpp
@@ -69,6 +69,9 @@ class LPNodeBase : public Node {
     /// Return whether the state is feasible for the LP model.
     virtual bool feasible(const State& state) const = 0;
 
+    /// Return the arguments mapped to predecessor indices
+    virtual std::unordered_map<std::string, ssize_t> get_arguments() const = 0;
+
     /// Any upper bound equal to or greater (or lower bound less to or equal)
     /// to this will be treated as unbounded.
     static const double infinity();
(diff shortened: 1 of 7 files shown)


arcondello (approved):
Seems reasonable so far

dwave/optimization/include/dwave-optimization/nodes/lp.hpp
```diff
@@ -69,6 +69,9 @@ class LPNodeBase : public Node {
     /// Return whether the state is feasible for the LP model.
     virtual bool feasible(const State& state) const = 0;
 
+    /// Return the arguments mapped to predecessor indices
+    virtual std::unordered_map<std::string, ssize_t> get_arguments() const = 0;
```
arcondello, line 73: Hmm, what about returning as a vector or span with `nullptr`s in order to match the input arguments?

wbernoudy: I did that originally. I settled on this because I think it adds a bit more safety to ensuring the arguments in C++ and in python are well matched. Otherwise we maintain two separate orderings of the arguments in cython and C++. I'm fine to go back to vector though.

arcondello: Nah, the distinction is very much bike shedding. I am happy with your current approach.

wbernoudy: Also if we return `Node*` or similar so we can return `nullptr` for empty arguments, how can the cython match up with the right predecessor during deserialization? I guess you would do the mapping to index in the cython serialization...

PR #258, opened 2025-03-11
Add C++ `SolveResultResult.set_solution()` method

arcondello:


dwave/optimization/src/simplex.hpp
```diff
@@ -17,6 +17,7 @@
 #include <cassert>
 #include <cmath>
 #include <span>
+#include <utility>
```
wbernoudy, line 20: Why is this needed?

arcondello: For `std::move()`

dwave/optimization/src/simplex.hpp
```diff
@@ -54,6 +55,13 @@ class SolveResult {
         partial_solution_set = true;
     }
 
+    // Set the solution explicitly. Throws an error
```
wbernoudy, line 58: when solution is an unexpected size?

arcondello: Oops, will fix.

wbernoudy (approved):

PR #259, opened 2025-03-12
Add LP state serialization

arcondello:


dwave/optimization/symbols.pyx
```diff
@@ -2225,6 +2285,56 @@ cdef class LP(Symbol):
...
+
+        # Test whether we have any states saved.
+        try:
+            states_info = zf.getinfo(f"nodes/{self.topological_index()}/states.npy")
+        except KeyError:
+            # No states, so nothing to load
+            return
+
+        # If we have states to load, then go ahead and do so.
+        # The ascontiguousarray is almost certainly redundant, but we use
```
arcondello, line 2306: Oops, this is a lie. Will remove

dwave/optimization/symbols.pyx
```diff
@@ -2200,6 +2208,58 @@ cdef class LP(Symbol):
             self._objective_value = LPObjectiveValue(self)
         return self._objective_value
 
+    def state(self, Py_ssize_t index = 0):
```
arcondello, line 2211: I am also missing some docstrings, will add

dwave/optimization/src/nodes/lp.cpp
```diff
@@ -242,17 +244,23 @@ void LPNode::readout_predecessor_data(const State& state, LPData& lp) const {
...
-    assert(static_cast<int>(state.size()) > index && "unexpected state length");
-    assert(state[index] == nullptr && "already initialized state");
-
     LPData lp;
     readout_predecessor_data(state, lp);
     SolveResult result = linprog(lp.c, lp.b_lb, lp.A, lp.b_ub, lp.A_eq, lp.b_eq, lp.lb, lp.ub,
                                  FEASIBILITY_TOLERANCE);
 
-    state[index] = std::make_unique<LPNodeData>(std::move(result));
+    emplace_data_ptr<LPNodeData>(state, std::move(result));
```
wbernoudy, line 252: Thanks, forgot about `emplace_data_ptr`

dwave/optimization/symbols.pyx
```diff
@@ -2200,6 +2208,42 @@ cdef class LP(Symbol):
...
+
+        if not self.model.is_locked() and self.node_ptr.topological_index() < 0:
+            raise TypeError("the state of an intermediate variable cannot be accessed without "
+                            "locking the model first. See model.lock().")
+
+        # Rather than using a StateView, let's just do an explicit copy here
+
+        cdef States states = self.model.states  # for Cython access
+        states.resolve()
+        self.model._graph.recursive_initialize(states._states.at(index), self.node_ptr)
```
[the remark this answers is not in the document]
arcondello: Yeah, state access needs a refactor. Probably a `cdef` method on the `Symbol` class.... but I am still thinking about it.

dwave/optimization/symbols.pyx
```diff
@@ -2200,6 +2208,42 @@ cdef class LP(Symbol):
...
+
+        solution = self.ptr.solution(states._states.at(index))
+
+        cdef double[::1] state = np.empty(self._num_columns(), dtype=np.double)
+
+        if <Py_ssize_t>solution.size() != state.shape[0]:
+            raise RuntimeError  # should never happen, but avoid the segfault just in case
+
+        for i in range(state.shape[0]):
```
wbernoudy, line 2242: Might be worth `cdef`ing `i`

arcondello: Cython (usually) does it automatically these days. And I did double check in this case.

dwave/optimization/symbols.pyx
```diff
@@ -2225,6 +2269,77 @@ cdef class LP(Symbol):
             encoder.encode({arg.decode(): val for arg, val in self.ptr.get_arguments()})
         )
 
+    cdef Py_ssize_t _num_columns(self) except -1:
+        """The number of columns in the LP."""
+        if self.ptr.variables_shape().size() != 1:
```
wbernoudy, line 2274: Just curious, why not use `assert` for these?

arcondello: `assert`s in Cython are treated as Python asserts. Which most people leave on all the time anyway. So I generally just do a full exception as a habit :shrug: I think either would be OK

dwave/optimization/symbols.pyx
```diff
@@ -2225,6 +2269,77 @@ cdef class LP(Symbol):
...
+        # We'll save our states into a dense array. And use NaN to signal when no state
+        # is present.
+        states = np.empty((num_states, self._num_columns()), dtype=np.double)
+        for state_index in range(num_states):
+            if self.has_state(state_index):
+                # we save the state regardless of whether it is feasible or not
+                # In the future we could choose to ignore infeasible states.
+                states[state_index, :] = self.state(state_index)
+            else:
+                raise NotImplementedError
```
wbernoudy, line 2336: Why not `states[state_index, :] = np.nan` or similar since deserializing already supports that? Or better just initialize the array as all nan?

arcondello: because I forgot to implement and more importantly to test this branch :laughing: Yes that's what it'll be

dwave/optimization/symbols.pyx
wbernoudy: I thought it might, but I couldn't remember so I looked up their docs and they are still adding a `cdef int i` in their examples!

dwave/optimization/symbols.pyx
wbernoudy: Yep, was just thinking about it as simply a stylistic thing, i.e. `assert` implies that it's for developers and the case should never happen

wbernoudy (approved):

PR #262, opened 2025-03-13
Add `linprog()` function

arcondello:
The docstring definitely needs even more, but this should suffice as an MVP.

### DIFF
--- a/dwave/optimization/mathematical.py
+++ b/dwave/optimization/mathematical.py
@@ -449,7 +449,7 @@ class LPResult:
 
     @functools.cached_property
     def success(self) -> LPFeasible:
-        """``True`` is the linear program found the optimal value as an array symbol."""
+        """``True`` if the linear program found the optimal value as an array symbol."""
         return LPFeasible(self.lp)
 
     @functools.cached_property
--- a/dwave/optimization/mathematical.py
+++ b/dwave/optimization/mathematical.py
@@ -26,6 +26,10 @@ from dwave.optimization.symbols import (
     Expit,
     Log,
     Logical,
+    LP,
+    LPFeasible,
+    LPObjectiveValue,
+    LPSolution,
     Maximum,
     Minimum,
     Modulus,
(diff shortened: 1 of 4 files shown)


dwave/optimization/mathematical.py
```diff
@@ -428,6 +433,131 @@ def hstack(arrays: collections.abc.Sequence[ArraySymbol]) -> ArraySymbol:
...
+        self.lp = lp
+
+    @functools.cached_property
+    def fun(self) -> LPObjectiveValue:
+        """The value of the objective as an array symbol."""
+        return LPObjectiveValue(self.lp)
+
+    @functools.cached_property
+    def success(self) -> LPFeasible:
+        """``True`` is the linear program found the optimal value as an array symbol."""
```
wbernoudy, line 452: if*

dwave/optimization/mathematical.py
```diff
@@ -428,6 +433,131 @@ def hstack(arrays: collections.abc.Sequence[ArraySymbol]) -> ArraySymbol:
...
+
+    Linear programs solve problems of the form:
+
+    .. math::
+        \text{minimize:} \\
+        & c^T x \\
+        \text{subject to:} \\
+        b_{lb} &\leq A x &\leq b_{ub} \\
+        b_{eq} &= A_{eq} x \\
+        l &\leq x &\leq u
```
wbernoudy, line 483: Perhaps we should use `lb` and `ub` here instead of `l` and `u`?

wbernoudy (approved):

dwave/optimization/mathematical.py
wbernoudy: And in the example below

arcondello: It messed up the sphinx math formatting unfortunately. So this is a compromise

dwave/optimization/mathematical.py
arcondello, line 452:
```suggestion
        """``True`` if the linear program found the optimal value as an array symbol."""
```

PR #263, opened 2025-03-14
Rename LP -> LinearProgram and LPNode -> LinearProgramNode

arcondello:
No functional changes, but hopefully a clearer (and final) name

### DIFF
--- a/dwave/optimization/symbols.pyx
+++ b/dwave/optimization/symbols.pyx
@@ -68,14 +68,14 @@ from dwave.optimization.libcpp.nodes cimport (
     ExpitNode as cppExpitNode,
     IntegerNode as cppIntegerNode,
     LessEqualNode as cppLessEqualNode,
+    LinearProgramFeasibleNode as cppLinearProgramFeasibleNode,
+    LinearProgramNode as cppLinearProgramNode,
+    LinearProgramNodeBase as cppLinearProgramNodeBase,
+    LinearProgramObjectiveValueNode as cppLinearProgramObjectiveValueNode,
+    LinearProgramSolutionNode as cppLinearProgramSolutionNode,
     ListNode as cppListNode,
     LogNode as cppLogNode,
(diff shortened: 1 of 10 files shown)


dwave/optimization/symbols.pyx
```diff
@@ -2329,12 +2189,12 @@ cdef class LP(Symbol):
...
-    cdef cppLPNode* ptr
+    cdef cppLinearProgramNode* ptr
 
-_register(LP, typeid(cppLPNode))
+_register(LinearProgram, typeid(cppLinearProgramNode))
 
 
-cdef class LPFeasible(ArraySymbol):
+cdef class LinearProgramFeasible(ArraySymbol):
     """Return whether the parent LP symbol's current solution is feasible.
```
wbernoudy, line 2198: Might want to use `LinearProgram` in these docstrings?

dwave/optimization/symbols.pyx
```diff
@@ -2013,147 +2013,7 @@ cdef class LessEqual(ArraySymbol):
...
-        x.initialize_arraynode(symbol.model, ptr)
-        return x
-
-    cdef cppLogicalNode* ptr
-
-_register(Logical, typeid(cppLogicalNode))
-
-
-cdef class LP(Symbol):
+cdef class LinearProgram(Symbol):
```
wbernoudy, line 2016: Diff didn't do so well here, huh

dwave/optimization/symbols.pyx
arcondello: Chose the more succinct wording for now. "LP symbol" is descriptive, if we were doing
```
:class:`LinearProgram`
```
I would be more explicit

dwave/optimization/symbols.pyx
arcondello: Had to move it for alphabetical

wbernoudy (approved):

PR #266, opened 2025-03-25
Add typed/broadcasting iterators

arcondello:
Replace the current iterator with a typed/broadcasting iterator. No nodes take advantage of that functionality yet though.


dwave/optimization/include/dwave-optimization/array.hpp
```diff
@@ -38,6 +38,515 @@ namespace dwave::optimization {
...
+concept DType = std::same_as<T, float> ||  // np.float32
+        std::same_as<T, double> ||         // np.float64
+        std::same_as<T, std::int8_t> ||    // np.int8
+        std::same_as<T, std::int16_t> ||   // np.int16
+        std::same_as<T, std::int32_t> ||   // np.int32
+        std::same_as<T, std::int64_t>;     // np.int64
+
+/// The `OptionalDType<T>` concept is satisfied if and only if `T` is `void`
+/// or satisfies `DType<T>`.
+/// `void` is used to signal that the dtype is unknown at compile-time.
```
wbernoudy, line 56: Would `std::monostate` be more appropriate here?

arcondello: I think `monostate` is doing more than we want here.
```
Unit type intended for use as a well-behaved empty alternative in [std::variant](https://en.cppreference.com/w/cpp/utility/variant). In particular, a variant of non-default-constructible types may list std::monostate as its first alternative: this makes the variant itself default-constructible.
```
https://en.cppreference.com/w/cpp/utility/variant/monostate

I don't think that's what we're trying to signal. It's not empty, it's undefined.

I am using void following from say [std::multiplies](https://en.cppreference.com/w/cpp/utility/functional/multiplies)
```
The standard library provides a specialization of std::multiplies when T is not specified, which leaves the parameter types and return type to be deduced. 
```

wbernoudy (approved):
Nicely done!

PR #270, opened 2025-04-15
Add simple propagation overloads

arcondello:
I have not touched the existing methods, though probably at some point we should switch them over to using `std::ranges::for_each(...)` as well.


wbernoudy (approved):
Only thing I can think of is that it might be nice to have a note on these methods or on `Graph` that they may be significantly slower than the descendants versions.

arcondello:
> Only thing I can think of is that it might be nice to have a note on these methods or on Graph that they may be significantly slower than the descendants versions.

I am on the fence. I think if this was the Python level I would do it. But at the C++ level I think "all nodes in the graph" should tip our hand pretty well.

PR #271, opened 2025-04-21
Add default state support to `ScalarOutputMixin`

arcondello:
Removes a fair amount of code duplication.

Could be replicated for `ArrayOutputMixin`, but IMO we should play with this one for a while first.

One thing that this is highlighting for me is that we have a circular dependency between `array.hpp` and `graph.hpp`. We've gotten around it for now by messing with templates and the like, but probably there are nicer things we could do in the future.


wbernoudy (approved):
LGTM!

PR #274, opened 2025-04-25
Include <cstdint> on all files that use std::int*_t

wbernoudy:
Otherwise compilation fails on some compilers that seem not to include cstdint implicitly


arcondello (approved):
Thanks!

PR #275, opened 2025-04-29
Improve state setting for List and Set variables

arcondello:


dwave/optimization/src/nodes/collections.cpp
```diff
@@ -11,36 +11,115 @@
...
+        throw std::invalid_argument("values values must be non-negative");
+    }
+    if (std::ranges::any_of(values, [&n](const auto& val) -> bool { return val >= n; })) {
+        throw std::invalid_argument("values values must be less than " + std::to_string(n));
+    }
+
+    // Next check that we have a sub-permutation of range(n)
+
+    // We'll track how many times each value appears in the values.
+    // We void the weird vector<bool> overload for performance even though
```
arcondello, line 43: I meant "avoid" but tempted to keep this...

samuelbodin (approved):

PR #277, opened 2025-04-30
Allow shape arg for Model.input()

wbernoudy:
Fix docstring, move the method to python Model class, and add input validation when initializing InputNode.


dwave/optimization/_model.pyx
```diff
@@ -140,58 +140,6 @@ cdef class _Graph:
         """
         return sum(sym.state_size() for sym in self.iter_decisions())
 
-    def input(
```
arcondello, line 143: Also needs to be removed from `_model.pyi` I believe

dwave/optimization/model.py
```diff
@@ -257,6 +257,62 @@ def feasible(self, index: int = 0) -> bool:
         """
         return all(sym.state(index) for sym in self.iter_constraints())
 
+    def input(
+        self,
+        shape: tuple[int] = tuple(),
```
arcondello, line 262:
```python
        shape: tuple[int, ...] = tuple(),
```
what you have is only satisfied by a tuple of length 1

arcondello (approved):

PR #279, opened 2025-05-01
Move the utility functions out of symbols.pyx

arcondello:


dwave/optimization/symbols.pyx
```diff
@@ -583,6 +511,10 @@ cdef class ARange(ArraySymbol):
 _register(ARange, typeid(cppARangeNode))
 
 
+cdef bool _empty_slice(object slice_) noexcept:
```
wbernoudy, line 514: Why didn't `_empty_slice()` make it out? And moved here?

arcondello: It is only used by that one symbol. Ultimately this PR is on the way to breaking the `symbols.pyx` into multiple files. So the methods that would be shared got moved out but the ones that wouldn't were kept "local".

wbernoudy (approved):

PR #288, opened 2025-05-15
Remove symbol construction boilerplate from many symbols

arcondello:
Many symbols don't actually need to hold a pointer of their specific type. We can remove a lot of boilerplate from such symbols.

### DIFF
--- a/dwave/optimization/symbols.pyx
+++ b/dwave/optimization/symbols.pyx
@@ -839,7 +839,7 @@ cdef class BSpline(ArraySymbol):
     def _from_symbol(cls, Symbol symbol):
         cdef cppBSplineNode * ptr = dynamic_cast_ptr[cppBSplineNode](symbol.node_ptr)
         if not ptr:
-            raise TypeError("given symbol cannot be used to construct a BSpline")
+            raise TypeError(f"given symbol cannot construct a {cls.__name__}")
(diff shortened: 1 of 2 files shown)


dwave/optimization/symbols.pyx
```diff
@@ -762,11 +686,11 @@ cdef class BinaryVariable(ArraySymbol):
...
         self.initialize_arraynode(model, self.ptr)
 
-    @staticmethod
-    def _from_symbol(Symbol symbol):
+    @classmethod
+    def _from_symbol(cls, Symbol symbol):
         cdef cppBinaryNode* ptr = dynamic_cast_ptr[cppBinaryNode](symbol.node_ptr)
         if not ptr:
-            raise TypeError("given symbol cannot be used to construct a BinaryVariable")
+            raise TypeError(f"given symbol cannot construct a {cls.__name__}")
```
alexzucca90, line 693: I see this change in the error string happens only here and not consistently through the file. You left the original error string elsewhere

arcondello: You're right. It's an internal error so I doubt it matters much. But sure, I can change it to be consistent elsewhere.

alexzucca90 (approved):

alexzucca90 (approved):

Issue #287, opened 2025-05-15
iterating over an ArraySymbol yields ValueError


p-ortmann:
Iterating over an ArraySymbol always yields a value error at the end of the loop, if an IndexError was raised instead the loop would finish normally.

See the following reproducer below:
```python 
from dwave.optimization.model import Model
from dwave.optimization.symbols import BinaryVariable, BasicIndexing

model: Model = Model()
y: BinaryVariable = model.binary(shape = 3)

for i in y:
    i: BasicIndexing
    i.id
```
dwave.optimization version 0.6.0

PR #289, opened 2025-05-15
Fix iteration over `ArraySymbols`

arcondello:
Closes https://github.com/dwavesystems/dwave-optimization/issues/287


dwave/optimization/_model.pyi
```diff
@@ -101,6 +101,7 @@ class ArraySymbol(Symbol):
 
     def __iadd__(self, rhs: ArraySymbol) -> NaryAdd: ...
     def __imul__(self, rhs: ArraySymbol) -> NaryMultiply: ...
+    def __iter__(self) -> typing.Iterator[ArraySymbol]: ...
```
arcondello, line 104: In practice this will always be `BasicIndexing` for now. But let's leave the door open for the future.

p-ortmann (approved):
LGTM - The reproducer I sent is passing on this branch, thanks!

Issue #290, opened 2025-05-16
Consider letting the user provide a value for dividing by zero


arcondello:
Currently the [divide(x1, x2)](https://github.com/dwavesystems/dwave-optimization/blob/60e139d456e33df448b70170817d0ca3309c0263/dwave/optimization/mathematical.py#L364) operation requires that `x2` be either strictly positive or strictly negative. This allows us to safely avoid division by zero. 

One thing we could do is allow the user to specify a finite value to propagate in the case of division by zero. Something like
```python
z = divide(x, y, infinity=0)
```
this would allow us to relax that requirement. Though it shifts the burden of "correctness" on the user.

arcondello:
I think https://xenaproject.wordpress.com/2020/07/05/division-by-zero-in-type-theory-a-faq/ makes a pretty compelling case for defining the result of `1 / 0 = 0`. I expect that the behavior will be unexpected, but it leads me to think we should instead provide a boolean flag to the `divide(...)` function that toggles this behavior. Now just need to figure out what to call it 🤔

arcondello:
I think for now, the best thing is to actually introduce a new `safe_divide()` function that has the `1/0 = 0` behavior. We can consider changing the behavior of  `a / b` later, or adding some ability to toggle it.

PR #291, opened 2025-05-16
Update mathematical nodes to hold their op

arcondello:
Rather than merely tracking the type, they now hold an instantiated functional object. This allows for statefull operations in the future.

This is required for say #290


alexzucca90 (approved):
lgtm

PR #292, opened 2025-05-20
Add `SafeDivide` symbol

arcondello:
Closes https://github.com/dwavesystems/dwave-optimization/issues/290

### DIFF
--- a/dwave/optimization/src/nodes/mathematical.cpp
+++ b/dwave/optimization/src/nodes/mathematical.cpp
@@ -270,6 +270,13 @@ std::pair<double, double> BinaryOpNode<BinaryOp>::minmax(
             // rhs's range includes zero, but it's not currently accounted for
             // so let's do that.
             combos.emplace_back(op(1, 0));
+
+            if (!rhs_ptr->integral()) {
+                combos.emplace_back(std::copysign(std::numeric_limits<double>::max(), lhs_low));
+                combos.emplace_back(std::copysign(std::numeric_limits<double>::max(), -lhs_low));
+                combos.emplace_back(std::copysign(std::numeric_limits<double>::max(), lhs_high));
+                combos.emplace_back(std::copysign(std::numeric_limits<double>::max(), -lhs_high));
+            }
         } else if (rhs_low == 0 && rhs_high != 0) {
             if (rhs_ptr->integral()) {
                 assert(rhs_high >= rhs_low + 1);  // the bounds should leave room
--- a/dwave/optimization/src/nodes/mathematical.cpp
+++ b/dwave/optimization/src/nodes/mathematical.cpp
@@ -203,7 +203,8 @@ bool BinaryOpNode<BinaryOp>::integral() const {
     auto lhs_ptr = operands_[0];
     auto rhs_ptr = operands_[1];
 
-    if constexpr (std::is_same<BinaryOp, std::divides<double>>::value) {
+    if constexpr (std::is_same<BinaryOp, std::divides<double>>::value ||
+                  std::is_same<BinaryOp, functional::safe_divides<double>>::value) {
         return false;
     }
     if constexpr (std::is_same<BinaryOp, functional::max<double>>::value ||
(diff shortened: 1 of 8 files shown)


dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -259,6 +260,40 @@ std::pair<double, double> BinaryOpNode<BinaryOp>::minmax(
...
+    if constexpr (std::same_as<BinaryOp, functional::safe_divides<double>>) {
+        // safe_divide is, well, safe. So we start by calculating all combos.
+        // Though there are some possible other values depending on our rhs.
+        std::vector<double> combos = {op(lhs_low, rhs_low), op(lhs_low, rhs_high),
+                                      op(lhs_high, rhs_low), op(lhs_high, rhs_high)};
+
+        if (rhs_low < 0 && 0 < rhs_high) {
+            // rhs's range includes zero, but it's not currently accounted for
+            // so let's do that.
+            combos.emplace_back(op(1, 0));
```
smharwood, line 272: Is this to future-proof in case the "safe" value (that is returned when the rhs is zero) changes?

arcondello: yes, probably it's hard-coded elsewhere but seemed like an easy bit of future proofing.

smharwood: Shouldn't this case (`rhs_low < 0 && 0 < rhs_high`) also add the `std::numeric_limits<double>::max()` values?

arcondello: ah, good point. That's likely also a bug with the existing `DivideNode`.

arcondello: Nevermind, not an issue for `DivideNode` although I added some tests to be sure. Fixed by [59c8ee4](https://github.com/dwavesystems/dwave-optimization/pull/292/commits/59c8ee4fa2730a13121adfebbb98b895e927ee16)

dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -259,6 +260,45 @@ std::pair<double, double> BinaryOpNode<BinaryOp>::minmax(
...
+            if (rhs_ptr->integral()) {
+                assert(rhs_low + 1 <= rhs_high);
+                combos.emplace_back(op(lhs_low, rhs_high - 1));
+                combos.emplace_back(op(lhs_high, rhs_high - 1));
+            } else {
+                // We can get very close to dividing by -0
+                combos.emplace_back(std::copysign(std::numeric_limits<double>::max(), -lhs_low));
+                combos.emplace_back(std::copysign(std::numeric_limits<double>::max(), -lhs_high));
+            }
+        }
```
smharwood, line 298: Argh, so many edge cases. Here's the organization of the logic that makes the most sense to me:
```suggestion
       if (rhs_ptr->integral()) {
            if (rhs_low < 0 && 0 < rhs_high) {
                combos.emplace_back(op(1, 0));
            }
            if (rhs_low < -1 && -1 < rhs_high) {
                combos.emplace_back(op(lhs_low, -1));
                combos.emplace_back(op(lhs_high, -1));
            }
            if (rhs_low < 1 && 1 < rhs_high) {
                combos.emplace_back(op(lhs_low, 1));
                combos.emplace_back(op(lhs_high, 1));
            }
        } else {
            if (rhs_low < 0 && 0 < rhs_high) {
                // rhs's range includes zero, but it's not currently accounted for
                // so let's do that.
                // combos.emplace_back(op(1, 0)); // not necessary b/c max and lowest cover any value this could be
                combos.emplace_back(std::numeric_limits<double>::max());
                combos.emplace_back(std::numeric_limits<double>::lowest());
            } else if (rhs_low == 0 && rhs_high != 0) {
                // We can get very close to dividing by 0
                combos.emplace_back(std::copysign(std::numeric_limits<double>::max(), lhs_low));
                combos.emplace_back(std::copysign(std::numeric_limits<double>::max(), lhs_high));
            } else if (rhs_low != 0 && rhs_high == 0) {
                // We can get very close to dividing by -0
                combos.emplace_back(std::copysign(std::numeric_limits<double>::max(), -lhs_low));
                combos.emplace_back(std::copysign(std::numeric_limits<double>::max(), -lhs_high));
            }
        }
```

arcondello: Github needs a :headdesk: emoji. You're right, something like this might be easier to follow. Let me rework the logic.

smharwood (approved):
LGTM!

PR #293, opened 2025-05-21
Generalize C++ functional objects

arcondello:


dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
```diff
@@ -81,16 +72,17 @@ struct min {
...
+struct modulus {
+    constexpr T operator()(const T& x, const T& y) const {
         // Copy numpy behavior and return 0 for `x % 0`
-        if (y == 0) {
-            return 0;
+        if (y == 0) return 0;
+
+        T result;
+        if constexpr (std::integral<T>) {
+            result = x % y;
```
wbernoudy, line 82: I think `%` has implementation-dependent behavior if either operand is negative, perhaps worth using `std::div` here as it should consistent behavior with `std::fmod`?

arcondello: I think that's handled just below?

arcondello: I suppose I can/should add some tests for consistency with NumPy anyway

wbernoudy: I'm not sure there's any version of any compiler we support for which this is true, but some older compilers may return a positive value when `x` is negative but `y` is positive for example (which I think wouldn't work with the logic below?). Using `std::div` simply has a guarantee on that behavior. Definitely a nit

wbernoudy (approved):

PR #298, opened 2025-05-23
Make default Input limits finite and consistent

wbernoudy:


dwave/optimization/src/nodes/inputs.cpp
```diff
@@ -21,6 +21,30 @@
 
 namespace dwave::optimization {
 
+void check_limits(double min, double max) {
```
arcondello, line 24: Not that it hurts, by why break this out? Looks like it's only used in one place.

arcondello (approved):
LGTM

dwave/optimization/src/nodes/inputs.cpp
wbernoudy: I originally was using that in two constructors but then removed one. I'll move it back!

wbernoudy: Done

PR #302, opened 2025-05-28
Switch from pkgutil to native namespace package

randomir:


arcondello:
Interesting. Cython claims to support namespace packages as of Cython 3. See https://github.com/cython/cython/issues/2918. Looking...

PR #303, opened 2025-05-29
Support index-wise bounds for NumberNode

wbernoudy:


arcondello:
Just to save this somewhere, you can currently get something like the same API with

```python
import numbers

import dwave.optimization
import numpy as np
import numpy.typing


def integers(
	model: dwave.optimization.Model,
	shape: int | tuple[int, ...],
	*,
	lower_bound: None | numpy.typing.ArrayLike = None,
	upper_bound: None | numpy.typing.ArrayLike = None,
) -> dwave.optimization.model.ArraySymbol:
	# Handle the case that they're both scalar
	if ((lower_bound is None or isinstance(lower_bound, numbers.Number)) and
		(upper_bound is None or isinstance(upper_bound, numbers.Number))):
		return model.integer(shape, lower_bound=lower_bound, upper_bound=upper_bound)

	# First make sure everything is our desired shape
	# Works just fine with `None` so that's nice!
	lower_bound = np.broadcast_to(lower_bound, shape)
	upper_bound = np.broadcast_to(upper_bound, shape)

	# Now construct a scalar IntegerVariable symbol for each ub/lb
	symbols = []
	for lb, ub in zip(lower_bound.flatten(), upper_bound.flatten()):
		symbols.append(model.integer(lower_bound=lb, upper_bound=ub))

	# Finally put them into a 1D array and then reshape it to our desired shape
	# Also, we should update dwave-optimization to return self when reshape matches
	# the current shape, but that's a future problem and probably doesn't matter
	# for performance.
	return dwave.optimization.hstack(symbols).reshape(shape)
```

but that's a workaround

PR #307, opened 2025-06-06
Support two dynamic predecessors for BinaryOp

wbernoudy:
This is not an optimal implementation for handling the dynamic case with both parents having updates, but I tried to strike a balance between efficiency and keeping the changes minimal (by using `deduplicate_diff()`, even though it preserves the update values we don't care about).

### DIFF
--- a/dwave/optimization/src/nodes/mathematical.cpp
+++ b/dwave/optimization/src/nodes/mathematical.cpp
@@ -111,9 +111,7 @@ BinaryOpNode<BinaryOp>::BinaryOpNode(ArrayNode* a_ptr, ArrayNode* b_ptr)
     // Otherwise both arrays must be the same shape and not be dynamic
     if (lhs_ptr->size() == 1 || rhs_ptr->size() == 1) {
         // this is allowed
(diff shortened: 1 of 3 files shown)


dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -328,40 +326,58 @@ void BinaryOpNode<BinaryOp>::propagate(State& state) const {
...
-                    changes.emplace_back(index, old, values[index]);
+            auto apply_op = [this, &lit, &rit](const Update& up){
+                if (up.removed()) {
+                    return up;
                 }
-            }
-        } else if (lhs_ptr->diff(state).size()) {
+                // Use the update's old value in case it's a placment
+                // ArrayNodeStateData.update() otherwise ignores the old value
+                return Update(up.index, up.old, op(*(lit + up.index), *(rit + up.index)));;
```
arcondello, line 360: extra semicolon

arcondello (approved):
Surprisingly straightforward! Nice work!

PR #308, opened 2025-06-09
Fix AdvancedIndexing::sizeinfo()

wbernoudy:
Previously `AdvancedIndexing::sizeinfo()` in the dynamic case was returning a new `SizeInfo()` derived only from the first indexer. In the case where there are no empty slices, this is technically correct, but needlessly drops the information returned by that predecessors `sizeinfo()`. In the case where there are empty slices, this is incorrect.

This should be fixed for both cases by using the predecessors `sizeinfo()` and multiplying by the remaining dimensions in the shape of the advanced indexing node.


arcondello (approved):
Nice!

PR #309, opened 2025-06-10
Implement a proof-of-concept of an expression() function

arcondello:
DO NOT MERGE


dwave/optimization/expression.py
```diff
@@ -0,0 +1,55 @@
...
+
+    @staticmethod
+    def _from_function(function: typing.Callable) -> Expression:
+        expr = Expression.__new__(Expression)
+        expr._function = function
+        expr._model = model = Model()
+
+        # Use function to construct the symbols. Let the relevant methods to raise
+        # any relevant exceptions.
+        inputs = [model.input() for _ in inspect.signature(function).parameters]
```
wbernoudy, line 45: This currently will make almost all expressions unusable with NaryReduce as the default bounds for `model.input()` are two restrictive. And more generally, I think we do need to allow users to set the bounds on the inputs of an expression.

wbernoudy: I like the latter, that should work fine.

arcondello: There's also
```python
@expression
def func(a: Input[-1, 1, False], b: Input[None, 100]): ...
```
not mutually exclusive :thinking:

arcondello:
Closing in favour of https://github.com/dwavesystems/dwave-optimization/pull/312 and a PR that I'll make to https://github.com/dwavesystems/dwave-optimization/pull/166

PR #312, opened 2025-06-13
Implement a few QOL features for Input symbol

arcondello:


dwave/optimization/include/dwave-optimization/nodes/inputs.hpp
```diff
@@ -61,7 +61,7 @@ class InputNode : public ArrayOutputMixin<ArrayNode> {
 
     [[noreturn]] void initialize_state(State& state) const override {
         throw std::logic_error(
-                "InputNode must have state explicity initialized (with `initialize_state(state, "
+                "InputNode must have state explicitly initialized (with `initialize_state(state, "
```
arcondello, line 64: This is also raised to the Python level so maybe needs a different error message, but I chose to leave it alone for now (other than fixing the spelling).

wbernoudy (approved):

PR #317, opened 2025-06-13
Add Extract symbol

### LINKED ISSUE #272: Consider Extract symbol or similar
We would like to be able to do constructions like
```python
x = np.asarray([0, 1, 2, 3, 2])
A = np.asarray([10, 20, 30, 40, 50])

y = A[x == 2]  # y = [30, 50]
```
but we don't current allow boolean indexing, and likely won't until we allow typed arrays (see https://github.com/dwavesystems/dwave-optimization/pull/266 for the first step in that project).

NumPy does have an explicit function though, [numpy.extract()](https://numpy.org/doc/2.2/reference/generated/numpy.extract.html). So we should consider making an equivalent symbol.

### Workaround

For now, it's possible to get close with some fiddling.
```python
from dwave.optimization import Model, where

model = Model()
model.states.resize(1)

x = model.integer(5, upper_bound=5)
x.set_state(0, [0, 1, 2, 3, 2])

A = model.constant([10, 20, 30, 40, 50])
two = model.constant(2)

# Before making y, we need some intermediate symbols

# The `indices` array will be used to get the indices where x == 2
indices = model.constant(range(5))

# `neq` will be the value where x != 2
neq = model.constant([5] * 5)

# We add an extra 0 to A, to give us an "out of bounds" value
A_augmented = model.constant([10, 20, 30, 40, 50, 0])

# Now we can make something like y
y = A_augmented[where(x == two, indices, neq)]

with model.lock():
    print(y.state())
```
which outputs
```
[ 0.  0. 30.  0. 50.]
```

wbernoudy:
Closes #272


releasenotes/notes/add-extract-symbol-126b39be743c8c66.yaml
```diff
@@ -0,0 +1,5 @@
+---
+features:
+  - |
+    Add the ``Extract`` symbol and node, following the behavior of
+    ``numpy.extract``.
```
arcondello, line 5:
```
    ``numpy.extract``. See `#272 <https://github.com/dwavesystems/dwave-optimization/issues/272>`_.
```

dwave/optimization/src/nodes/_state.hpp
```diff
@@ -39,38 +39,39 @@ class ArrayStateData {
...
 
-    // Assign new values to the state, tracking the changes from the previous state to the new
-    // one. Including resizes.
-    bool assign(std::ranges::sized_range auto&& values) {
-        // dev note: we could implement a version of this that doesn't need sized_range.
-        const ssize_t overlap_length = std::min<ssize_t>(buffer.size(), std::ranges::size(values));
+    // Assign new values to the state, starting from an offset and optionally "trimming" the
+    // end of the previous buffer if it extends past the range of the newly set values.
+    bool assign_partial(std::ranges::sized_range auto&& values, ssize_t offset,
+                        bool trim_end = false) {
```
arcondello, line 45: As far as I can tell, we never use `trim_end = False`. Is this legacy for an old implementation or?

dwave/optimization/src/nodes/flow.cpp
```diff
@@ -138,6 +245,7 @@ void WhereNode::commit(State& state) const { data_ptr<WhereNodeData>(state)->com
 
 std::span<const Update> WhereNode::diff(const State& state) const {
     return data_ptr<WhereNodeData>(state)->diff();
+    this->sizeinfo();
```
arcondello, line 248: debugging code?

dwave/optimization/src/nodes/flow.cpp
```diff
@@ -42,6 +42,113 @@ std::span<const ssize_t> same_shape(const Array* node_ptr,
     return node_ptr->shape();
 }
 
+/// ExtractNode
+
+ExtractNode::ExtractNode(ArrayNode* condition_ptr, ArrayNode* arr_ptr)
+        : ArrayOutputMixin({-1}), condition_ptr_(condition_ptr), arr_ptr_(arr_ptr) {
+    if (condition_ptr_->sizeinfo() != arr_ptr_->sizeinfo()) {
```
arcondello, line 49: I think you want `same_shape(condition_ptr_, arr_ptr_)`, which does the recursive check.

dwave/optimization/src/nodes/flow.cpp
```diff
@@ -42,6 +42,113 @@ std::span<const ssize_t> same_shape(const Array* node_ptr,
...
+std::pair<double, double> ExtractNode::minmax(
+        optional_cache_type<std::pair<double, double>> cache) const {
+    return arr_ptr_->minmax(cache);
+}
+
+void ExtractNode::propagate(State& state) const {
+    auto node_data = data_ptr<ArrayNodeStateData>(state);
+
+    // Nothing to do in this case
+    if (condition_ptr_->diff(state).size() == 0 && arr_ptr_->diff(state).size() == 0) return;
```
arcondello, line 95:
```
    if (condition_ptr_->diff(state).empty() && arr_ptr_->diff(state).empty()) return;
```
slightly more succinct

dwave/optimization/src/nodes/_state.hpp
wbernoudy: I had it in mind that this could be used by a few more nodes where we want to assign subrange of the buffer. But now I can't find any more examples so I can just remove the parameter for now.

arcondello: In that case I am tempted to keep `bool assign(..., ssize_t offset = 0)` rather than adding an overload. But could go either way

dwave/optimization/src/nodes/flow.cpp
wbernoudy: `same_shape()` actually checks only the non-dynamic part of the shape. And what we care about here is same size, as different shapes are explicitly allowed. But is there some other check I should be doing?

arcondello: You're right, the suggestion was wrong.

BUt I think this check can fail in some cases where it should succeed if we don't do something like
```c++
if (condition_ptr_->sizeinfo().substitute(100) != arr_ptr_->sizeinfo().substitute(100)) {
```

I thought we had a utility function but I cannot find it :dizzy_face:

PR #320, opened 2025-06-20
Support all dynamic predecessors in WhereNode

wbernoudy:


wbernoudy:
Don't worry, I'll add a release note soon

dwave/optimization/src/nodes/_state.hpp
```diff
@@ -173,6 +176,20 @@ class ArrayStateData {
         return size_ - previous_size_;
     }
 
+    bool trim_to(ssize_t new_size) {
```
arcondello, line 179: I've been trying to emulate C++ container methods for the most part. So I'd rather add a more generic `resize()` method here.

dwave/optimization/src/nodes/_state.hpp
```diff
@@ -163,6 +158,14 @@ class ArrayStateData {
         return true;
     }
 
+    bool set_or_emplace(ssize_t index, double value) {
```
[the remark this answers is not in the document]
wbernoudy: Fair enough. We just end up doing this exact operation many times on dynamic nodes that "pass through" updates. So I figured having a method with an assert is safer.

dwave/optimization/src/nodes/_state.hpp
wbernoudy: Makes sense, but I'm also not sure we want a `resize()` with full functionality... would we want it to fill in with 0 updates when increasing the size?

arcondello: I'm inclined to say yes. Though obviously developers should be aware of that they are doing. There might be cases where that is sensible :shrug:

dwave/optimization/src/nodes/_state.hpp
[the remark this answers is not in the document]
arcondello: Ah, I think for pass-through you should be able to do just `state.update({update})`

arcondello (approved):

PR #324, opened 2025-06-27
Use ArrayStateData set/get methods in NaryOp

wbernoudy:
This will help performance by preventing no-op updates from being propagated by the NaryOp nodes.


wbernoudy:
Close in favor of #325

PR #326, opened 2025-07-03
Add methods for printing information about C++ Nodes

arcondello:
This is hard to test because of some of the compiler-specific behavior. But basically it looks like
### Not topologically sorted
```
 ptr->str(): ExtractNode
ptr->repr(): <ExtractNode at 0x55a9acd0ce78>
       *ptr: <ExtractNode at 0x55a9acd0ce78>
```
### Topologically sorted
```
 ptr->str(): ExtractNode
ptr->repr(): <ExtractNode at 0x55a9acd0ce78, topological_index=2>
       *ptr: <ExtractNode at 0x55a9acd0ce78, topological_index=2>
```


dwave/optimization/src/graph.cpp
```diff
@@ -386,6 +393,56 @@ void Node::initialize_state(State& state) const {
...
+#if defined(_HAS_CXXABI)
+    // if cxxabi.h is available, we can demangle
+    std::string name;
+
+    // __cxa_demangle requires us to manage the lifespan of the returned pointer
+    // so we do this a bit carefully
+    {
+        int status = 0;
+        std::size_t size = 0;
+        char* unmangled = abi::__cxa_demangle(compiler_name, NULL, &size, &status);
```
wbernoudy, line 411: I think you should ensure `status` is equal to 0 after this, otherwise you may be initializing `std::string` from a null pointer and/or trying to free a null pointer.

wbernoudy (approved):

PR #333, opened 2025-07-10
Use std::ranges::view_interface for Array::View

arcondello:
Turns out `std::ranges::view_interface` will do a lot of the work we did for us.


arcondello:
Re the changes in https://github.com/dwavesystems/dwave-optimization/pull/333/commits/3cb41e7557eb69844386be928a60d0d251fa4bf2, this was actually a bug on `main`! Not sure why the compiler didn't check it until these changes though.

PR #338, opened 2025-07-16
ArgSort symbol

wbernoudy:


wbernoudy:
Originally this simply resorted the array (using `std::sort`) on every propagation. However, switching to using `std::set` results in a measured ~35x speedup for propagation of a large 10k items array.

As noted in the comments there is a further optimization that could be made but this implementation is very simple and almost certainly fast enough for now.

dwave/optimization/src/nodes/sorting.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+std::span<const Update> ArgSortNode::diff(const State& state) const {
+    return data_ptr<ArgSortNodeData>(state)->diff();
+}
+
+void ArgSortNode::initialize_state(State& state) const {
+    const Array::View arr = arr_ptr_->view(state);
+
+    std::vector<double> vals{arr.begin(), arr.end()};
+    emplace_data_ptr<ArgSortNodeData>(state, vals);
```
arcondello, line 69:
```c++
    emplace_data_ptr<ArgSortNodeData>(state, std::move(vals));
```
or just
```c++
emplace_data_ptr<ArgSortNodeData>(state, std::vector<double>(arr.begin(), arr.end()));
```

dwave/optimization/src/nodes/sorting.cpp
```diff
@@ -0,0 +1,119 @@
...
+        if (!update.removed()) {
+            node_data->order.insert(std::make_pair(update.value, update.index));
+        }
+    }
+
+    // Assign the new order as determined by the ordering of the std::set.
+    // A further optimization could be to track the earliest modified (final) index
+    // and only assign from there which might help when all the updates only affect
+    // the end region of the final ordering.
+    node_data->assign(node_data->order |
```
arcondello, line 99: Nice!

dwave/optimization/src/nodes/sorting.cpp
```diff
@@ -0,0 +1,119 @@
...
+                      std::views::transform([](std::pair<double, ssize_t> p) { return p.second; }));
```
arcondello, line 100: I think you can do `std::views::transform(std::get<1>)` which might make let it fit on one line. Though perhaps less readable

wbernoudy: I tried that first, unfortunately it doesn't work with `std::transform` without annoying fiddling so I just went with what I thought was the clearer lambda. [See this SO post](https://stackoverflow.com/a/32000899)

dwave/optimization/symbols.pyx
```diff
@@ -437,6 +439,38 @@ cdef class ARange(ArraySymbol):
...
+    def _from_symbol(cls, Symbol symbol):
+        cdef cppArgSortNode* ptr = dynamic_cast_ptr[cppArgSortNode](symbol.node_ptr)
+        if not ptr:
+            raise TypeError(f"given symbol cannot construct a {cls.__name__}")
+        cdef ArgSort x = ArgSort.__new__(ArgSort)
+        x.ptr = ptr
+        x.initialize_arraynode(symbol.model, ptr)
+        return x
+
+    cdef cppArgSortNode* ptr
```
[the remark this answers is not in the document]
wbernoudy: Cool, hadn't written a node like this yet!

Issue #109, opened 2024-09-18
Consider "promoting" arrays/scalars to constants


arcondello:
Currently, in order to use a scalar/array, you need to cast it to a `Constant` symbol. E.g.
```python
model = Model()

x = model.list(5)

a = model.constant([0, 1, 2, 3, 4])  # this step is required

x * a
```

with https://github.com/dwavesystems/dwave-optimization/issues/38 it should be possible to simply do
```python
model = Model()

x = model.list(5)

a = [0, 1, 2, 3, 4]

x * a
```
by promoting `a` to a `Constant` in `x`'s `__mul__()`.

The main problem is that, in addition to #38, we would probably need some sort of caching. Consider
```python
model = Model()

x = model.binary(100)

out = 0
for i in range(100):
    out += x[i] * 2
```
if each time we see `2` we create a new constant, we end up with 100 identical `Constant` symbols, each encoding the same value of `2`. Whereas
```python
model = Model()

x = model.binary(100)
two = model.constant(2)

out = 0
for i in range(100):
    out += x[i] * two
```
only creates one.

arcondello:
One limitation to promotion is that it will likely only be possible for binary operations. So
```python
model = Model()

x = model.list(5)

a = [0, 1, 2, 3, 4]

x * a  # this will work because we're calling x's __mul__() method
a * x  # this will work because first a's __mul__() will fail, so Python will try x's __mul__()
x[a]  # this will work because we're calling x's __index__() method
a[x]  # this will not work
```
which might be a bit mysterious for users.

PR #342, opened 2025-07-22
Enable constant promotion on arithmetic dunder methods

### LINKED ISSUE #38: Consider adding `as_symbol()` method or similar

wbernoudy:
Closes #38 and #109

### DIFF
--- a/dwave/optimization/_model.pyx
+++ b/dwave/optimization/_model.pyx
@@ -1359,6 +1359,12 @@ def _split_indices(indices):
     return tuple(basic_indices), tuple(advanced_indices)
 
 
+def as_array_symbol(model, symbol):
+    if isinstance(symbol, ArraySymbol):
+        return symbol
+    return model.constant(np.asarray(symbol, dtype=np.double))
+
+
 # Ideally this wouldn't subclass Symbol, but Cython only allows a single
 # extension base class, so to support that we assume all ArraySymbols are
 # also Symbols (probably a fair assumption)
@@ -1398,11 +1404,14 @@ cdef class ArraySymbol(Symbol):
         return Absolute(self)
 
     def __add__(self, rhs):
-        if isinstance(rhs, ArraySymbol):
-            from dwave.optimization.symbols import Add  # avoid circular import
-            return Add(self, rhs)
+        try:
+            rhs = as_array_symbol(self.model, rhs)
+        except TypeError:
+            return NotImplemented
+        assert isinstance(rhs, ArraySymbol)
 
-        return NotImplemented
+        from dwave.optimization.symbols import Add  # avoid circular import
+        return Add(self, rhs)
 
     def __bool__(self):
         # In the future we might want to return a Bool symbol, but __bool__ is so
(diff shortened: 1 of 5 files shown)


arcondello (commented):
Needs a pile of tests and a release note.

dwave/optimization/_model.pyx
```diff
@@ -1359,6 +1359,12 @@ def _split_indices(indices):
     return tuple(basic_indices), tuple(advanced_indices)
 
 
+def as_array_symbol(model, symbol):
+    if isinstance(symbol, ArraySymbol):
+        return symbol
+    return model.constant(np.asarray(symbol, dtype=np.double))
```
arcondello, line 1365:
```python
    try:
        return model.constant(symbol)
    except TypeError as err:
        raise TypeError(f"{symbol!r} cannot be interpreted as an array symbol") from err
```
might be a bit more explicit for the context we want to use it and it also let's `Constant` do the work. We should triple check that `Constant` raises a `TypeError` when we expect it to.

tests/test_model.py
```diff
@@ -130,6 +130,7 @@ class UnknownType():
         for op, method, cls in operators:
             with self.subTest(method):
                 self.assertIsInstance(op(x, y), cls)
+                self.assertIsInstance(op(x, 5.7), cls)
```
arcondello, line 133: Ah, I see that this is itself a bunch of tests. Still, a few more explicit ones wouldn't hurt I don't think

dwave/optimization/_model.pyx
```diff
@@ -1398,25 +1404,31 @@ cdef class ArraySymbol(Symbol):
...
 
     def __add__(self, rhs):
-        if isinstance(rhs, ArraySymbol):
-            from dwave.optimization.symbols import Add  # avoid circular import
-            return Add(self, rhs)
+        try:
+            rhs = as_array_symbol(self.model, rhs)
+        except TypeError:
+            return NotImplemented
+        assert isinstance(rhs, ArraySymbol)
```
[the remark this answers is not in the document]
wbernoudy: Fair enough... I think I'll just remove these as `_as_array_symbol()` should guarantee to either raise or return an ArraySymbol

dwave/optimization/_model.pyx
wbernoudy: Agreed!

arcondello:
Oh, also, https://github.com/dwavesystems/dwave-optimization/blob/ab73c3c3675e7afaeb1fb11a30b101ac8046afec/dwave/optimization/symbols.pyx#L2773 and maybe one or two more.

arcondello (approved):

PR #345, opened 2025-07-23
Support array-likes as the left-hand side of binary ops with array symbols

arcondello:
Imports some of the fixes from https://github.com/dwavesystems/dwave-optimization/pull/329. Unrelated change I know, but overlaps enough that was worth fixing I think.


dwave/optimization/_model.pyx
```diff
@@ -1567,6 +1575,46 @@ cdef class ArraySymbol(Symbol):
             return out
         raise ValueError("only integer exponents of 1 or greater are supported")
 
+    def __radd__(self, lhs):
+        try:
+            lhs = _as_array_symbol(self.model, lhs)
+        except TypeError:
+            return NotImplemented
+
+        return lhs + self
```
wbernoudy, line 1584: For the commutative operators, can we avoid the extra `_as_array_symbol()` call here and just `return self + lhs`, letting `__add__` take care of it?

wbernoudy (approved):
LGTM, thanks for picking up my slack.

PR #346, opened 2025-07-23
Use cache by default in `Array::min()/max()`

wbernoudy:


arcondello:
I guess this means that using `minmax()` without a cache will still ignore the cache. Do we want to instead change https://github.com/dwavesystems/dwave-optimization/blob/336f5d00c88d85a1f3fd8028774a4a58abbeb457/dwave/optimization/include/dwave-optimization/array.hpp#L410 to not ignore the cache? That would change things globally I think? Or might require more changes.

wbernoudy:
> I guess this means that using `minmax()` without a cache will still ignore the cache. Do we want to instead change
> 
> https://github.com/dwavesystems/dwave-optimization/blob/336f5d00c88d85a1f3fd8028774a4a58abbeb457/dwave/optimization/include/dwave-optimization/array.hpp#L410
> to not ignore the cache? That would change things globally I think? Or might require more changes.

I think this would be preferred but unfortunately I don't think there is an easy way to initialize the cache from within `memoize()` and have it persist through the recursion. Maybe this could be done in `minmax()` directly but I don't see any easy way without just adding the line to every node's `minmax()` implementation.

PR #356, opened 2025-07-28
CI: test c++23, gcc 15

wbernoudy:


wbernoudy:
C++ appears to be set correctly from looking at the meson logs

.circleci/config.yml
```diff
@@ -223,10 +225,14 @@ jobs:
           name: run c++ tests with meson
           command: |
             . env/bin/activate
-            meson setup _build -Dbuildtype=debug -Dwerror=true
+            meson setup _build -Dbuildtype=debug -Dwerror=true $(if [ "<< parameters.cpp-standard >>" != "default" ]; then echo "-Dcpp_std=<< parameters.cpp-standard >>"; fi)
```
arcondello, line 228: If we want to do conditional, then something like also works.
```
run:
    command: ...
    when: { equal: [<< parameters.cpp-standard >>, "default" ]}
run:
    command: ...
    when: {not : {equal : [<< parameters.cpp-standard >>, "default" ] }}
```

That said, IMO we should just explicitly sweep compiler versions, see other comment

.circleci/config.yml
```diff
@@ -379,8 +385,12 @@ workflows:
       - cpp-gcc:
           matrix:
             parameters:
-              gcc-version: ["11", "12", "13", "14"]
-      - cpp-macOS
+              gcc-version: ["11", "12", "13", "14", "15"]
+              cpp-standard: ["default", "c++23"]
```
arcondello, line 389: I understand you're trying to test that the default works, but IMO we should just explicitly sweep compiler versions. It's simpler and I think this will rarely change.
```
             cpp-standard: ["c++20", "c++23"]
```

.circleci/config.yml
```diff
@@ -379,8 +385,12 @@ workflows:
       - cpp-gcc:
           matrix:
```
arcondello, line 386:
```suggestion
      - cpp-gcc:
          name: cpp-gcc-<< matrix.gcc-version >>-<< matrix.cpp-standard >>
          matrix:
```
otherwise it looks like
image

Issue #353, opened 2025-07-25
Consider caching sizeinfo, min/max, integrality, etc on `ArrayNode`s


arcondello:
Currently we have several different mechanisms for inquiring about `ArrayNode`s.
1) `ArrayNode::sizeinfo()` won't recurse by default and has a [substitute()](https://github.com/dwavesystems/dwave-optimization/blob/3b026b7d8843181e5327dbfef52175681d415da1/dwave/optimization/include/dwave-optimization/array.hpp#L62) method that avoids expensive recursions by only ever deriving it's value from a single predecessor.
2) `ArrayNode::minmax()` allows the user to provide a cache, and `::min()` and `::max()` will cache by default (see https://github.com/dwavesystems/dwave-optimization/pull/346).
3) `ArrayNode::integral()` will just recurse.

This is inconsistent, and sometimes expensive. We should consider adding a mechanism for storing this information on each array node. Perhaps something like [what we have for `ConstantNode`](https://github.com/dwavesystems/dwave-optimization/blob/3b026b7d8843181e5327dbfef52175681d415da1/dwave/optimization/include/dwave-optimization/nodes/constants.hpp#L165).

PR #357, opened 2025-07-28
Add Resize symbol

arcondello:
A few places this differs from NumPy:
* The NumPy method acts in-place. We obviously don't do that.
* The NumPy function doesn't use a fill value, instead it starts over from the beginning of the array
* Neither the method nor the function allow a user-specified fill value.

IMO the approach I took makes sense for our framework and the differences won't be too confusing for users.


dwave/optimization/_model.pyx
```diff
@@ -1835,7 +1835,41 @@ cdef class ArraySymbol(Symbol):
...
 
         return Reshape(self, shape)
-    
+
+    def resize(self, shape, fill_value=None):
+        """Return a new :class:`~dwave.optimization.symbols.Resize` symbol with the given shape.
+
+        Args:
+            shape: Shape of the new array. All dimension sizes must be non-negative.
+            fill_value: The value to be used if the resulting array is larger than the given one.
```
alexzucca90, line 1844: Explaining the default behavior might be useful here :)

dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -578,6 +578,107 @@ std::pair<double, double> ReshapeNode::minmax(
...
+    // our incoming array can be any shape/size, but we cannot be dynamic
+    if (this->dynamic()) throw std::invalid_argument("cannot resize to a dynamic shape");
+
+    // determine whether our fill value will ever matter. We need this info later for
+    // min/max/integral. We could do this lazily, but let's just be proactive for
+    // simplicity
+    if (array_ptr_->dynamic()) {
+        // We'll never use the fill value of the minimize size of our predecessor
+        // array is greater than or equal to our size.
+        const ssize_t min_size = array_ptr_->sizeinfo().substitute(100).min.value_or(0);
```
alexzucca90, line 592: also magic number for max_depth here..

arcondello: yes, the same magic number used everywhere else. There is a bigger fix forthcoming: https://github.com/dwavesystems/dwave-optimization/issues/353

alexzucca90 (approved):
lgtm

PR #359, opened 2025-07-29
Add SoftMaxNode/Symbol

### LINKED ISSUE #281: Consider adding `softmax` node
Softmax: https://en.wikipedia.org/wiki/Softmax_function

fastbodin:
Fix #281. Basic implementation of softmax (https://en.wikipedia.org/wiki/Softmax_function) without axis arguments. 

I am concerned about errors in doubles due to rounding over time but thought it would be best to get feedback on implementation as is.

### DIFF
--- a/dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
+++ b/dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
@@ -421,8 +421,6 @@ class SoftMaxNode : public ArrayOutputMixin<ArrayNode> {
     /// @copydoc Array::sizeinfo()
     SizeInfo sizeinfo() const override;
 
-    double get_denominator(const State& state) const;
-
  private:
     // these are redundant, but convenient
     const Array* arr_ptr_;
--- a/dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
+++ b/dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
@@ -374,6 +374,60 @@ using MinNode = ReduceNode<functional::min<double>>;
 using ProdNode = ReduceNode<std::multiplies<double>>;
 using SumNode = ReduceNode<std::plus<double>>;
 
+// SoftMaxNode *****************************************************************
+
+class SoftMaxNode : public ArrayOutputMixin<ArrayNode> {
+ public:
+    SoftMaxNode(ArrayNode* arr_ptr);
+
+    /// @copydoc Array::buff()
+    double const* buff(const State& state) const override;
+
+    /// @copydoc Node::commit()
+    void commit(State& state) const override;
+
+    /// @copydoc Array::diff()
+    std::span<const Update> diff(const State& state) const override;
+
+    /// @copydoc Node::initialize_state()
+    void initialize_state(State& state) const override;
+
+    /// @copydoc Array::integral()
+    bool integral() const override;
+
+    /// @copydoc Array::minmax()
+    std::pair<double, double> minmax(
+            optional_cache_type<std::pair<double, double>> cache = std::nullopt) const override;
+
+    /// @copydoc Node::propagate()
+    void propagate(State& state) const override;
+
+    /// @copydoc Node::revert()
+    void revert(State& state) const override;
+
+    using Array::shape;
+
+    /// @copydoc Array::shape()
+    std::span<const ssize_t> shape(const State& state) const override;
+
+    using Array::size;
+
+    /// @copydoc Array::size()
+    ssize_t size(const State& state) const override;
+
+    /// @copydoc Array::size_diff()
+    ssize_t size_diff(const State& state) const override;
+
+    /// @copydoc Array::sizeinfo()
+    SizeInfo sizeinfo() const override;
+
+    double get_denominator(const State& state) const;
+
+ private:
+    // these are redundant, but convenient
+    const Array* arr_ptr_;
+};
+
 template <class UnaryOp>
 class UnaryOpNode : public ArrayOutputMixin<ArrayNode> {
  public:
--- a/dwave/optimization/mathematical.py
+++ b/dwave/optimization/mathematical.py
@@ -50,11 +50,8 @@ from dwave.optimization.symbols import (
     Resize,
     Rint,
     SafeDivide,
-<<<<<<< HEAD
     Sin,
-=======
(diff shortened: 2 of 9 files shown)


dwave/optimization/mathematical.py
```diff
@@ -1237,6 +1239,39 @@ def safe_divide(x1: ArraySymbol, x2: ArraySymbol) -> SafeDivide:
...
+        >>> import numpy as np
+        ...
+        >>> model = Model()
+        >>> i = model.integer(3)
+        >>> sm = softmax(i)
+        >>> expected = np.array([0.0900305731703, 0.6652409557748, 0.244728471054])
+        >>> with model.lock():
+        ...     model.states.resize(1)
+        ...     i.set_state(0, [1, 3, 2])
+        ...     print(sum(np.isclose(sm.state(0), expected)))
```
wbernoudy, line 1264: Minor nit but I might frame this as `print(np.isclose(sm.state(), expected).all())`

tests/cpp/nodes/mathematical/test_softmax.cpp
```diff
@@ -0,0 +1,170 @@
...
+        auto softmax_ptr = graph.emplace_node<SoftMaxNode>(a_ptr);
+        graph.emplace_node<ArrayValidationNode>(softmax_ptr);
+
+        WHEN("We initialize a state") {
+            auto state = graph.initialize_state();
+
+            THEN("The initial softmax state is correct") {
+                // softmax should be ~ [0.00216569646006 0.00588697333334
+                // 0.11824302025266 0.87370430995393]
+                CHECK(softmax_ptr->view(state)[0] > 0.00216569646);
```
wbernoudy, line 40: I think you'll want to use `CHECK_THAT(..., WithinRel(...));` for these

fastbodin: Much better that what I was doing. Thanks!

fastbodin:
Made above suggested changes, fixed a few typos, and added a test for when resizing occurs.

dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
```diff
@@ -364,6 +364,60 @@ using MinNode = ReduceNode<functional::min<double>>;
...
+    /// @copydoc Array::size()
+    ssize_t size(const State& state) const override;
+
+    /// @copydoc Array::size_diff()
+    ssize_t size_diff(const State& state) const override;
+
+    /// @copydoc Array::sizeinfo()
+    SizeInfo sizeinfo() const override;
+
+    double get_denominator(const State& state) const;
```
wbernoudy, line 414: I think this method should be private

dwave/optimization/mathematical.py
```diff
@@ -1237,6 +1239,39 @@ def safe_divide(x1: ArraySymbol, x2: ArraySymbol) -> SafeDivide:
...
+        >>> with model.lock():
+        ...     model.states.resize(1)
+        ...     i.set_state(0, [1, 3, 2])
+        ...     print(np.isclose(sm.state(), expected).all())
+        True
+
+    See Also:
+        :class:`~dwave.optimization.symbols.SoftMax`: equivalent symbol.
+
+    .. versionadded:: 0.6.4
```
wbernoudy, line 1270: We'll have to bump these to 0.6.5 as 0.6.4 was just released, sorry!

dwave/optimization/src/nodes/mathematical.cpp
```diff
@@ -1671,6 +1670,147 @@ template class ReduceNode<std::logical_or<double>>;
...
+            new_denominator += exp_val;
+            // Temporarily store in format of unmodified elements. Will multiply
+            // all elements by prior_denominator / new_denominator after.
+            node_data->set(u.index, exp_val / prior_denominator, true);
+        } else {
+            double exp_val = std::exp(u.value);
+            new_denominator += exp_val - node_data->get(u.index) * prior_denominator;
+            // Temporarily store in format of unmodified elements. Will multiply
+            // all elements by prior_denominator / new_denominator after.
+            node_data->set(u.index, exp_val / prior_denominator, true);
```
wbernoudy, line 1768: One issue I see with using `set()` here is that it will cause an `Update` to be emitted on each changed/placed value of the predecessor, but then if the denominator has changed (which it very likely has if there are any updates to the pred array), another `Update` will be emitted when the value is adjusted for the new denominator.

I'm thinking that in this loop, you should not use `set()` to avoid emitting these extraneous updates, and then use `set()` as you update them for the new denominator. This would mean you can't (easily) do the check to see if the new denominator is different from the prior denominator and skip emitting the updates if so.

fastbodin: I forgot that doing this could result in two updates for the same index. This makes extra work for any predecessor. Good point.

fastbodin: I used `std::vector` to track the indices I need to divide by the new denominator and to record the `exp()` values. I could alternatively use `std::unordered_map`. Recommendations for which is best for this scenario?

dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
wbernoudy: IMO there's no need to test the denominator directly in the unit tests, I would just remove those

fastbodin: Removed. To make up for the fact that we are no longer testing it directly, I added a test which 
- `commits` a `state`,
- `propagates` a `state`,
- `reverts` a  `state`, and
- `propagates` a `state`.

This can only result in the correct values of SoftMaxNode at the end if the denominator is properly reset in revert.

fastbodin:
Corrections include:
1. Removed SoftMaxNode::get_denominator and relevant C++ tests. Added a C++ test to compensate.
2.  Opted to not implement a method to check if `prior_denominator == new_denominator` (which could reduce the # of updates) since this edge case is unlikely in practical use cases. 
3. Opted to continue to cache updated `exp()` in `std::unordered_map` (as opposed to recording the indices that have been updated and recompute the value at these indices fully with an additional `exp()` call) because there was a bug that occurred when an update was called more than once on the same index. This choice increases memory usage but should result in faster computations.

PR #364, opened 2025-08-05
Add FindNode/Symbol

### LINKED ISSUE #139: Consider adding `Find` symbol or similar
A symbol that would return the index of the first instance of a value in an array.

fastbodin:
Added FirstInstanceNode cpp base class which returns the smallest index (should it exist) of element of an array where the `condition` is true. The `condition` must be defined by derived class. Returns -1 if no such index exists.

Added FindNode/Symbol which returns the smallest index (should it exist) of a non-zero element of predecessor. This class is derived from FirstInstanceNode. 

Possibly fix #139. Depends on what is needed, Find() vs ArgWhere().

### DIFF
--- a/dwave/optimization/include/dwave-optimization/nodes/flow.hpp
+++ b/dwave/optimization/include/dwave-optimization/nodes/flow.hpp
@@ -79,9 +79,9 @@ class ExtractNode : public ArrayOutputMixin<ArrayNode> {
 
 // FirstInstanceNode *****************************************************************
 //
-// A base class that returns the smallest index (should it exist) of element of
-// an array where the `condition` is true. The `condition` must be defined by
-// derived class.
+// A base class that returns the smallest index (should it exist) of the
+// element of an array where the `condition` is true. The `condition` must be
+// defined by derived class.
 class FirstInstanceNode : public ScalarOutputMixin<ArrayNode, true> {
  public:
     explicit FirstInstanceNode(ArrayNode* arr_ptr);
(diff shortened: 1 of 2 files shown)


arcondello:
The code looks reasonable, though I have only done a quick pass. But before jumping into that I need to do some ruminating about what API we want here. This PR implements https://github.com/dwavesystems/dwave-optimization/issues/139 as specified so these concerns are entirely my fault :flushed: Nonetheless I need to think a bit about them.

As a general design principal, we like our nodes/symbols to correspond to NumPy methods or functions. The reasoning here is that we assume our users are reasonably familiar with NumPy and therefore if we deviate too much from them we add confusion. Unfortunately, NumPy doesn't provide a `find()` function. A quick look through their [searching functions](https://numpy.org/devdocs/reference/routines.sort.html#searching) turns up a few candidates, most notably [`np.argwhere()`](https://numpy.org/doc/stable/reference/generated/numpy.argwhere.html). But, annoyingly, `np.argwhere()` returns _all_ indices where the condition matches. 

This is where our attempts to match NumPy get tricky. Because while we can support that behavior with a dynamically sized node, dynamically sized nodes are less well behaved and don't work out-of-the-box with all other nodes. For example, a user might know _by construction_ that the array they are searching through will have unique elements and will always include the element they are searching for, but the node might not know that.

OTOH, returning the indices of all elements that match the condition has the advantage of handling the no-match case (though returning `-1` as you do in this PR also makes sense if we must return something).

Anyway, I need to do some thinking about design. Sorry for the wall of text, but wanted to give you some idea as to why this PR hasn't been addressed yet.

fastbodin:
Fixed issue in `FirstInstanceNode::propagate` where an index in predecessor was changed multiple times. 

**Example:** Predecessor = `[0, 0, 1, 2]` and condition is `non-zero`. Previously, if predecessor changed at index 2 multiple times and the final update resulted in it being non-zero, this would result in `FindNode` incorrectly reporting index 3 being the first non-zero index of predecessor. 

This updated resulted in a simplified `FirstInstanceNode::propagate`. Added a C++ test to check the case described above.

PR #367, opened 2025-08-07
Add trigonometric sine and cosine

arcondello:


dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
```diff
@@ -34,6 +34,11 @@ struct abs {
     constexpr T operator()(const T& x) const { return std::abs(x); }
 };
 
+template <class T>
+struct cos {
```
wbernoudy, line 38: We need these so that they accept any type right?

wbernoudy (approved):

dwave/optimization/include/dwave-optimization/nodes/mathematical.hpp
arcondello: It's because we need `BinaryOp` to be a class, and `std::cos` and `std::sin` are functions.

PR #368, opened 2025-08-07
Add IsInNode/Symbol

### LINKED ISSUE #76: Consider adding `Contains` symbol or similar
Something like [`np.isin()`](https://numpy.org/doc/stable/reference/generated/numpy.isin.html).

fastbodin:
Initial minimal implementation of [numpy.isin() ](https://numpy.org/doc/stable/reference/generated/numpy.isin.html#numpy.isin) using lookup tables. Fixes #76. Given that this fixes an old issue, I wanted to check which functionality from [numpy.isin()](https://numpy.org/doc/stable/reference/generated/numpy.isin.html#numpy.isin) we should port over for this node. Should this node include any of the following?

1. `bool assume_unique` (kwarg)
2. `bool invert` (kwarg)
3. `kind {None, sort, table}` (kwarg)

Regarding 2, I believe this could just be done with the EqualNode. For example:
```
>>> from dwave.optimization.mathematical import isin
>>> model = Model()
>>> model.states.resize(1)
>>> element = model.constant([1.3, -1.0, 3.0])
>>> test_elements = model.constant([3.0, -1.1, -2.0, 4.1, -1.0])
>>> contains = isin(element, test_elements)
>>> inverted_contains = contains == [0.0, 0.0, 0.0]
>>> with model.lock():
...     print(contains.state(0))
...     print(inverted_contains.state(0))
[0. 1. 1.]
[1. 0. 0.]
```

Regarding 3, given we already have an ArgSortNode, it seems reasonable that we could pass this node a bool like `sorted` which could indicate whether its predecessors are ArgSortNodes. We could implement methods to deal with these cases and call them or the current lookup table methods from `IsInNode::initialize_state` and `IsInNode::propagate`.

I believe there is additional optimization that can be done to `IsInNode::propagate` but I wanted to deal with the above questions before optimizing it further.

### DIFF
--- a/dwave/optimization/include/dwave-optimization/nodes/set_routines.hpp
+++ b/dwave/optimization/include/dwave-optimization/nodes/set_routines.hpp
@@ -22,8 +22,6 @@
 
 namespace dwave::optimization {
 
-struct IsInNodeData;
-
 class IsInNode : public ArrayOutputMixin<ArrayNode> {
  public:
     IsInNode(ArrayNode* element_ptr, ArrayNode* test_elements_ptr);
@@ -75,8 +73,6 @@ class IsInNode : public ArrayOutputMixin<ArrayNode> {
     // these are redundant, but convenient
     const Array* element_ptr_;
     const Array* test_elements_ptr_;
-
-    inline void rm_index(IsInNodeData*& node_data, const ssize_t rm_index, const double key) const;
 };
 
 }  // namespace dwave::optimization
--- a/dwave/optimization/include/dwave-optimization/nodes/set_routines.hpp
+++ b/dwave/optimization/include/dwave-optimization/nodes/set_routines.hpp
@@ -74,8 +74,7 @@ class IsInNode : public ArrayOutputMixin<ArrayNode> {
     const Array* element_ptr_;
     const Array* test_elements_ptr_;
 
-    inline void rm_index_from_element_indices_key(IsInNodeData*& node_data, const ssize_t rm_index,
-                                                  const double key) const;
+    inline void rm_index(IsInNodeData*& node_data, const ssize_t rm_index, const double key) const;
 };
 
 }  // namespace dwave::optimization
--- a/dwave/optimization/include/dwave-optimization/nodes/set_routines.hpp
+++ b/dwave/optimization/include/dwave-optimization/nodes/set_routines.hpp
@@ -22,6 +22,8 @@
 
 namespace dwave::optimization {
 
+struct IsInNodeData;
+
 class IsInNode : public ArrayOutputMixin<ArrayNode> {
  public:
     IsInNode(ArrayNode* element_ptr, ArrayNode* test_elements_ptr);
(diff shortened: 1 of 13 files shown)


dwave/optimization/src/nodes/set_routines.cpp
```diff
@@ -0,0 +1,202 @@
...
+    for (const Update& update : test_elements_diff) {
+        if (!update.placed()) {  // i.e. changed or removed
+            test_elements_counter[update.old] -= 1;
+
+            if (test_elements_counter[update.old] == 0) {
+                // Remove key to reduce bloating
+                test_elements_counter.erase(update.old);
+                // Each index of `element` with the value `update.old` is no
+                // longer in `test_elements`
+                for (const auto index : element_indices[update.old]) {
```
fastbodin, line 110: Based on feedback from SoftMaxNode, this needs to change to reduce multiple updates on the same index.

dwave/optimization/src/nodes/set_routines.cpp
```diff
@@ -0,0 +1,254 @@
...
+                                   std::vector<double>{test_elements.begin(), test_elements.end()});
+}
+
+bool IsInNode::integral() const { return true; }  // All values are true/false
+
+double IsInNode::min() const { return 0.0; }  // All values are true/false
+
+double IsInNode::max() const { return 1.0; }  // All values are true/false
+
+inline void IsInNode::rm_index(IsInNodeData*& node_data, const ssize_t rm_index,
```
wbernoudy, line 102: I think you can leave this as a function in the cpp without adding it to the `IsInNode` class. Also, I would rename either the function or the `rm_index` argument to avoid confusion

dwave/optimization/src/nodes/set_routines.cpp
```diff
@@ -0,0 +1,254 @@
...
+    // Save a copy of the updates for IsInNode::revert()
+    node_data->test_element_updates.assign(test_elements_diff.begin(), test_elements_diff.end());
+    node_data->element_updates.assign(element_diff.begin(), element_diff.end());
+
+    // Handle changes to `test_elements`
+    for (const Update& update : test_elements_diff) {
+        if (!update.placed()) {
+            // i.e. changed or removed.
+            ssize_t& count = node_data->set_data[update.old].test_elements_count;
+            // `update.old` must be in `test_elements_counter`
```
wbernoudy, line 142: I think this could go in the assert like `assert(count > 0 && "update.old should be in test_elements_counter")`

dwave/optimization/src/nodes/set_routines.cpp
fastbodin: Good point.

dwave/optimization/include/dwave-optimization/nodes/set_routines.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+#include <span>
+
+#include "dwave-optimization/array.hpp"
+#include "dwave-optimization/graph.hpp"
+#include "dwave-optimization/state.hpp"
+
+namespace dwave::optimization {
+
+struct IsInNodeData;
```
arcondello, line 25: I don't think this forward declaration gets used here

fastbodin:
Face palm. Thought the branch had been merged so I deleted it.

PR #369, opened 2025-08-08
Allow reshaping some dynamic nodes

arcondello:
In a chicken-and-egg way, I need this to test the `BroadcastToNode` I am working on, and that node would make this one easier to test. So doing this one first and then I'll add a few more tests in that next PR.

### DIFF
--- a/dwave/optimization/include/dwave-optimization/nodes/manipulation.hpp
+++ b/dwave/optimization/include/dwave-optimization/nodes/manipulation.hpp
@@ -155,7 +155,6 @@ class PutNode : public ArrayOutputMixin<ArrayNode> {
     const Array* values_ptr_;
 };
 
-
 /// Propagates the values of its predecessor, interpreted into a different shape.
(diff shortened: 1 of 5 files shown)


dwave/optimization/include/dwave-optimization/nodes/manipulation.hpp
```diff
@@ -182,16 +181,35 @@ class ReshapeNode : public ArrayOutputMixin<ArrayNode> {
...
+    void initialize_state(State& state) const override;
+
     /// @copydoc Array::integral()
     bool integral() const override;
 
     /// @copydoc Array::minmax()
     std::pair<double, double> minmax(
             optional_cache_type<std::pair<double, double>> cache = std::nullopt) const override;
 
+    /// @copydoc Node::initialize_state()
```
wbernoudy, line 194:
```suggestion
    /// @copydoc Node::propagate()
```

PR #377, opened 2025-09-04
Fix the creation of Permutation symbols

arcondello:
In some cases we would get errors when we shouldn't.


dwave/optimization/symbols.pyx
```diff
@@ -3474,7 +3496,7 @@ cdef class Permutation(ArraySymbol):
         >>> type(p)
         <class 'dwave.optimization.symbols.Permutation'>
     """
-    def __init__(self, Constant array, ListVariable x):
+    def __init__(self, Constant array, ArraySymbol x):
```
arcondello, line 3499: This is the main fix, but there were a few other edge cases as well

PR #378, opened 2025-09-04
Eager min/max/integral calculation

wbernoudy:
Related: #353 

This PR eliminates `Array::minmax()` and instead requires array nodes to implement `min()` and `max()`. It re-implements all `min()`, `max()` and `integral()` on all the existing array nodes to ensure they do not call any methods on predecessors (thus ensuring there are no exponential recursion issues during model construction).

### DIFF
--- a/dwave/optimization/include/dwave-optimization/array.hpp
+++ b/dwave/optimization/include/dwave-optimization/array.hpp
@@ -281,10 +281,10 @@ class Array {
     /// A std::random_access_iterator over the values in the array.
     using const_iterator = BufferIterator<double, double, true>;
 
-    template<class T>
+    template <class T>
     using cache_type = std::unordered_map<const Array*, T>;
 
-    template<class T>
+    template <class T>
     using optional_cache_type = std::optional<std::reference_wrapper<cache_type<T>>>;
 
     using View = std::ranges::subrange<const_iterator>;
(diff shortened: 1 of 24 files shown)


wbernoudy:
Realizing now I could have made `Array::minmax()` non-virtual and implemented it on `Array`, thus avoiding some changes where nodes call `minmax()`. I don't mind doing that if it makes sense.

dwave/optimization/include/dwave-optimization/graph.hpp
```diff
@@ -374,4 +374,8 @@ class DecisionNode : public Decision, public virtual Node {
     bool removable() const override { return false; }
 };
 
+std::vector<const Array*> cast_to_array(std::span<const ArrayNode* const> node_ptrs);
```
arcondello, line 377: Could/should replace with
```c++
auto cast_to_array(std::span<const ArrayNode* const> ptrs) {
    return std::ranges::transform_view(ptrs, [](auto ptr) -> const Array* { return ptr; });
}
```
or
```c++
template<ranges::viewable_range R>
auto cast_to_array(R&& ptrs) {
    return std::ranges::transform_view(ptrs, [](auto ptr) -> const Array* { return ptr; });
}
```
we could use `require` and `std::range_value_t` if we're worried about using ranges of other types.

Or, IMO best of all, just use `std::ranges::transform_view(ptrs, [](auto ptr) -> const Array* { return ptr; })` inline anywhere we might want this behavior and not add a new function at all!

wbernoudy: It is nice to have a version of this for `initializer_list` though, which I don't think we can use `transform_view` on?

dwave/optimization/include/dwave-optimization/nodes/constants.hpp
```diff
@@ -159,18 +159,11 @@ class ConstantNode : public ArrayOutputMixin<ArrayNode> {
     // holds its values on the object itself rather than in a State.
     const double* buffer_ptr_;
 
-    std::unique_ptr<DataSource> data_source_;
+    ValuesInfo calculate_values_info(std::span<const double> buffer);
```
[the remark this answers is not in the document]
wbernoudy: Yep I should, was just being lazy with moving all the constructors to the `.cpp` as well

dwave/optimization/include/dwave-optimization/graph.hpp
wbernoudy: Switched to just two constructors on `ValuesInfo` for this, one for `initializer_list` and one that takes `viewable_range` directly so these methods are no longer necessary.

wbernoudy: ... and that doesn't work with GCC11 so had to add the `span` constructor back.

PR #383, opened 2025-09-09
Break `symbols.pyx` into a subpackage

arcondello:


dwave/optimization/symbols/_unaryop.pyx
```diff
@@ -0,0 +1,104 @@
...
+        elif node_type is _UnaryOpNodeType.SquareRoot:
+            _register(cls, typeid(SquareRootNode))
+        else:
+            raise RuntimeError(f"unexpected _UnaryOpNodeType: {<object>node_type!r}")
+
+        cls._node_type = node_type
+
+    def __init__(self, ArraySymbol x):
+        cdef _Graph model = x.model
+        cdef _UnaryOpNodeType node_type = self._node_type
```
arcondello, line 73: We should change this so it raises a nicer error if the user tries to construct `_UnaryOpSymbol` directly

PR #393, opened 2025-09-17
Fix modifying Model.objective directly

arcondello:
Prior to https://github.com/dwavesystems/dwave-optimization/pull/179, `Model.objective` was readonly. After that PR it was possible to set or mutate the `model.objective`. And while that would not update the underlying C++ model, the serialization would read the Python attribute. That is until https://github.com/dwavesystems/dwave-optimization/pull/166 changed the behavior to read from the underlying graph.

This PR makes support for modifying `Model.objective` official.

### DIFF
--- a/dwave/optimization/model.py
+++ b/dwave/optimization/model.py
@@ -119,28 +119,6 @@ class Model(_Graph):
         >>> model = flow_shop_scheduling(processing_times=processing_times)
     """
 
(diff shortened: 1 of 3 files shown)


arcondello:
Incidentally, we don't actually support clearing the objective from the Python level and apparently never have. We can add support for that in a followup.

wbernoudy (approved):

PR #397, opened 2025-09-24
QOL improvements for DisjointLists

wbernoudy:
Now `Model.disjoint_lists()` returns only the `DisjointLists` symbol. The `DisjointList` successors can be accessed by indexing/iterating the `DisjointLists` symbol.

This is backwards compatibility-breaking.


arcondello (commented):
I agree we should implement `DisjointLists.__getitem__()`.

For the return value, I agree this is much better but unfortunately changing the return value is a significant backwards compatibility break so we'd want to save that for a major release.

PR #400, opened 2025-09-26
Fix BroadcastTo symbol's diff propagation

arcondello:


dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -47,21 +47,18 @@ std::vector<ssize_t> diff_offsets(std::span<const ssize_t> from_shape,
...
-            // In the case that we're dynamic, this will result in a negative multiplier,
-            // but that's OK because it's only the 0th dimension so we'll never use it.
-            multiplier *= *from_it;
-        } else {
+        if (*from_it != *to_it) {
             assert(*from_it == 1);  // should be checked in constructor
             offsets = expand(offsets, *to_it, multiplier);
         }
+
+        multiplier *= *to_it;
```
arcondello, line 55: Turns out we always need to take into account the size of the "upstream" dimensions. Which seems obvious in hindsight.

wbernoudy (approved):

PR #402, opened 2025-10-07
Unified `ReduceNode`

arcondello:
DRAFT PR. There's still a lot of cruft in the code that needs to cleaned out. But I want to break out a few smaller features and I think it will help to have a reference to what they are being implemented for.

### DIFF
--- a/dwave/optimization/include/dwave-optimization/nodes/reduce.hpp
+++ b/dwave/optimization/include/dwave-optimization/nodes/reduce.hpp
@@ -64,15 +64,15 @@ class ReduceNode2 : public ArrayOutputMixin<ArrayNode> {
     /// @copydoc Array::min()
     double min() const override;
 
-    // // The predecessor of the reduction, as an Array*.
-    // std::span<Array* const> operands() {
-    //     assert(predecessors().size() == 1);
-    //     return std::span<Array* const, 1>(&array_ptr_, 1);
-    // }
-    // std::span<const Array* const> operands() const {
-    //     assert(predecessors().size() == 1);
-    //     return std::span<const Array* const, 1>(&array_ptr_, 1);
-    // }
+    // The predecessor of the reduction, as an Array*.
+    std::span<Array* const> operands() {
+        assert(predecessors().size() == 1);
+        return std::span<Array* const, 1>(&array_ptr_, 1);
+    }
+    std::span<const Array* const> operands() const {
+        assert(predecessors().size() == 1);
+        return std::span<const Array* const, 1>(&array_ptr_, 1);
+    }
 
     void propagate(State& state) const override;
 
@@ -91,19 +91,27 @@ class ReduceNode2 : public ArrayOutputMixin<ArrayNode> {
     /// Otherwise uses the first element in the reduction.
     const std::optional<double> initial;
  private:
+    // std::ranges::subrange<BufferIterator<double, double, true>, std::default_sentinel_t>
+    auto reduce_(const State& state, ssize_t index) const;
+
     BinaryOp op;
 
-    ArrayNode* const array_ptr_;
+    Array* const array_ptr_;
 
     // The axes we're reducing in a sorted, unique vector.
     // An empty vector means we're reducing over everything
     std::vector<ssize_t> axes_;
 
-    // The minimum and maximum (inclusive) value that might be returned
+    // The minimum and maximum (inclusive) value that might be returned as well
+    // as whether we're integral or not.
     const ValuesInfo values_info_;
 };
 
 
+using AllNode2 = ReduceNode2<std::logical_and<double>>;
+using AnyNode2 = ReduceNode2<std::logical_or<double>>;
+using MaxNode2 = ReduceNode2<functional::max<double>>;
+using MinNode2 = ReduceNode2<functional::min<double>>;
 using ProdNode2 = ReduceNode2<std::multiplies<double>>;
 using SumNode2 = ReduceNode2<std::plus<double>>;
 
--- a/dwave/optimization/include/dwave-optimization/nodes/reduce.hpp
+++ b/dwave/optimization/include/dwave-optimization/nodes/reduce.hpp
@@ -55,8 +55,8 @@ class ReduceNode2 : public ArrayOutputMixin<ArrayNode> {
     /// @copydoc Node::initialize_state()
     void initialize_state(State& state) const override;
 
-    // /// @copydoc Array::integral()
-    // bool integral() const override;
+    /// @copydoc Array::integral()
+    bool integral() const override;
 
     /// @copydoc Array::max()
     double max() const override;
@@ -100,10 +100,11 @@ class ReduceNode2 : public ArrayOutputMixin<ArrayNode> {
     std::vector<ssize_t> axes_;
 
     // The minimum and maximum (inclusive) value that might be returned
-    std::pair<double, double> minmax_;
+    const ValuesInfo values_info_;
 };
 
 
+using ProdNode2 = ReduceNode2<std::multiplies<double>>;
 using SumNode2 = ReduceNode2<std::plus<double>>;
 
 
--- a/dwave/optimization/include/dwave-optimization/nodes/reduce.hpp
+++ b/dwave/optimization/include/dwave-optimization/nodes/reduce.hpp
@@ -31,6 +31,82 @@
 
 namespace dwave::optimization {
 
+template <class BinaryOp>
+class ReduceNode2 : public ArrayOutputMixin<ArrayNode> {
+ public:
+    ReduceNode2(ArrayNode* array_ptr);
+
+    ReduceNode2(ArrayNode* array_ptr, std::initializer_list<ssize_t> axes,
+                std::optional<double> initial = std::nullopt);
+    ReduceNode2(ArrayNode* array_ptr, std::span<const ssize_t> axes,
+                std::optional<double> initial = std::nullopt);
+
+    // std::span<const ssize_t> axes() const;
+
+    /// @copydoc Array::buff()
+    double const* buff(const State& state) const override;
+
+    /// @copydoc Node::commit()
+    void commit(State& state) const override;
+
+    /// @copydoc Array::diff()
+    std::span<const Update> diff(const State& state) const override;
+
+    /// @copydoc Node::initialize_state()
+    void initialize_state(State& state) const override;
+
+    // /// @copydoc Array::integral()
+    // bool integral() const override;
+
+    /// @copydoc Array::max()
+    double max() const override;
+
+    /// @copydoc Array::min()
+    double min() const override;
+
+    // // The predecessor of the reduction, as an Array*.
+    // std::span<Array* const> operands() {
+    //     assert(predecessors().size() == 1);
+    //     return std::span<Array* const, 1>(&array_ptr_, 1);
+    // }
+    // std::span<const Array* const> operands() const {
+    //     assert(predecessors().size() == 1);
+    //     return std::span<const Array* const, 1>(&array_ptr_, 1);
+    // }
+
+    void propagate(State& state) const override;
+
+    /// @copydoc Node::revert()
+    void revert(State& state) const override;
+
+    using ArrayOutputMixin::shape;
+    std::span<const ssize_t> shape(const State& state) const override;
+
+    using ArrayOutputMixin::size;
+    ssize_t size(const State& state) const override;
+
+    ssize_t size_diff(const State& state) const override;
+
+    /// The initial value if one was provided.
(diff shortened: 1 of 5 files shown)


tests/cpp/nodes/test_reduce.cpp
```diff
@@ -28,13 +29,118 @@ using Catch::Matchers::RangeEquals;
...
+                graph.propagate(state);
+                CHECK_THAT(x_ptr->view(state), RangeEquals({11, 11, 11, 11, 8, 8, 8, 8}));
+
+                graph.revert(state);
+                CHECK_THAT(x_ptr->view(state), RangeEquals({5, 5, 5, 5, 8, 8, 8, 8, 14, 14, 14, 14, 11, 11, 11, 11}));
+            }
+        }
+    }
+
+    // todo: check growing then shrinking as part of the same propagation
```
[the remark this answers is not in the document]
arcondello: You're right. And looking through this with semi-fresh eyes this whole file needs a reorganization and expansion.

dwave/optimization/src/nodes/reduce.cpp
```diff
@@ -22,996 +22,741 @@
...
-    }
-    void revert() {
-        values.value = values.old;
-        extra.revert();
-    }
+ValuesInfo values_info(const Array* array_ptr, std::span<const ssize_t> axes,
+                       std::optional<double> initial) {
+    // This function assumes axes is sorted and unique
+    assert(std::ranges::is_sorted(axes));
+    assert(is_unique(axes));
```
wbernoudy: Can't you use `std::ranges::unique()` here as above?

dwave/optimization/src/nodes/reduce.cpp
```diff
@@ -22,996 +22,741 @@
...
+        static_assert(decltype(ufunc)::associative);
+
+        // Check that we're not trying to add a removal
+        assert(not std::isnan(value));
+
+        // And that we're within bounds. One past the end is OK
+        assert(0 <= index and static_cast<std::size_t>(index) <= reductions_.size());
+
+        if (static_cast<std::size_t>(index) == reductions_.size()) {
+            // If we're one-past-end then it's an append (which even rhymes!)
```
fastbodin: Ha! Appreciated.

dwave/optimization/src/nodes/reduce.cpp
```diff
@@ -22,996 +22,741 @@
...
+            case ReductionFlag::unchanged:
+                // We haven't previously made any changes to this location, so we
+                // save the old value, mark ourselves as updated, and then
+                //
+                reductions_diff_.emplace_back(index, reductions_[index]);
+                reductions_[index] = std::move(reduction);
+                flags_[index] = ReductionFlag::updated;
+                break;
+            case ReductionFlag::updated:
+                // We've already updated this index once, so need to save the old value
```
fastbodin: so **no** need?

dwave/optimization/symbols/_reduce.pyx
```diff
@@ -0,0 +1,201 @@
...
+        elif node_type is _ReduceNodeType.Any:
+            initial = (<AnyNode*>ptr).initial
+        elif node_type is _ReduceNodeType.Max:
+            initial = (<MaxNode*>ptr).initial
+        elif node_type is _ReduceNodeType.Min:
+            initial = (<MinNode*>ptr).initial
+        elif node_type is _ReduceNodeType.Prod:
+            initial = (<ProdNode*>ptr).initial
+        elif node_type is _ReduceNodeType.Sum:
+            initial = (<SumNode*>ptr).initial
```
wbernoudy: Could having a base class help with these? Other than the one in the init

dwave/optimization/src/nodes/reduce.cpp
```diff
@@ -22,996 +22,741 @@
...
-    }
+        // Add the value to the reduction first
+        reduction = ufunc(std::move(reduction), to);
 
-    // Validate axes
-    /// TODO: support negative axes
-    if (axes.size() != 1) {
-        throw std::invalid_argument("Partial reduction support only one axis");
-    }
+        // Then try to remove the value
```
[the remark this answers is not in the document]
wbernoudy: That shouldn't be possible due to the constraints on size and having an initial value I believe

dwave/optimization/src/nodes/reduce.cpp
```diff
@@ -22,996 +22,741 @@
...
+        // likewise these should be unique
+        for (const auto& [index, old_value, _] : diff_ | std::views::reverse) {
+            assert(0 <= index and static_cast<std::size_t>(index) <= buffer_.size());
+            if (static_cast<std::size_t>(index) == buffer_.size()) {
+                buffer_.emplace_back(old_value);
+            } else {
+                buffer_[index] = old_value;
+            }
+        }
+        diff_.clear();
```
fastbodin: Why is it that we do not need to resize `buffer_` like we do `reductions_`?

dwave/optimization/src/nodes/reduce.cpp
```diff
@@ -22,996 +22,741 @@
...
-}
+    // Revert any changes that have been made to the buffer
+    void revert() {
+        // Confirm that all of our flags have been reset
+        assert(std::ranges::all_of(flags_, [](const ReductionFlag& flag) {
+            return flag == ReductionFlag::unchanged;
+        }));
+
+        // First revert changes to the reductions_
+        if (shape_info_) reductions_.resize(shape_info_->previous_size, reduction_type(0));
```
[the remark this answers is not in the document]
arcondello: Yes, exactly

dwave/optimization/src/nodes/reduce.cpp
```diff
@@ -22,996 +22,739 @@
 #include <tuple>
 #include <type_traits>
```
fastbodin: Do not believe this is used?

dwave/optimization/src/nodes/reduce.cpp
```diff
@@ -22,996 +22,739 @@
...
-            // if low is positive, then we're interested in the minimum size
-            // of the array
-            low = init_ + sizeinfo.min.value_or(0) * low;
-        }
+    auto bounds = min_bounds | max_bounds;  // the union of the domains
 
-        return {low, high, integral};
-    }
+    // One more case (empty array)
+    if (min_size == 0) bounds |= initial_bounds;
```
fastbodin: This is because, at a minimum, the initial value need to be accepted? And if the array is empty then the default values for `bounds` may not admit the initial value?

dwave/optimization/include/dwave-optimization/nodes/reduce.hpp
fastbodin: Do not believe this is used?

dwave/optimization/src/nodes/reduce.cpp
```diff
@@ -22,996 +22,739 @@
...
     }
 
-    Node::propagate(state);
-}
-
-template <>
-void ReduceNode<std::multiplies<double>>::propagate(State& state) const {
-    auto ptr = data_ptr<ReduceNodeData<std::multiplies<double>>>(state);
+    const auto subspace_shape = keep_axes(array_ptr_->shape(state), axes_);  // state-dependent!
+    const auto subspace_strides = keep_axes(array_ptr_->strides(), axes_);
```
fastbodin: Rather convenient that `strides()` is not state-dependent because arrays can only be dynamic in the 0th axis and the stride length to move along the 0th axis is dependant upon all the other axes which are fixed.... It all comes together 🧠 🌈

dwave/optimization/src/nodes/reduce.cpp
```diff
@@ -22,996 +22,739 @@
...
-    if (ptr->values.value != ptr->values.old) Node::propagate(state);
-}
+    // It's OK for these to both be dynamic
+    auto array_shape = array_ptr_->shape();
+    auto reduce_shape = this->shape();
 
-template <>
-void ReduceNode<std::logical_or<double>>::propagate(State& state) const {
-    auto ptr = data_ptr<ReduceNodeData<std::logical_or<double>>>(state);
+    if (this->dynamic() and array_ptr_->size_diff(state) > 0 and initial) {
```
fastbodin: There currently is no test for which `this-dynamic() == true`, `array_ptr_->size_diff(state) > 0`, and no `initial value` is provided. This feels like a something worth testing?

dwave/optimization/src/nodes/reduce.cpp
```diff
@@ -22,996 +22,739 @@
...
-    // If the output of the operation is boolean, the min/max are just [false, true]
-    using result_type = typename std::invoke_result<BinaryOp, double&, double&>::type;
-    if constexpr (std::same_as<result_type, bool>) {
-        return ValuesInfo::logical_output();
+            return ValuesInfo(init, init, is_integer(init));
+        } else {
+            throw std::invalid_argument(
+                    "cannot perform a reduction operation with no "
+                    "identity on an array that might be empty");
+        }
```
fastbodin: I believe this is not tested. Might be worth throwing something like this in the `C++` tests.

```cpp
    GIVEN("An empty array reduced without an explicit init") {
        std::vector<double> values_a = {};
        auto a_ptr = graph.emplace_node<ConstantNode>(values_a);
        CHECK_THROWS(graph.emplace_node<ReduceNode<TestType>>(a_ptr));
    }
```

tests/cpp/nodes/test_reduce.cpp
```diff
@@ -456,9 +192,63 @@ TEMPLATE_TEST_CASE("ReduceNode", "", functional::max<double>, functional::min<do
...
+            }
+        }
+    }
 }
 
-TEST_CASE("ReduceNode - AllNode/AnyNode") {
+TEST_CASE("AllNode/AnyNode") {
     auto graph = Graph();
 
     GIVEN("x = BinaryNode({5}), y = x.any()") {
```
fastbodin:
```cpp
    GIVEN("x = BinaryNode({5}), y = x.all(), z = x.any()") {
```

tests/cpp/nodes/test_reduce.cpp
fastbodin: In this test, we test `grow` and `shrink`. I would opt to include a `grow` then `shrink` or a `shrink` then `grow` as well. Something like 
```cpp
            AND_WHEN("We grow the set by three and shrink by one") {
                a_ptr->grow(state);
                a_ptr->grow(state);
                a_ptr->grow(state);
                a_ptr->shrink(state);
                a_ptr->propagate(state);
                r_ptr->propagate(state);
                a_ptr->commit(state);
                r_ptr->commit(state);

                THEN("The reduction has the value and shape we expect") {
                    CHECK(r_ptr->ndim() == 0);
                    CHECK(r_ptr->size(state) == 1);
                    CHECK(r_ptr->view(state)[0] == std::reduce(a_ptr->view(state).begin(),
                                                               a_ptr->view(state).end(), init,
                                                               func));
                }
            }
```

fastbodin:
```cpp
                THEN("prod([ 1 2 3 4 ]) == 24") { CHECK(prod_ptr->view(state)[0] == 24); }
```

fastbodin: Odd that this test is found here and no where else.

dwave/optimization/src/nodes/reduce.cpp
```diff
@@ -22,996 +22,739 @@
 #include <tuple>
 #include <type_traits>
 
+#include "../functional_.hpp"
 #include "_state.hpp"
```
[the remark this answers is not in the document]
arcondello: There were also a few missing. Fixed.

dwave/optimization/src/nodes/reduce.cpp
arcondello: A test was added in fcc4ed0

dwave/optimization/src/nodes/reduce.cpp
arcondello: I'll add a test at the Python level

tests/cpp/nodes/test_reduce.cpp
arcondello: A lot of these tests are old and I see no harm in leaving a few odd ones :shrug:

dwave/optimization/src/nodes/reduce.cpp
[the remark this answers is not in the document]
wbernoudy: I would think the current order of add then remove would be more likely to be pathological (given that elements are more likely be like the others). But certainly both ways can be. And I see your second point, I thought the failures would be somewhat common for some nodes. Let's leave it as is

wbernoudy (approved):

Issue #390, opened 2025-09-17
Consider making the `ARange` symbol aware of the `Size` symbol


arcondello:
It is useful to make constructions such as
```python
from dwave.optimization import Model, arange

model = Model()

s = model.set(10)

size_of_s = s.size()  # Size symbol

a = arange(0, size_of_s)
```
in this case by construction the size of `a` is the same size as `s`, however the symbols are not smart enough to know that.

It should be relatively straightforward to make `ARange` (and maybe other symbols such as `Resize`) "aware" that then they have a `Size` symbol as a predecessor that they should calculate their size accordingly.

Blocked by https://github.com/dwavesystems/dwave-optimization/issues/381

fastbodin:
Blocked by #409

PR #410, opened 2025-10-22
Fix how ARangeNode calculates sizeinfo.min / .max

fastbodin:
Close #409

PR #411, opened 2025-10-23
Make ARangeNode aware of SizeNode

fastbodin:
Fix https://github.com/dwavesystems/dwave-optimization/issues/390. Should be merged after https://github.com/dwavesystems/dwave-optimization/pull/410.

### DIFF
--- a/releasenotes/notes/make-arange-sizenode-aware-c07e238b7c3e870f.yaml
+++ b/releasenotes/notes/make-arange-sizenode-aware-c07e238b7c3e870f.yaml
@@ -1,4 +1,4 @@
 ---
 features:
   - |
-    Make ARange symbol aware of the Size symbol 
+    Make ``ARange`` symbol aware of the ``Size`` symbol. See `#390 <https://github.com/dwavesystems/dwave-optimization/issues/390>`_.


dwave/optimization/src/nodes/creation.cpp
```diff
@@ -168,6 +174,77 @@ std::pair<double, double> calculate_values_minmax(array_or_int start_, array_or_
     unreachable();
 }
 
+const SizeInfo calculate_arange_sizeinfo(const ArrayNode* node_ptr, const array_or_int start,
+                                         const array_or_int stop, const array_or_int step) {
```
fastbodin: I could include something like `const bool exactly_one_predecessor_and_defines_stop = false` as an argument here and change the following constructor from
```cpp
ARangeNode::ARangeNode(ssize_t start, ArrayNode* stop, ssize_t step): 
           ... 
          sizeinfo_(calculate_arange_sizeinfo(this, start, stop, step)) {
    add_predecessor(stop);
}
```
to
```cpp
ARangeNode::ARangeNode(ssize_t start, ArrayNode* stop, ssize_t step): 
           ... 
          sizeinfo_(calculate_arange_sizeinfo(this, start, stop, step, **true**)) {
    add_predecessor(stop);
}
```
and change
```cpp
    if (std::holds_alternative<const Array*>(start) || std::holds_alternative<ssize_t>(stop) ||
        std::holds_alternative<const Array*>(step) ||
        dynamic_cast<const SizeNode*>(std::get<const Array*>(stop)) == nullptr) {
        return SizeInfo(node_ptr, min, max);
    }
```
to
```cpp
    if (dynamic_cast<const SizeNode*>(std::get<const Array*>(stop)) == nullptr) {
        return SizeInfo(node_ptr, min, max);
    }
```
but I felt apprehensive about making the logic not all self-contained.

releasenotes/notes/make-arange-sizenode-aware-c07e238b7c3e870f.yaml
```diff
@@ -0,0 +1,4 @@
+---
+features:
+  - |
+    Make ARange symbol aware of the Size symbol
```
arcondello:
```suggestion
    Make ``ARange`` symbol aware of the ``Size`` symbol. See `#390 <https://github.com/dwavesystems/dwave-optimization/issues/390>`_.
```

PR #412, opened 2025-10-27
Add new private `functional_.hpp`

arcondello:
Broke these out of https://github.com/dwavesystems/dwave-optimization/pull/402 for easier review.

In the long run, this is intended to replace the current public `functional.hpp`, but for now we'd like to get some experience with the methods before we make a public (stable) API.

These are meant to be analogous to NumPy's universal function (ufunc) notion.

### DIFF
--- a/dwave/optimization/src/functional_.hpp
+++ b/dwave/optimization/src/functional_.hpp
@@ -1,0 +1,509 @@
+// Copyright 2025 D-Wave
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+/// These are intended to replace the current public functional.hpp but need
+/// more thought. So for now we keep them private.
+/// These classes are obviously heavily inspired by NumPy's universal functions
+/// (ufunc). With the key difference that we wish to do operations like removing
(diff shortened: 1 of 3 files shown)


dwave/optimization/src/functional_.hpp
```diff
@@ -0,0 +1,496 @@
...
+                                              const DType auto& rhs) noexcept {
+        if (lhs > rhs) return lhs;  // We're removing a value smaller than our current max
+        return {};                  // Otherwise its ambiguous or undefined
+    }
+
+    /// @copydoc Add::result_bounds
+    static ValuesInfo result_bounds(ValuesInfo lhs, ValuesInfo rhs) {
+        return ValuesInfo(std::max(lhs.min, rhs.min),  //
+                          std::max(lhs.max, rhs.max),  //
+                          lhs.integral and rhs.integral);
```
fastbodin: Recalling the docs of `Add::result_bounds`:
```cpp
/// For `ValuesInfo result_bounds(ValuesInfo lhs, ValuesInfo rhs)` return
/// the bounds/integrality that would result by applying the operation to
/// elements with the given `lhs` and `rhs` bounds.
```
To clarify my understanding with valuesinfo, the line
```cpp
lhs.integral and rhs.integral
```
should NOT be replaced with something like
```cpp
(lhs.max > rhs.max) ? lhs.integral : rhs:integral
```
because the max being taken needs to work for all values found in `lhs` and `rhs`?

dwave/optimization/src/functional_.hpp
```diff
@@ -0,0 +1,496 @@
...
+        if (n > 1) {
+            candidates.emplace_back(low * std::pow(high, n - 1));
+            candidates.emplace_back(high * std::pow(low, n - 1));
+        }
+
+        return ValuesInfo(std::ranges::min(candidates), std::ranges::max(candidates),
+                          bounds.integral);
+    }
+    static ValuesInfo result_bounds(ValuesInfo bounds, limit_type) {
+        assert(false && "not yet implemeted");
```
arcondello: Oh, oops, this totally should be implemented and tested

PR #414, opened 2025-10-31
Switch from requirements.txt to PEP735 dependency groups

arcondello:
See [PEP 735](https://peps.python.org/pep-0735/)

The main downside is it does require `pip>=25.1`


randomir (approved):
LGTM

PR #416, opened 2025-11-04
Improve `ReduceNode::propagate()`

fastbodin:
Replace sequential calls to `unravel_index()`, `drop_axes()`, and `ravel_multi_index()` by a single more efficient method `convert_predecessor_index_()`.


releasenotes/notes/improve_reducenode_propagate-032756e0488d649f.yaml
```diff
@@ -0,0 +1,6 @@
+---
+features:
```
arcondello: In this case, because there hasn't been a deploy yet that includes the new `ReduceNode`, we don't need a release note.

PR #419, opened 2025-11-04
Non-functional improvement to `MeanNode`

fastbodin:
The way I wrote it previously annoyed me. The rest of the changes are just clang-format


fastbodin:
Removed the release note because this is a non-functional change.

Issue #429, opened 2025-11-17
Incorrect shape broadcast by `BinaryOpNode` when broadcasting


arcondello:
`BinaryOpNode` supports broadcasting of not just scalars, but any length-1 array. However this leads to some errors when using high demensional length one arrays. E.g.,

```python
model = Model()
lhs = model.set(10)
rhs = model.constant([[[0]]])
expr = lhs <= rhs

print(expr.shape())

model.states.resize(1)
with model.lock():
    lhs.set_state(0, [0, 1, 2])
    print(expr.state().shape)
```
```
(1, 1, -1)
(3, 3, 129)
```
neither of which are sensible.

Further, because we [don't store the shape on the data class](https://github.com/dwavesystems/dwave-optimization/blob/9c2a44fc85541c07147761164cf39cd56d3377cf/dwave/optimization/src/nodes/binaryop.cpp#L366-L375) we cannot even express the shape correctly in the above example.

PR #433, opened 2025-11-21
Fix shape when broadcasting with BinaryOpNode

arcondello:
Closes https://github.com/dwavesystems/dwave-optimization/issues/429

### DIFF
--- a/dwave/optimization/src/nodes/binaryop.cpp
+++ b/dwave/optimization/src/nodes/binaryop.cpp
@@ -367,25 +367,40 @@ template <class BinaryOp>
 std::span<const ssize_t> BinaryOpNode<BinaryOp>::shape(const State& state) const {
     if (!this->dynamic()) return this->shape();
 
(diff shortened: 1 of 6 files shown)


dwave/optimization/src/nodes/binaryop.cpp
```diff
@@ -367,25 +367,40 @@ template <class BinaryOp>
 std::span<const ssize_t> BinaryOpNode<BinaryOp>::shape(const State& state) const {
     if (!this->dynamic()) return this->shape();
 
-    const ssize_t lhs_size = operands_[0]->size(state);
+    const auto [lhs, rhs] = operands_;
 
-    if (lhs_size == operands_[1]->size(state)) return operands_[0]->shape(state);
```
arcondello: To explain, this line meant that if say you has a `set(5)` broadcasting with a `0`, they would have the same size and we therefore would return the shape of the rhs, which in that example would be the shape of `0`, i.e. empty.

PR #434, opened 2025-11-21
Add MatMul symbol

wbernoudy:
<s>(still working on the symbol part)</s>

`MatrixMultiply` symbol now handles vector-vector, matrix-vector, and ndarray-vector multiplication. The `matmul` method will also add in `Broadcast` symbols to one or both operands to handle the rest of the implicit broadcasting behavior that `np.matmul` supports.

### DIFF
--- a/dwave/optimization/mathematical.py
+++ b/dwave/optimization/mathematical.py
@@ -1072,6 +1072,13 @@ def matmul(x: ArraySymbol, y: ArraySymbol) -> MatrixMultiply:
             respectively. If both are 1-d, this will produce a scalar (and the
             operation is equivalent to the dot product of two vectors).
 
+            Otherwise, if it is possible that the shapes can be broadcast
+            together, either by broadcasting missing leading axes (e.g.
+            `(2, 5, 3, 4, 2)` and `(2, 6)` -> `(2, 5, 3, 4, 6)`) or by
+            broadcasting axes for which one of the operands has size 1 (e.g.
+            `(3, 1, 4, 2)` and `(1, 7, 2, 3)` -> `(3, 7, 4, 3)`), then this
+            will return the result after broadcasting.
+
     Returns:
         A MatrixMultiply symbol representing the matrix product. If `x` and
         `y` have shapes `(..., n, k)` and `(..., k, m)`, then the output will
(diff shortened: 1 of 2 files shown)


dwave/optimization/include/dwave-optimization/nodes/linear_algebra.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+    /// @copydoc Node::propagate()
+    void propagate(State& state) const override;
+
+    /// @copydoc Node::revert()
+    void revert(State& state) const override;
+
+    using Array::shape;
+
+    std::span<const ssize_t> shape(const State& state) const override;
```
arcondello: missing a few `@copydoc`

dwave/optimization/src/nodes/linear_algebra.cpp
```diff
@@ -0,0 +1,294 @@
...
+}
+
+MatMulNode::MatMulNode(ArrayNode* x_ptr, ArrayNode* y_ptr)
+        : ArrayOutputMixin(output_shape(x_ptr, y_ptr)),
+          x_ptr_(x_ptr),
+          y_ptr_(y_ptr),
+          sizeinfo_(get_sizeinfo(x_ptr, y_ptr)),
+          values_info_(get_values_info(x_ptr, y_ptr)) {}
+
+class MatMulNodeData : public ArrayNodeStateData {
```
arcondello: IMO clearer to put this above the `MatMulNode` constructor

dwave/optimization/include/dwave-optimization/nodes/linear_algebra.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+    using Array::size;
+
+    ssize_t size(const State& state) const override;
+    ssize_t size_diff(const State& state) const override;
+    SizeInfo sizeinfo() const override;
+
+ private:
+    void matmul(State& state, std::span<double> out, std::span<const ssize_t> out_shape) const;
+    void update_shape(State& state) const;
```
arcondello: Maybe this should be a method on `MatMulNodeData` instead?

dwave/optimization/src/nodes/linear_algebra.cpp
```diff
@@ -0,0 +1,294 @@
...
+    if (shape.size() == 1 and not as_row) return {shape[0], 1};
+    return {shape.begin(), shape.end()};
+}
+
+MatMulNode::MatMulNode(ArrayNode* x_ptr, ArrayNode* y_ptr)
+        : ArrayOutputMixin(output_shape(x_ptr, y_ptr)),
+          x_ptr_(x_ptr),
+          y_ptr_(y_ptr),
+          sizeinfo_(get_sizeinfo(x_ptr, y_ptr)),
+          values_info_(get_values_info(x_ptr, y_ptr)) {}
```
wbernoudy: Shockingly easy to get quite far without calling `add_predecessor()` 😬

dwave/optimization/include/dwave-optimization/array.hpp
```diff
@@ -76,6 +76,31 @@ struct SizeInfo {
...
+    friend SizeInfo operator*(SizeInfo lhs, const std::integral auto rhs) {
+        lhs *= rhs;
+        return lhs;
+    }
+
+    constexpr SizeInfo& operator/=(const std::integral auto n) {
+        if (!n) throw std::invalid_argument("cannot divide by 0");
+        multiplier /= n;
+        offset /= n;
+        if (min.has_value()) min.value() /= n;
```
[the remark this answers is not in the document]
wbernoudy: I don't think so. The min/max should always be multiplies of any valid divisor here. Though I should assert that this is the case

dwave/optimization/src/nodes/linear_algebra.cpp
```diff
@@ -0,0 +1,339 @@
...
+////////////////////// MatrixMultiplyNode
+
+// Valid shapes to multiply, and the resulting shape
+// (-1, 2, 5, 3) and (-1, 2, 3, 7) -> (-1, 2, 5, 7)
+// (-1, 3) and (3) -> (-1)
+// (-1, 3) and (3, 7) -> (-1, 7)
+// (-1) and (-1) -> ()
+// (-1) and (-1, 5) -> (5)
+
+ssize_t size_from_shape(std::span<const ssize_t> shape) {
```
arcondello: https://github.com/dwavesystems/dwave-optimization/blob/d1b5555abacbb7ba5fad159261a8196fdd95dd5f/dwave/optimization/include/dwave-optimization/array.hpp#L488
I also keep forgetting it exists.

Though I think maybe you're using the fact that it will return a negative number in the case of dynamic arrays?

arcondello:
Just occurred to me, we should probably overload [`.__matmul__()`](https://docs.python.org/3/reference/datamodel.html#object.__matmul__) as well

Issue #216, opened 2025-01-28
Consider adding the ability to specify one axis bound to `BinaryVariable`


arcondello:
Very often we need constraints such as
```python
model = Model()

x = model.binary((5, 5))

model.add_constraint((x.sum(axis=0) == 1).all())
```
which specifies that each column can have exactly one `1` in it.

It would be nice to be able to make such constraints implicit. Maybe an API like
```python
x = model.binary((5, 5), subject_to=(axis, operator, rhs))
```
so the above example would be
```python
x = model.binary((5, 5), subject_to=(0, '==', 1))
```

This would also allow us to deprecate the `DisjointBitSets` and `DisjointBitSet` symbols.

PR #444, opened 2025-12-15
Reformat `IntegerNode` index-wise bound checker

fastbodin:
Moving index-wise bound checks into its own function in anticipation of a PR for #216.


fastbodin:
Appears Mac OSX CircleCI tests are failing because of scheduled maintenance. Will re-init tests post maintenance.

arcondello (approved):

PR #446, opened 2025-12-17
Add hyperbolic tangent

### LINKED ISSUE #445: Consider implementing hyperbolic tangent symbol
It would be useful to have a `tanh` symbol to go with sine, cosine, etc.

See also https://github.com/dwavesystems/dwave-optimization/pull/367

fastbodin:
Close #445.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -488,6 +488,7 @@ class UnaryOpTests(SymbolTests):
 
     def test_scalar_input(self):
         for scalar in [-5, -.5, 0, 1, 1.5]:
+
             with self.subTest(f"a = {scalar}"):
                 model = Model()
                 a = model.constant(scalar)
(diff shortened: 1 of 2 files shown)


tests/utils.py
```diff
@@ -500,7 +500,10 @@ def test_scalar_input(self):
 
                 model.lock()
                 model.states.resize(1)
-                self.assertEqual(op_a.state(0), self.op(scalar))
+                if type(op_a) == dwave.optimization.symbols.unaryop.Tanh:
```
fastbodin: Added this because I was getting the following error when running `python -m unittest`

```
FAIL: test_scalar_input (tests.test_symbols.TestTanh.test_scalar_input) [a = 1.5]
----------------------------------------------------------------------
Traceback (most recent call last):
  File "dwave-optimization/tests/utils.py", line 507, in test_scalar_input
    self.assertEqual(op_a.state(0), self.op(scalar))
AssertionError: array(0.90514825) != np.float64(0.9051482536448665)
```

PR #447, opened 2025-12-18
Add automatic broadcasting for binaryop methods

arcondello:


dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -146,6 +146,12 @@ BroadcastToNode::BroadcastToNode(ArrayNode* array_ptr, std::span<const ssize_t>
                 "number of dimensions");
     }
 
+    if (target_shape.size() and target_shape[0] < 0 and not array_ptr_->dynamic()) {
+        throw std::invalid_argument("cannot broadcast an array with a fixed shape " +
+                                    shape_to_string(array_ptr_->shape()) + " to a dynamic shape " +
+                                    shape_to_string(target_shape));
+    }
```
fastbodin: I do not believe there is a C++ test for this? Went to verify but am having issues compiling (personal problem).

dwave/optimization/mathematical.py
```diff
@@ -112,19 +118,57 @@
...
 
 
-def _op(BinaryOp: type, NaryOp: type, reduce_method: str):
+ArraySymbolLike: typing.TypeAlias = ArraySymbol | np.typing.ArrayLike
+
+
+def _binaryop(
+    binaryop: type,
+    naryop: type | None = None,
+):
```
fastbodin: This is slick 👏 🚀

dwave/optimization/mathematical.py
```diff
@@ -1108,6 +1273,7 @@ def matmul(x: ArraySymbol, y: ArraySymbol) -> MatrixMultiply:
 
     .. versionadded:: 0.6.10
     """
+    x, y = as_array_symbols(x, y)
```
fastbodin: Given that both `x` and `y` are defined to be of type `ArraySymbol` already, functionally this just checks that they have the same underlying model right? If so, this seems overkill? I think I am missing something

dwave/optimization/mathematical.py
```diff
@@ -266,6 +315,47 @@ def argsort(array: ArraySymbol) -> ArgSort:
...
+    # Find the model, if one wasn't provided
+    if model is None:
+        for array in arrays:
+            if isinstance(array, Symbol):
+                model = array.model
+                break
+        else:
+            raise TypeError("given arrays must contain at least one ArraySymbol")
+
+    # Try to convert anythin that isn't already an array symbol into a constant
```
wbernoudy:
```suggestion
    # Try to convert anything that isn't already an array symbol into a constant
```

dwave/optimization/mathematical.py
```diff
@@ -579,7 +708,24 @@ def divide(x1: ArraySymbol, x2: ArraySymbol) -> Divide:
...
+
+@_binaryop(Equal)
+def equal(x1: ArraySymbolLike, x2: ArraySymbolLike) -> Equal:
+    """Return element-wise equality between two symbols.
+
+    Args:
+        x1, x2: Input array symbol.
+
+    Returns:
+        A symbol the is the element-wise equality.
```
wbernoudy:
```suggestion
        A symbol that is the element-wise equality.
```

dwave/optimization/mathematical.py
wbernoudy: Still need to do this? Probably also worth updating the `Args` section to say they can be numpy arrays or similar

PR #449, opened 2025-12-23
Allow array symbols to have a state-dependent shape but not size

arcondello:
I did a moderately thorough search through our other nodes and I cannot find any places where this change matters. But no doubt there are a few other edge cases out there.

### DIFF
--- a/tests/cpp/nodes/test_manipulation.cpp
+++ b/tests/cpp/nodes/test_manipulation.cpp
@@ -80,6 +80,7 @@ TEST_CASE("BroadcastToNode") {
         auto b = graph.emplace_node<BroadcastToNode>(i, std::vector<ssize_t>{0, 3});
         graph.emplace_node<ArrayValidationNode>(b);
 
+        CHECK(b->size() == 0);
         CHECK_THAT(b->shape(), RangeEquals({0, 3}));
         CHECK_THAT(b->strides(), RangeEquals({0, 8}));
 


dwave/optimization/src/nodes/binaryop.cpp
```diff
@@ -277,8 +277,8 @@ void BinaryOpNode<BinaryOp>::propagate(State& state) const {
             rhs_diff_copy.assign(rhs_ptr->diff(state).begin(), rhs_ptr->diff(state).end());
             deduplicate_diff(lhs_diff_copy);
             deduplicate_diff(rhs_diff_copy);
-            lhs_diff = std::span<const Update>(lhs_diff.begin(), lhs_diff.end());
-            rhs_diff = std::span<const Update>(rhs_diff.begin(), rhs_diff.end());
+            lhs_diff = std::span<const Update>(lhs_diff_copy.begin(), lhs_diff_copy.end());
```
arcondello: I'll admit I snuck this (unrelated) fix and test in.

arcondello:
Shoot. While working on `Reduce` I found another problem. We currently propagate `size_diff()` but that's not sufficient to detect whether we changed the shape. I am not sure we require that anywhere (there is a workaround for `ReduceNode` specifically) but it does seem to be missing.

arcondello:
`AdvancedIndexingNode` avoids needing any changes by being restricted to 0d and 1d indexing (though I think we could change that pretty easily in the future given all of the QOL features we've added in the mean time).

dwave/optimization/include/dwave-optimization/nodes/manipulation.hpp
```diff
@@ -89,6 +89,8 @@ class BroadcastToNode : public ArrayNode {
     std::unique_ptr<ssize_t[]> shape_;
     std::unique_ptr<ssize_t[]> strides_;
 
+    bool contiguous_;
```
wbernoudy: Why not make this `const`?

arcondello: We have two different conflicting conventions in play in the same node, most of the data members are not `const`, but one is. I chose to go with the majority vote for consistency.

Also, in the long run I do think we want nodes to be copyable and moveable so [this advice](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c12-dont-make-data-members-const-or-references-in-a-copyable-or-movable-type) would apply. Certainly we don't currently disallow copy/move operations.

So basically the whole library is a mess about this. So for this one line I could go either way, it'll be inconsistent regardless :smile:

dwave/optimization/src/nodes/reduce.cpp
```diff
@@ -737,7 +767,7 @@ void ReduceNode<BinaryOp>::propagate(State& state) const {
         const ssize_t subspace_size = std::reduce(subspace_shape.begin(), subspace_shape.end(), 1,
                                                   std::multiplies<ssize_t>());
         for (ssize_t i = 0, stop = size_diff / subspace_size; i < stop; ++i) {
-            state_ptr->append_reduction(initial.value());
+            state_ptr->append_reduction(*initial);
```
wbernoudy: Just curious, why change this?

PR #454, opened 2026-01-05
Improve error message for dynamic `NumberNode`s.

fastbodin:
As opposed to the following error message:
```
FAILED:
  {Unknown expression after the reported line}
due to unexpected exception with message:
  vector
```


arcondello:
Which reminds me, we need a release note :smile:

PR #456, opened 2026-01-07
Moving `NumberNode` constructor from header to src

fastbodin:
Reformatting in anticipation of a PR for #216. Reduced excessive calls for `NumberNode::initialize_state()` with `std::move()`.

### DIFF
--- a/dwave/optimization/src/nodes/numbers.cpp
+++ b/dwave/optimization/src/nodes/numbers.cpp
@@ -56,7 +56,7 @@ void NumberNode::initialize_state(State& state) const {
     for (ssize_t i = 0, stop = this->size(); i < stop; ++i) {
         values.push_back(default_value(i));
     }
-    initialize_state(state, values);
+    initialize_state(state, std::move(values));
 }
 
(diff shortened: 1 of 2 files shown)


dwave/optimization/src/nodes/numbers.cpp
```diff
@@ -56,7 +56,7 @@ void NumberNode::initialize_state(State& state) const {
     for (ssize_t i = 0, stop = this->size(); i < stop; ++i) {
         values.push_back(default_value(i));
     }
-    initialize_state(state, values);
+    initialize_state(state, std::move(values));
```
fastbodin: I believe this previously was being handled by invoking the following `NumberNode` method?
```cpp
    // Initialize a state from an existing container, making a copy.
    template <std::ranges::range R>
    void initialize_state(State& state, const R& values) const {
        return initialize_state(state, std::vector<double>(values.begin(), values.end()));
    }
```
with the `std::move`, I believe this avoids this by going directly to 
```cpp
void NumberNode::initialize_state(State& state, std::vector<double>&& number_data) const {
```
without an additional copy?

PR #457, opened 2026-01-07
Reduce `SoftMaxNode` memory and reduce copying

fastbodin:
`SoftMaxNode`, `IsInNode`, and `ArgSortNode` all use the same method to store state dependent data on the node. This method originally copied a rvalue reference as opposed to accepting a rvalue reference.

The `SoftMaxNode` constructor takes a `std::vector<double>` as an argument. This vector can be reused in initialization as opposed to copying its values.

### DIFF
--- a/dwave/optimization/src/nodes/set_routines.cpp
+++ b/dwave/optimization/src/nodes/set_routines.cpp
@@ -31,7 +31,7 @@ struct IsInNodeSetData {
 };
 
 struct IsInNodeDataHelper_ {
-    IsInNodeDataHelper_(std::vector<double> element, std::vector<double> test_elements) {
+    IsInNodeDataHelper_(std::vector<double>&& element, std::vector<double>&& test_elements) {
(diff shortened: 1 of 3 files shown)


fastbodin:
For reference, the constructors of the `StateData` for each of these nodes is called [here](https://github.com/fastbodin/dwave-optimization/blob/main/dwave/optimization/include/dwave-optimization/graph.hpp#L309-L318) 

```cpp
    template <std::derived_from<NodeStateData> StateData, class... Args>
    void emplace_data_ptr(State& state, Args&&... args) const {
        const ssize_t index = topological_index();
        assert(index >= 0 && "must be topologically sorted");
        assert(state.size() > static_cast<std::size_t>(index) && "unexpected state length");
        assert(state[index] == nullptr && "already initialized state");

        state[index] = std::make_unique<StateData>(std::forward<Args&&>(args)...); // <---- Here
    }
```

PR #461, opened 2026-01-12
Narrow contract for `IntegerNode::set_value()`

fastbodin:


arcondello:
Makes sense to me, especially because this method is not exposed at the Python level https://github.com/dwavesystems/dwave-optimization/blob/main/dwave/optimization/libcpp/nodes/numbers.pxd

PR #471, opened 2026-01-30
Enhancement/Sum constraints on NumberNode

### LINKED ISSUE #216: Consider adding the ability to specify one axis bound to `BinaryVariable`

fastbodin:
Do not merge. Draft PR. ~~Only C++ has been implemented.~~ Simply checking CircleCi.

### DIFF
--- a/dwave/optimization/include/dwave-optimization/nodes/numbers.hpp
+++ b/dwave/optimization/include/dwave-optimization/nodes/numbers.hpp
@@ -144,7 +144,8 @@ class NumberNode : public ArrayOutputMixin<ArrayNode>, public DecisionNode {
     const std::vector<AxisBound>& axis_wise_bounds() const;
 
     /// Return the state-dependent sum of the values within each hyperslice
-    /// along each bound axis.
+    /// along each bound axis. The returned vector is indexed by the
+    /// bound axes in the same ordering that `axis_wise_bounds()` returns.
     const std::vector<std::vector<double>>& bound_axis_sums(State& state) const;
 
  protected:
--- a/dwave/optimization/include/dwave-optimization/nodes/numbers.hpp
+++ b/dwave/optimization/include/dwave-optimization/nodes/numbers.hpp
@@ -157,7 +157,7 @@ class NumberNode : public ArrayOutputMixin<ArrayNode>, public DecisionNode {
     /// Return the state-dependent sum of the values within each slice
     /// along each bound axis. The returned vector is indexed by the
     /// bound axes in the same ordering that `axis_wise_bounds()` returns.
-    const std::vector<std::vector<double>>& bound_axis_sums(State& state) const;
+    const std::vector<std::vector<double>>& bound_axis_sums(const State& state) const;
 
  protected:
     explicit NumberNode(std::span<const ssize_t> shape, std::vector<double> lower_bound,
--- a/dwave/optimization/include/dwave-optimization/nodes/numbers.hpp
+++ b/dwave/optimization/include/dwave-optimization/nodes/numbers.hpp
@@ -28,23 +28,24 @@ namespace dwave::optimization {
 /// A contiguous block of numbers.
 class NumberNode : public ArrayOutputMixin<ArrayNode>, public DecisionNode {
  public:
-    /// Allowable axis-wise bound operators.
-    enum BoundAxisOperator { Equal, LessEqual, GreaterEqual };
-
     /// Struct for stateless axis-wise bound information. Given an `axis`, define
     /// constraints on the sum of the values in each slice along `axis`.
     /// Constraints can be defined for ALL slices along `axis` or PER slice along
-    /// `axis`. Allowable operators are defined by `BoundAxisOperator`.
-    struct BoundAxisInfo {
+    /// `axis`. Allowable operators are defined by `Operator`.
+    struct AxisBound {
+        /// Allowable axis-wise bound operators.
+        enum class Operator { Equal, LessEqual, GreaterEqual };
+
         /// To reduce the # of `IntegerNode` and `BinaryNode` constructors, we
         /// allow only one constructor.
-        BoundAxisInfo(ssize_t axis, std::vector<BoundAxisOperator> axis_operators,
-                      std::vector<double> axis_bounds);
+        AxisBound(ssize_t axis, std::vector<Operator> axis_operators,
+                  std::vector<double> axis_bounds);
+
         /// The bound axis
         ssize_t axis;
         /// Operator for ALL axis slices (vector has length one) or operators PER
         /// slice (length of vector is equal to the number of slices).
-        std::vector<BoundAxisOperator> operators;
+        std::vector<Operator> operators;
         /// Bound for ALL axis slices (vector has length one) or bounds PER slice
         /// (length of vector is equal to the number of slices).
         std::vector<double> bounds;
@@ -53,7 +54,7 @@ class NumberNode : public ArrayOutputMixin<ArrayNode>, public DecisionNode {
         double get_bound(const ssize_t slice) const;
 
         /// Obtain the operator associated with a given slice along `axis`.
-        BoundAxisOperator get_operator(const ssize_t slice) const;
+        Operator get_operator(const ssize_t slice) const;
     };
 
     NumberNode() = delete;
@@ -140,7 +141,7 @@ class NumberNode : public ArrayOutputMixin<ArrayNode>, public DecisionNode {
     void clip_and_set_value(State& state, ssize_t index, double value) const;
 
     /// Return the stateless axis-wise bound information i.e. bound_axes_info_.
-    const std::vector<BoundAxisInfo>& axis_wise_bounds() const;
+    const std::vector<AxisBound>& axis_wise_bounds() const;
 
     /// Return the state-dependent sum of the values within each hyperslice
     /// along each bound axis.
@@ -148,8 +149,7 @@ class NumberNode : public ArrayOutputMixin<ArrayNode>, public DecisionNode {
 
  protected:
     explicit NumberNode(std::span<const ssize_t> shape, std::vector<double> lower_bound,
-                        std::vector<double> upper_bound,
-                        std::vector<BoundAxisInfo> bound_axes = {});
+                        std::vector<double> upper_bound, std::vector<AxisBound> bound_axes = {});
 
     // Return truth statement: 'value is valid in a given index'.
     virtual bool is_valid(ssize_t index, double value) const = 0;
@@ -171,7 +171,7 @@ class NumberNode : public ArrayOutputMixin<ArrayNode>, public DecisionNode {
     std::vector<double> upper_bounds_;
 
     /// Stateless information on each bound axis.
-    std::vector<BoundAxisInfo> bound_axes_info_;
+    std::vector<AxisBound> bound_axes_info_;
 };
 
 /// A contiguous block of integer numbers.
@@ -191,39 +191,39 @@ class IntegerNode : public NumberNode {
     IntegerNode(std::span<const ssize_t> shape,
                 std::optional<std::vector<double>> lower_bound = std::nullopt,
                 std::optional<std::vector<double>> upper_bound = std::nullopt,
-                std::vector<BoundAxisInfo> bound_axes = {});
+                std::vector<AxisBound> bound_axes = {});
     IntegerNode(std::initializer_list<ssize_t> shape,
                 std::optional<std::vector<double>> lower_bound = std::nullopt,
                 std::optional<std::vector<double>> upper_bound = std::nullopt,
-                std::vector<BoundAxisInfo> bound_axes = {});
+                std::vector<AxisBound> bound_axes = {});
     IntegerNode(ssize_t size, std::optional<std::vector<double>> lower_bound = std::nullopt,
                 std::optional<std::vector<double>> upper_bound = std::nullopt,
-                std::vector<BoundAxisInfo> bound_axes = {});
+                std::vector<AxisBound> bound_axes = {});
 
     IntegerNode(std::span<const ssize_t> shape, double lower_bound,
                 std::optional<std::vector<double>> upper_bound = std::nullopt,
-                std::vector<BoundAxisInfo> bound_axes = {});
+                std::vector<AxisBound> bound_axes = {});
     IntegerNode(std::initializer_list<ssize_t> shape, double lower_bound,
                 std::optional<std::vector<double>> upper_bound = std::nullopt,
-                std::vector<BoundAxisInfo> bound_axes = {});
+                std::vector<AxisBound> bound_axes = {});
     IntegerNode(ssize_t size, double lower_bound,
                 std::optional<std::vector<double>> upper_bound = std::nullopt,
-                std::vector<BoundAxisInfo> bound_axes = {});
+                std::vector<AxisBound> bound_axes = {});
 
     IntegerNode(std::span<const ssize_t> shape, std::optional<std::vector<double>> lower_bound,
-                double upper_bound, std::vector<BoundAxisInfo> bound_axes = {});
+                double upper_bound, std::vector<AxisBound> bound_axes = {});
     IntegerNode(std::initializer_list<ssize_t> shape,
                 std::optional<std::vector<double>> lower_bound, double upper_bound,
-                std::vector<BoundAxisInfo> bound_axes = {});
+                std::vector<AxisBound> bound_axes = {});
     IntegerNode(ssize_t size, std::optional<std::vector<double>> lower_bound, double upper_bound,
-                std::vector<BoundAxisInfo> bound_axes = {});
+                std::vector<AxisBound> bound_axes = {});
 
     IntegerNode(std::span<const ssize_t> shape, double lower_bound, double upper_bound,
-                std::vector<BoundAxisInfo> bound_axes = {});
+                std::vector<AxisBound> bound_axes = {});
     IntegerNode(std::initializer_list<ssize_t> shape, double lower_bound, double upper_bound,
-                std::vector<BoundAxisInfo> bound_axes = {});
+                std::vector<AxisBound> bound_axes = {});
     IntegerNode(ssize_t size, double lower_bound, double upper_bound,
-                std::vector<BoundAxisInfo> bound_axes = {});
+                std::vector<AxisBound> bound_axes = {});
 
     // Overloads needed by the Node ABC ***************************************
 
@@ -259,38 +259,38 @@ class BinaryNode : public IntegerNode {
     BinaryNode(std::span<const ssize_t> shape,
                std::optional<std::vector<double>> lower_bound = std::nullopt,
                std::optional<std::vector<double>> upper_bound = std::nullopt,
-               std::vector<BoundAxisInfo> bound_axes = {});
+               std::vector<AxisBound> bound_axes = {});
     BinaryNode(std::initializer_list<ssize_t> shape,
                std::optional<std::vector<double>> lower_bound = std::nullopt,
                std::optional<std::vector<double>> upper_bound = std::nullopt,
-               std::vector<BoundAxisInfo> bound_axes = {});
+               std::vector<AxisBound> bound_axes = {});
     BinaryNode(ssize_t size, std::optional<std::vector<double>> lower_bound = std::nullopt,
                std::optional<std::vector<double>> upper_bound = std::nullopt,
-               std::vector<BoundAxisInfo> bound_axes = {});
+               std::vector<AxisBound> bound_axes = {});
 
     BinaryNode(std::span<const ssize_t> shape, double lower_bound,
                std::optional<std::vector<double>> upper_bound = std::nullopt,
-               std::vector<BoundAxisInfo> bound_axes = {});
+               std::vector<AxisBound> bound_axes = {});
     BinaryNode(std::initializer_list<ssize_t> shape, double lower_bound,
                std::optional<std::vector<double>> upper_bound = std::nullopt,
-               std::vector<BoundAxisInfo> bound_axes = {});
+               std::vector<AxisBound> bound_axes = {});
     BinaryNode(ssize_t size, double lower_bound,
                std::optional<std::vector<double>> upper_bound = std::nullopt,
-               std::vector<BoundAxisInfo> bound_axes = {});
+               std::vector<AxisBound> bound_axes = {});
 
     BinaryNode(std::span<const ssize_t> shape, std::optional<std::vector<double>> lower_bound,
-               double upper_bound, std::vector<BoundAxisInfo> bound_axes = {});
+               double upper_bound, std::vector<AxisBound> bound_axes = {});
     BinaryNode(std::initializer_list<ssize_t> shape, std::optional<std::vector<double>> lower_bound,
-               double upper_bound, std::vector<BoundAxisInfo> bound_axes = {});
+               double upper_bound, std::vector<AxisBound> bound_axes = {});
     BinaryNode(ssize_t size, std::optional<std::vector<double>> lower_bound, double upper_bound,
-               std::vector<BoundAxisInfo> bound_axes = {});
+               std::vector<AxisBound> bound_axes = {});
 
     BinaryNode(std::span<const ssize_t> shape, double lower_bound, double upper_bound,
-               std::vector<BoundAxisInfo> bound_axes = {});
+               std::vector<AxisBound> bound_axes = {});
     BinaryNode(std::initializer_list<ssize_t> shape, double lower_bound, double upper_bound,
-               std::vector<BoundAxisInfo> bound_axes = {});
+               std::vector<AxisBound> bound_axes = {});
     BinaryNode(ssize_t size, double lower_bound, double upper_bound,
-               std::vector<BoundAxisInfo> bound_axes = {});
+               std::vector<AxisBound> bound_axes = {});
 
     // Flip the value (0 -> 1 or 1 -> 0) at index i in the given state.
     void flip(State& state, ssize_t i) const;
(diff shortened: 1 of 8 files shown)


fastbodin:
Feature is fully implemented.

fastbodin:
Closes https://github.com/dwavesystems/dwave-optimization/issues/216.

dwave/optimization/include/dwave-optimization/nodes/numbers.hpp
```diff
@@ -28,6 +28,34 @@ namespace dwave::optimization {
 /// A contiguous block of numbers.
 class NumberNode : public ArrayOutputMixin<ArrayNode>, public DecisionNode {
  public:
+    /// Allowable axis-wise bound operators.
+    enum BoundAxisOperator { Equal, LessEqual, GreaterEqual };
```
wbernoudy: This should be `enum class`

dwave/optimization/symbols/numbers.pyx
```diff
@@ -27,25 +27,108 @@ from dwave.optimization._model cimport _Graph, _register, ArraySymbol, Symbol
...
+            raise TypeError("Bound axis must be an int.")
+
+        cpp_ops.clear()
+        if isinstance(py_ops, str):
+            # One operator defined for all slices.
+            cpp_ops.push_back(_parse_python_operator(py_ops))
+        else:
+            # Operator defined per slice.
+            ops_array = np.asarray(py_ops, order='C')
+            if (ops_array.ndim <= 1):
```
wbernoudy: I would probably just assume you have a nice iterable of strings at this point, converting to the numpy array just to check the possible other inputs and raise a special error message seems excessive to me (especially as the `bound_axes_data` argument here and in the symbol init should be explicitly type hinted).

dwave/optimization/symbols/numbers.pyx
```diff
@@ -27,25 +27,108 @@ from dwave.optimization._model cimport _Graph, _register, ArraySymbol, Symbol
...
+                for op in ops_array:
+                    # Convert op to `str` because _parse_python_operator()
+                    # does not expect a `numpy.str_`.
+                    cpp_ops.push_back(_parse_python_operator(str(op)))
+            else:
+                raise TypeError("Bound axis operator(s) should be str or"
+                                " a 1D-array of str(s).")
+
+        cpp_bounds.clear()
+        bound_array = np.asarray_chkfinite(py_bounds, dtype=np.double, order='C')
```
wbernoudy: `order='C'` is redundant here as you are not passing in the array to any other function that might rely on memory order

dwave/optimization/libcpp/nodes/numbers.pxd
```diff
@@ -19,16 +19,31 @@ from dwave.optimization.libcpp.state cimport State
...
+            # Because of this, these very explict overrides are needed per enum item.
+            Equal "dwave::optimization::NumberNode::BoundAxisOperator::Equal"
+            LessEqual "dwave::optimization::NumberNode::BoundAxisOperator::LessEqual"
+            GreaterEqual "dwave::optimization::NumberNode::BoundAxisOperator::GreaterEqual"
+
+        struct BoundAxisInfo:
+            BoundAxisInfo(Py_ssize_t axis, vector[BoundAxisOperator] axis_opertors, 
+                          vector[double] axis_bounds)
+            Py_ssize_t axis
+            vector[BoundAxisOperator] operators;
```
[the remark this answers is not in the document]
fastbodin: All I see is C++ now.

dwave/optimization/symbols/numbers.pyx
```diff
@@ -27,25 +27,108 @@ from dwave.optimization._model cimport _Graph, _register, ArraySymbol, Symbol
...
+        if (bound_array.ndim <= 1):
+            mem = bound_array.ravel()
+            cpp_bounds.reserve(mem.shape[0])
+            for i in range(mem.shape[0]):
+                cpp_bounds.push_back(mem[i])
+        else:
+            raise TypeError("Bound axis bound(s) should be scalar or 1D-array.")
+
+        output.push_back(NumberNode.BoundAxisInfo(axis, cpp_ops, cpp_bounds))
```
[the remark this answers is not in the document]
fastbodin: Agreed on the `move`. I believe I should actually be using `resize()` here since, in the future when we have more bound axes, we want `cpp_bounds` and `cpp_ops` to not copy and data from previous bound axes.

dwave/optimization/symbols/numbers.pyx
```diff
@@ -27,25 +28,103 @@ from dwave.optimization._model cimport _Graph, _register, ArraySymbol, Symbol
...
+            cpp_ops.resize(1)
+            # One operator defined for all slices.
+            cpp_ops[0] = _parse_python_operator(py_ops)
+        elif isinstance(py_ops, collections.abc.Iterable):
+            # Operator defined per slice.
+            cpp_ops.reserve(len(py_ops))
+            for op in py_ops:
+                cpp_ops.push_back(_parse_python_operator(op))
+        else:
+            raise TypeError("Bound axis operator(s) should be str or a 1D-array"
```
wbernoudy:
```suggestion
            raise TypeError("Bound axis operator(s) should be str or an iterable"
```

wbernoudy:
I do see "hyperslice" used in a lot of comments. Would prefer that is changed to just "slice" as well, or at least a small comment maybe in the header `numbers.hpp` on what that term means here.

dwave/optimization/src/nodes/numbers.cpp
```diff
@@ -15,72 +15,370 @@
...
     assert(upper_bound(j) >= ptr->get(i));
-    // Assert that i and j are valid indices occurs in ptr->exchange().
-    // Exchange occurs IFF (i != j) and (buffer[i] != buffer[j]).
-    ptr->exchange(i, j);
+    // assert() that i and j are valid indices occurs in ptr->exchange().
+    // State change occurs IFF (i != j) and (buffer[i] != buffer[j]).
+    if (ptr->exchange(i, j)) {
+        // If exchange occured, update the bound axis sums.
+        const double difference = ptr->get(i) - ptr->get(j);
+        // Index i changed from (what is now) ptr->get(j) to ptr->get(i)
```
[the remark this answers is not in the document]
wbernoudy: Perhaps could go in `update_bound_axis_slice_sums()` though maybe that would cause other uses to be incorrect, not sure.

dwave/optimization/src/nodes/numbers.cpp
```diff
@@ -242,13 +668,18 @@ bool IntegerNode::is_valid(ssize_t index, double value) const {
...
     assert(lower_bound(index) <= value);
     assert(upper_bound(index) >= value);
     assert(value == std::round(value));
-    // Assert that i is a valid index occurs in data_ptr->set().
-    // Set occurs IFF `value` != buffer[i] .
-    data_ptr<ArrayNodeStateData>(state)->set(index, value);
+    // assert() that i is a valid index occurs in ptr->set().
+    // State change occurs IFF `value` != buffer[index].
+    if (ptr->set(index, value)) {
+        // If change occured, update bound axis sums by differnce.
```
wbernoudy:
```suggestion
        // If change occurred, update bound axis sums by difference.
```

tests/test_symbols.py
```diff
@@ -725,10 +725,51 @@ def test_bounds(self):
         self.assertTrue(np.all(x.upper_bound() == [[1, 0, 0], [1, 0, 0]]))
 
         with self.assertRaises(ValueError):
-            model.integer((2, 3), upper_bound=np.nan)
+            model.binary((2, 3), upper_bound=np.nan)
```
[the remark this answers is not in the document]
wbernoudy: Oh, I missed that this is in the `BinaryVariable` tests 👍

dwave/optimization/symbols/numbers.pyx
```diff
@@ -27,25 +28,103 @@ from dwave.optimization._model cimport _Graph, _register, ArraySymbol, Symbol
...
+
 cdef class BinaryVariable(ArraySymbol):
     """Binary decision-variable symbol.
 
     See also:
         :meth:`~dwave.optimization.model.Model.binary`: equivalent method.
     """
-    def __init__(self, _Graph model, shape=None, lower_bound=None, upper_bound=None):
+    def __init__(self, _Graph model, shape=None, lower_bound=None, upper_bound=None,
+                 subject_to=None):
```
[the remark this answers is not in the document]
fastbodin: Looks like I should have been using `[]` for `tuple` typing not `()`. Will change here and elsewhere.

https://typing.python.org/en/latest/spec/tuples.html

dwave/optimization/src/nodes/numbers.cpp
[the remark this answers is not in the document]
fastbodin: I have toyed with this idea and, in fact, previously implemented it before removing it because it possibly felt like premature optimization (removed it after closing [this](https://github.com/dwavesystems/dwave-optimization/pull/470)). Happy to hear you came to the same conclusion. Will implement.

arcondello (approved):
Looks great! Just two aesthetic comments

fastbodin:
Third round of comments addressed.

wbernoudy (approved):

dwave/optimization/symbols/numbers.pyx
```diff
@@ -212,13 +324,15 @@ cdef class IntegerVariable(ArraySymbol):
...
-    def __init__(self, _Graph model, shape=None, lower_bound=None, upper_bound=None):
+    def __init__(self, _Graph model, shape=None, lower_bound=None, upper_bound=None,
+                 subject_to: None | list[tuple[int, str | list[str], float | list[float]]] = None):
         cdef vector[Py_ssize_t] cppshape = as_cppshape(
             tuple() if shape is None else shape
         )
 
         cdef optional[vector[double]] cpplower_bound = nullopt
         cdef optional[vector[double]] cppupper_bound = nullopt
+        cdef vector[BinaryNode.AxisBound] cppbound_axes = _convert_python_bound_axes(subject_to)
```
wbernoudy:
```suggestion
        cdef vector[IntegerNode.AxisBound] cppbound_axes = _convert_python_bound_axes(subject_to)
```

dwave/optimization/symbols/numbers.pyx
```diff
@@ -116,10 +203,23 @@ cdef class BinaryVariable(ArraySymbol):
...
+        try:
+            info = zf.getinfo(directory + "subject_to.json")
+        except KeyError:
+            subject_to = None
+        else:
+            with zf.open(info, "r") as f:
+                # Note that import is a list of lists, not a list of tuples.
+                # Hence we convert to tuple. We could also support lists.
+                subject_to = [(item[0], item[1], item[2]) if len(item) == 3
+                              else (item[0], item[1]) for item in json.load(f)]
```
fastbodin: Suppose that this could simply be
```python                
subject_to = [tuple(item) for item in json.load(f)]
```

wbernoudy (approved):

dwave/optimization/src/nodes/numbers.cpp
```diff
@@ -71,30 +71,121 @@ NumberNode::SumConstraint::Operator NumberNode::SumConstraint::get_operator(
...
             : ArrayNodeStateData(std::move(input)),
-              sum_constraints_lhs(std::move(sum_constraints_lhs)),
-              prior_sum_constraints_lhs(this->sum_constraints_lhs) {}
+              sum_constraints_lhs(std::move(sum_constraints_lhs)) {}
 
     std::unique_ptr<NodeStateData> copy() const override {
         return std::make_unique<NumberNodeStateData>(*this);
     }
 
+    /// Commit the state dependant data of NumberNode.
```
wbernoudy: dependent (also in a few other places)

dwave/optimization/src/nodes/numbers.cpp
```diff
@@ -71,30 +71,121 @@ NumberNode::SumConstraint::Operator NumberNode::SumConstraint::get_operator(
...
+
+void NumberNodeStateData::update(const NumberNode& node, const ssize_t index,
+                                 const double difference,
+                                 std::optional<std::vector<ssize_t>> slices) {
+    const auto& sum_constraints = node.sum_constraints();
+    assert(sum_constraints.size() != 0);  // Should only call where applicable.
+    assert(difference != 0);              // Should not call when no change occurs.
+    assert(sum_constraints.size() == sum_constraints_lhs.size());
+
+    if (slices.has_value()) {
```
wbernoudy: Seeing that you have separate implementations for the two cases of slices being provided or not, did you consider making overloads for the these methods instead of having default values for `slices`? i.e. make overloads that take `std::vector<ssize_t> slices` as the final argument.

fastbodin: Yes. I flip-flopped on that. Happy to make the change 👍

fastbodin: Fixed in https://github.com/dwavesystems/dwave-optimization/pull/471/changes/dbaa0ece247ee81c127cf83f044056b55b094b77

fastbodin:
Added custom `BinaryNode` state data to track indices (per sum constraint and per slice) whose buffer value is 1 and 0 via `DisjointSparseSet`

dwave/optimization/src/nodes/numbers.cpp
```diff
@@ -931,9 +934,435 @@ BinaryNode::BinaryNode(ssize_t size, double lower_bound, double upper_bound,
...
+
+        /// Return the `i`th index in `dense_sets_[dense_set]` whose buffer value is 1.
+        ssize_t get_ith_true_index(const ssize_t dense_id, const ssize_t i) const {
+            assert(0 <= dense_id && dense_id < static_cast<ssize_t>(dense_sets_.size()));
+            assert(0 <= i && i < num_true(dense_id));
+            return dense_sets_[dense_id][i];
+        }
+
+        /// Return the `i`th index in `dense_sets_[dense_set]` whose buffer value is 0.
+        ssize_t get_ith_false_index(const ssize_t dense_id, const ssize_t i) const {
```
wbernoudy: If I call this immediately after construction, it will return `-1` correct? Do I need to call `add_index()` for each index before I am able to use this method?

dwave/optimization/src/nodes/numbers.cpp
```diff
@@ -931,9 +934,435 @@ BinaryNode::BinaryNode(ssize_t size, double lower_bound, double upper_bound,
...
+
+    /// Given a slice (w.r.t a sum constraint), return the number of buffer
+    /// values in the slice equal to 0.
+    ssize_t num_false(const ssize_t sum_constraint_id, const ssize_t slice) const {
+        return slice_indices_[sum_constraint_id].num_false(slice);
+    }
+
+    /// Given a slice (w.r.t a sum constraint), return the `i`th index in in
+    /// the slice whose buffer value is 1.
+    ssize_t get_ith_true_index(const ssize_t sum_constraint_id, const ssize_t slice,
```
wbernoudy: I would consider making `slice_indices_` public and having `BinaryNode` call these methods directly on the `DisjointSparseSet`s instead of having more pass through methods

dwave/optimization/src/nodes/numbers.cpp
fastbodin: Correct. There are assert to catch this in `update_true()` and `update_false()`.

dwave/optimization/src/nodes/numbers.cpp
fastbodin: I went back-and-forth on that. Happy to make the change

dwave/optimization/src/nodes/numbers.cpp
```diff
@@ -931,9 +934,435 @@ BinaryNode::BinaryNode(ssize_t size, double lower_bound, double upper_bound,
...
+        for (ssize_t i = updates.size() - 1; i >= 0; --i) {
+            const Update& update = updates[i];
+            const double difference = update.value - update.old;
+            assert(difference == 1.0 || difference == -1.0);
+            const std::vector<ssize_t>& slices = slice_cache_[i];
+
+            // Reverse the change applied to each slice.
+            for (size_t j = 0, stop = slices.size(); j < stop; ++j) {
+                sum_constraints_lhs[j][slices[j]] -= difference;
+                slice_indices_[j].update(update.index, slices[j], -difference);
```
wbernoudy: The definition of `update()` and calling it with the difference here feels confusing to me, could you instead just pass in the value (update.value) into `update()` and have it take true/false? Or just
```
if (update.value) slice_indices_[j].update_true(update.index, slices[j]);
else slice_indices_[j].update_false(update.index, slices[j]);
```

fastbodin: I did it to mirror updates to `sum_constraint_lhs`. Agreed it feels awkward. Happy to change 👍

dwave/optimization/src/nodes/numbers.cpp
wbernoudy: Was there a downside to exposing them publicly?

dwave/optimization/src/nodes/numbers.cpp
wbernoudy: Would it be possible to set up the dense sets on construction so that's not possible?

dwave/optimization/src/nodes/numbers.cpp
fastbodin: I don't think so. At first I thought it would be best practice to have them as pass through methods so that they handle their own reporting internally. But then, after having done it, the number of pass throughs felt excessive 🤷 . The fact that you feel the same suggests I should change it 👍

dwave/optimization/src/nodes/numbers.cpp
fastbodin: Definitely possible. I did it this way to make `BinaryNodeStateData::compute_slice_indices_()` simple. I could hand buffer iterators per slice to each dense set.... It will take some thinking to do it nicely.

I just realized that I did not answer your second question. You do not need to call `add_index()` for each index before you are able to use `get_ith_false_index()`. When you add indices it `emplace_back()` them on their appropriate dense set.

fastbodin: The only methods that you could not use are `update_false()` and `update_true()`, since you cannot update something that has not been added.

fastbodin: Given that `num_true()`, `num_false()`, `get_ith_true_index()` and `get_ith_false_index()` all work even when all indices have not been added, I am inclined to leave it as is. An index should only ever call `update_true()` or `update_false()` when it has already been added.

wbernoudy (approved):

dwave/optimization/src/nodes/numbers.cpp
```diff
@@ -287,53 +873,505 @@ std::vector<double> limit_bound_to_bool_domain(std::optional<std::vector<double>
...
+}
+
+ssize_t BinaryNode::get_ith_true_index(const State& state, const ssize_t sum_constraint,
+                                       const ssize_t slice, const ssize_t i) const {
+    const auto& indices = data_ptr<BinaryNodeStateData>(state)->slice_indices;
+    assert(0 <= sum_constraint && sum_constraint < static_cast<ssize_t>(indices.size()));
+    assert(0 <= slice && slice < static_cast<ssize_t>(indices[sum_constraint].dense_sets.size()));
+    assert(0 <= i && i < num_true(state, sum_constraint, slice));
+    assert(i < static_cast<ssize_t>(indices[sum_constraint].dense_sets[slice].size()));
+    assert(buff(state)[indices[sum_constraint].dense_sets[slice][i]] == 1.0);
```
fastbodin, line 1359: `assert()` Added in most recent commit.

dwave/optimization/src/nodes/numbers.cpp
```diff
@@ -287,53 +873,505 @@ std::vector<double> limit_bound_to_bool_domain(std::optional<std::vector<double>
...
+
+ssize_t BinaryNode::get_ith_false_index(const State& state, const ssize_t sum_constraint,
+                                        const ssize_t slice, const ssize_t i) const {
+    const auto& indices = data_ptr<BinaryNodeStateData>(state)->slice_indices;
+    assert(0 <= sum_constraint && sum_constraint < static_cast<ssize_t>(indices.size()));
+    const ssize_t num_true = this->num_true(state, sum_constraint, slice);
+    assert(0 <= slice && slice < static_cast<ssize_t>(indices[sum_constraint].dense_sets.size()));
+    assert(0 <= i && i < num_false(state, sum_constraint, slice));
+    assert(num_true + i < static_cast<ssize_t>(indices[sum_constraint].dense_sets[slice].size()));
+    assert(buff(state)[indices[sum_constraint].dense_sets[slice][num_true + i]] == 0.0);
```
fastbodin, line 1372: `assert()` Added in most recent commit.

dwave/optimization/include/dwave-optimization/nodes/numbers.hpp
```diff
@@ -28,6 +28,55 @@ namespace dwave::optimization {
...
+        /// allow only one constructor.
+        SumConstraint(std::optional<ssize_t> axis, std::vector<Operator> operators,
+                      std::vector<double> bounds);
+
+        /// Return the axis along which slices are defined.
+        /// If `std::nullopt`, the sum constraint applies to the entire array.
+        std::optional<ssize_t> axis() const { return axis_; };
+
+        /// Obtain the bound associated with a given slice.
+        double get_bound(const ssize_t slice) const;
```
arcondello: It's an aesthetic thing, but I usually find `get_` to be redundant. I'd name this `bound()`, `operator()`, etc.

E.g., one does `vector.size()` not `vector.get_size()`

fastbodin:
Failing tests appear to be a `CircleCI` issue.

dwave/optimization/symbols/numbers.pyx
```diff
@@ -27,26 +28,120 @@ from dwave.optimization._model cimport _Graph, _register, ArraySymbol, Symbol
...
 cdef class BinaryVariable(ArraySymbol):
     """Binary decision-variable symbol.
 
     See Also:
         :meth:`~dwave.optimization.model.Model.binary`: Instantiation and
         usage of this symbol.
     """
-    def __init__(self, _Graph model, shape=None, lower_bound=None, upper_bound=None):
+    def __init__(self, _Graph model, shape=None, lower_bound=None, upper_bound=None,
+                 subject_to: None | list[tuple[str, float]] = None,
```
arcondello: We don't usually to typing on the `.pyx` files because it interacts with Cython in a weird way. Specifically, Cython will sometimes attempt to do compile-time stuff based on the typing.

dwave/optimization/symbols/numbers.pyx
```diff
@@ -144,6 +260,32 @@ cdef class BinaryVariable(ArraySymbol):
         with zf.open(directory + "upper_bound.npy", mode="w", force_zip64=True) as f:
             np.save(f, upper_bound, allow_pickle=False)
 
+        sum_constraints = self.sum_constraints()
+        if len(sum_constraints) > 0:
+            # Using json here converts the tuples to lists
+            zf.writestr(directory + "sum_constraints.json", encoder.encode(sum_constraints))
+
+    def sum_constraints(self):
```
arcondello: Should add this to [numbers.pyi](https://github.com/dwavesystems/dwave-optimization/blob/bebd6eb55a035a351e7b93bef86760c9b63d3a79/dwave/optimization/symbols/numbers.pyi). For whatever reason we don't document the constructors there so I guess we can ignore that.

arcondello (approved):
I mostly reviewed the changes to the Python interface rather than all of the low level changes. Happy to rely on @wbernoudy's review for everything else

dwave/optimization/include/dwave-optimization/nodes/numbers.hpp
arcondello: Oops, this is an old pending comment. I'll leave it here for posterity but I think it's safe to ignore.

fastbodin: I originally had `bound()` for the reasons you highlighted, but `operator()` was not allowed "'operator()' cannot be the name of a variable or data member" so I changed them both to `get_*` so, at a minimum, they were symmetric.

arcondello: In this case I would probably go with `::op()`, but your solution makes sense too.

fastbodin: I actually also prefer `::op()`. Easy find-and-replace change.

dwave/optimization/symbols/numbers.pyx
fastbodin: Interesting. I did not know! This would just become 

```pyx
    def __init__(self, _Graph model, shape=None, lower_bound=None, upper_bound=None,
                 subject_to=None, axes_subject_to=None):
   ```
then right?

arcondello: right!

dwave/optimization/symbols/numbers.pyx
fastbodin: That is because they are not private methods right? The same could be said of `lower_bounds()` and `upper_bounds()` too I believe

arcondello: Yes. You know what, let's just leave the `.pyi` alone and we can do a proper fix in another PR.

fastbodin:
Reviewer comments have been addressed in most recent 2 PRs.

PR #480, opened 2026-02-23
Cache `SizeInfo` on `TransposeNode`

fastbodin:
`TransposeNode` is now smart enough to know that it's size is simply it's predecessor's size.


fastbodin:
Should add a python test to confirm behaviour similar to https://github.com/dwavesystems/dwave-optimization/blob/main/tests/test_symbols.py#L567-L579

wbernoudy (approved):

PR #481, opened 2026-02-24
Cache `SizeInfo` on `ReduceNode`

fastbodin:
`ReduceNode` can now infer when its size is a function of its predecessor's size.

### DIFF
--- a/dwave/optimization/src/nodes/reduce.cpp
+++ b/dwave/optimization/src/nodes/reduce.cpp
@@ -24,6 +24,8 @@
 #include <utility>
 
 #include "../functional_.hpp"
+#include "dwave-optimization/array.hpp"
+#include "dwave-optimization/state.hpp"
 
 namespace dwave::optimization {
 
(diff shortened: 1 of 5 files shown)


dwave/optimization/src/nodes/reduce.cpp
```diff
@@ -581,6 +583,19 @@ ValuesInfo values_info(const Array* array_ptr, std::span<const ssize_t> axes,
     return bounds;
 }
 
+SizeInfo reducenode_calculate_sizeinfo(const Array* node_ptr, const Array* array_ptr,
+                                       std::span<const ssize_t> axes) {
+    // Node is statically sized. Note: If node_shape = {}, product(node_shape) = 1.
+    if (!node_ptr->dynamic()) return SizeInfo(product(node_ptr->shape()));
+    assert(node_ptr->shape().size() && node_ptr->shape().front() == -1);
+
+    // Node is dynamically sized but predecessor is always empty.
```
wbernoudy: I would add a bit more detail here, something like "this means we can't easily make any deductions about the resulting size as it likely depends on predecessors of the input array" (assuming that's correct).

wbernoudy (approved):

PR #482, opened 2026-02-27
Add CP code base

alexzucca90:
Basic structure for integrating domains and filtering routines.


dwave/optimization/include/dwave-optimization/cp/core/advisor.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include "dwave-optimization/cp/core/propagator.hpp"
+namespace dwave::optimization::cp {
+
+/// Utility class to help schedule propagators when variable change their domains
+/// In our settings variables and constraints are arrays, so we need a way to:
+/// - trigger propagator indices when variable indices' domains change
+/// - map variable index to propagator
+class Advisor {
+ public:
+    Advisor(Propagator* p, ssize_t p_input, std::unique_ptr<IndexTransform> index_transform);
```
wbernoudy: `p_input` is the topological index of a node?

dwave/optimization/src/cp/core/cpvar.cpp
```diff
@@ -0,0 +1,168 @@
...
+    assert(this->cp_var_index_ >= 0);
+    assert(static_cast<ssize_t>(state.var_state_.size()) > cp_var_index_);
+
+    // TODO: for now we don't handle dynamically sized nodes
+
+    ssize_t max_size, min_size;
+
+    if (node_->size() > 0) {
+        // Static array
+        max_size = node_->size();
```
wbernoudy: You could use `sizeinfo.min/max` for these and simplify the logic I think

dwave/optimization/include/dwave-optimization/cp/core/domain_listener.hpp
```diff
@@ -0,0 +1,30 @@
...
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#pragma once
+
+namespace dwave::optimization::cp {
+class DomainListener {
+ public:
+    virtual ~DomainListener() = default;
```
wbernoudy: I think if you're defining a virtual destructor like this, you should also declare the copy and move constructors to follow [the rule of three](https://en.cppreference.com/w/cpp/language/rule_of_three.html), probably as `= delete` to prevent object slicing. 

The plan is to have other types of listeners (other than `CPVar::listener`) at some point right?

dwave/optimization/include/dwave-optimization/cp/core/advisor.hpp
alexzucca90: Good catch. I added it in case some propagators need to know which variable triggered the propagator. `p_input` is which input of that propagator the variable is. I'd wait a bit more to remove it for now, just in case I need it soon.

PR #483, opened 2026-03-13
Add example of iterating over DisjointLists symbol to docstring

smharwood:
Adds an example of iterating over DisjointLists symbol to the `model.disjoint_lists_symbol` docstring, similar to what is in the `DisjointLists.set_state` docstring


wbernoudy (approved):
Considering that `DisjointLists` does not document this behavior clearly either, this was sorely needed!

PR #496, opened 2026-04-07
Make all `BufferIterator`s shaped

arcondello:
~Work in progress~

Makes all iterators shaped.

This simplifies the implementation somewhat, and means that all iterators work with `std::default_sentinel` which is both convenient and faster.

The downside of this change is that our `Array::view()` method returns a [`std::range`](https://en.cppreference.com/cpp/ranges/range), but not a [`std::common_range`](https://en.cppreference.com/cpp/ranges/common_range). That means it doesn't play nicely with a lot of the older C++ algorithm methods because the type of the beginning and end iterators are not the same. For example, [std::accumulate](https://en.cppreference.com/cpp/algorithm/accumulate) has this signature:
```c++
template< class InputIt, class T >
T accumulate( InputIt first, InputIt last, T init );
```
so one cannot do
```c++
auto view = array->view(state);
std::accumulate(view.begin(), view.end());
```
One should instead do
```c++
std::accumulate(array->begin(state), array->end(state));
```
or
```c++
auto view = array->view(state);
std::accumulate(view.begin(), view.begin() + view.size());
```
The one that's most annoying is [`std::vector::assign_range`](https://en.cppreference.com/cpp/container/vector/assign_range) is a C++23 feature.

Overall though I think it's worth it.


arcondello:
I asked Gemini to review and this is what it said
```
<some preamble>

However, since you are working with C++20/23 and seem to be prioritizing performance for systems programming, there are several architectural "weight" issues and potential edge cases to consider.

### 1. The "Heavyweight Iterator" Problem
In C++, iterators are generally expected to be "lightweight" (cheap to copy/move, often fitting in registers). 

* **Heap Allocation on Copy:** Your copy constructor and `operator=` perform a heap allocation via `std::make_unique_for_overwrite`. This is a significant performance bottleneck if this iterator is used with standard algorithms (like `std::sort` or `std::transform`) which copy iterators frequently.
* **Recommendation:** Use a **Small Buffer Optimization (SBO)** for `loc_`. Since most tensors/arrays have a rank (`ndim`) less than 8, you could use a `std::inplace_vector` (C++26) or a fixed-size buffer with a fallback to the heap.
    ```cpp
    // Example SBO approach
    ssize_t static_loc_[8]; 
    std::unique_ptr<ssize_t[]> dynamic_loc_;
    ```

### 2. Standard Library Compatibility
To fully satisfy the `std::random_access_iterator` concept (required for many `<ranges>` and `<algorithm>` features), you are missing a few type definitions and operators.

* **Missing Tags:** You should explicitly define `iterator_concept` and `iterator_category`.
    ```cpp
    using iterator_concept = std::random_access_iterator_tag;
    using iterator_category = std::random_access_iterator_tag;
    ```
* **Arrow Operator:** Even if it returns a value (proxy), `operator->` is expected for completeness.
* **Const Correctness:** In `location()`, you return a `std::span<const ssize_t>`, but the internal `unique_ptr` is `ssize_t[]`. This is fine, but ensure the method is marked `const`.

### 3. Logic & Performance in `increment_impl_`
The use of `std::div` is clever for handling N-dimensional wrapping, but it comes with a cost.

* **The Cost of `std::div`:** Integer division is one of the most expensive CPU instructions. Calling it `ndim` times for every increment is heavy.
* **Optimization:** Most iterations are contiguous or simple strides. You might consider a "fast path" for `ndim == 1` or for cases where the increment doesn't cause a wrap-around (checking if `loc[axis] + n < shape[axis]`).
* **`operator[]` Cost:** Currently, `operator[]` is $O(ndim)$ because it calls `increment_impl_`. In a standard random-access iterator, this is expected to be $O(1)$. While $O(ndim)$ is technically necessary for non-contiguous multi-dim arrays, users might be surprised by the overhead.

### 4. Safety & Edge Cases
* **Dangling Pointers:** As noted in your comments, `shape_` and `strides_` are non-owning. If the `BufferIterator` outlives the container that owns those arrays, you'll have a hard-to-debug crash. If this is a public-facing API, consider if the iterator should hold a `std::shared_ptr` or a small `std::vector` of the shape/strides.
* **The `ndim_ <= 0` Sentinel Check:** ```cpp
    friend bool operator==(const BufferIterator& lhs, std::default_sentinel_t) {
        if (lhs.ndim_ <= 0) return true;
        return lhs.loc_[0] >= lhs.shape_[0];
    }
    ```
    If `ndim_ == 0` (a scalar), this returns `true` immediately. This means a scalar iterator is **always** equal to the sentinel, making it impossible to iterate over (and dereference) a single scalar value. Usually, a scalar should allow one dereference before reaching the end.

### 5. `typing.hpp` Observations
* **Type Exhaustiveness:** In `value_()`, you switch on `FormatCharacter`. Ensure that every type allowed by the `DType` concept is covered. 
* **`long` vs `long long`:** On many systems (like Windows/MSVC), `long` and `int` are both 32-bit, while `long long` is 64-bit. On Linux, `long` is often 64-bit. Your `FormatCharacter` enum and `value_()` switch should be careful to map to the explicit `std::int64_t` types used in the concept to avoid "unreachable" code or missing cases on different compilers.

### Summary of Recommendations
1.  **Eliminate heap allocation on copy** by using a fixed-size array for `loc_` if `ndim` is small.
2.  **Add `iterator_concept`** to allow C++20 ranges to recognize it as a random-access iterator.
3.  **Fix the scalar (`ndim == 0`) logic** in the sentinel comparison to allow one-time dereferencing.
4.  **Inline the "fast path"** for `n=1` and `ndim=1` in `operator++` to avoid the `std::div` overhead for the most common use case.

Would you like me to draft a version of the `increment_impl_` that includes a fast-path for contiguous data?
```
For (1), I need to think about it. I certainly wouldn't handle the 8dim case, but I am a little tempted to handle the 1D case with a union. Something like
```c++
union Loc {
    std::make_unique<ssize_t[]> multi;
    ssize_t flat;
}
```
the code complexity is quite non-trivial though. Needs thought.

(2) seems like a no-brainer worth doing

For (3) I think it's confused and dereferencing 0d iterators works just fine. Even though it is equal to `std::default_sentinel`.

For (4) I am not sure the performance will be much different, but I'll look into it.

I'll also add `operator->`, surprised that isn't done for me.

arcondello:
After some more reading:
For (2)
* We don't satisfy `using iterator_category = std::random_access_iterator_tag;`, because we don't always return a reference type. We do satisfy `using iterator_concept = std::random_access_iterator_tag;` though. FWIW that is the default assumption if you don't add `iterator_concept`, but we might as well be explicit. I did add a test that we work with `std::sort` as expected.
* We shouldn't add `operator->` because that would return a pointer to the underlying buffer, which we cannot always do and even when we can it's of dubious value.

For (4) I don't see any performance difference. 

That leaves (1) which I think is safe to leave alone for now.

So overall we got a few more tests, a missing `const`, and a more explicit `iterator_concept` out of the review :laughing: I did learn something though!

tests/cpp/test_iterators.cpp
```diff
@@ -14,30 +14,121 @@
 
 #include <array>
 #include <cstdint>
+#include <iostream>
```
arcondello: unneeded `#include`

dwave/optimization/include/dwave-optimization/iterators.hpp
```diff
@@ -27,163 +27,164 @@ namespace dwave::optimization {
...
               format_(other.format_),
-              shape_(other.shape_ ? std::make_unique<ShapeInfo>(*other.shape_) : nullptr) {}
-
-    /// Move Constructor.
-    BufferIterator(BufferIterator&& other) = default;
+              ndim_(other.ndim_),
+              shape_(other.shape_),
+              strides_(other.strides_),
+              loc_(ndim_ > 0 ? std::make_unique_for_overwrite<ssize_t[]>(ndim_) : nullptr) {
+        assert(ndim_ >= 0);
```
fastbodin, line 73: Given it compiles fine, it appears to be finding the `assert` macro somewhere but I do not know where 🤷  Could throw a `#include <cassert>` at the top?

dwave/optimization/include/dwave-optimization/iterators.hpp
```diff
@@ -293,252 +282,180 @@ class BufferIterator {
...
-            // calculate the difference in steps based on the respective locs
-            difference_type difference = 0;
-            difference_type step = 1;
-            for (ssize_t axis = lhs.ndim - 1; axis >= 0; --axis) {
-                difference += (lhs.loc[axis] - rhs.loc[axis]) * step;
-                step *= lhs.shape[axis];
+    // This method has a confusing signature! But this way we avoid a lot of code duplication.
+    // Args:
+    //   `n` is the number of steps to increment
+    //   `ptr` is a pointer to the pointer we wish to update to point to the new location.
```
fastbodin, line 355: I would have laughed really hard if this was: "is a pointer to the pointer we wish to update to point to the new **point**"

dwave/optimization/include/dwave-optimization/iterators.hpp
```diff
@@ -293,252 +282,180 @@ class BufferIterator {
...
+            // A bit of sanity checking on our location
+            // We can go "beyond" the relevant memory on the 0th axis, but
+            // otherwise the location must be nonnegative and strictly less
+            // than the size in that dimension.
+            assert(0 <= loc[axis]);
+            assert(axis == 0 || loc[axis] < shape_[axis]);
+
+            // if we're partway through the axis, we shift to
+            // the beginning by adding the number of steps to the total
+            // that we want to go, and updating the offset accordingly
```
fastbodin, line 389: We do this because it simplifies the logic correct?

dwave/optimization/include/dwave-optimization/iterators.hpp
```diff
@@ -293,252 +282,180 @@ class BufferIterator {
...
+            if (qr.rem < 0) {
+                qr.quot -= 1;
+                qr.rem += shape_[axis];
             }
 
-            return offset;
+            // finally adjust our location and offset
+            if constexpr (UpdateLoc) loc[axis] = qr.rem;
+            offset += qr.rem * strides_[axis];
+            // qr.rem = 0;  // overwritten later, so skip resetting the .rem
```
fastbodin, line 409: Sanity check: It is overwritten on line 395 
```cpp
            qr = std::div(qr.quot, shape_[axis]);
```

dwave/optimization/include/dwave-optimization/iterators.hpp
```diff
@@ -293,252 +282,180 @@ class BufferIterator {
...
-                difference += (lhs.loc[axis] - rhs.loc[axis]) * step;
-                step *= lhs.shape[axis];
+    // This method has a confusing signature! But this way we avoid a lot of code duplication.
+    // Args:
+    //   `n` is the number of steps to increment
+    //   `ptr` is a pointer to the pointer we wish to update to point to the new location.
+    //     it is always modified.
+    //   `loc` the current location in the array as a multi-index. It is only updated if
+    //     `UpdateLoc` is `true`.
+    template <bool UpdateLoc>
```
fastbodin, line 359: I appreciate that this template enables you index and increment with this method. :nice

dwave/optimization/src/nodes/flow.cpp
```diff
@@ -128,13 +128,13 @@ void ExtractNode::propagate(State& state) const {
...
-    for (auto cit = condition.begin() + min_changed_idx, arrit = arr.begin() + min_changed_idx,
-              end = condition.end();
+    const auto end = condition.end();
+    for (auto cit = condition.begin() + min_changed_idx, arrit = arr.begin() + min_changed_idx;
          cit != end; ++cit, ++arrit) {
         if (*cit) new_values.push_back(*arrit);
     }
 
-    node_data->assign(new_values, count);
+    node_data->assign(std::move(new_values), count);
```
fastbodin, line 136: Unrelated correct? Not bothered just checking

dwave/optimization/src/nodes/flow.cpp
```diff
@@ -128,13 +128,13 @@ void ExtractNode::propagate(State& state) const {
 
     // Get the new values
     std::vector<double> new_values;
-    for (auto cit = condition.begin() + min_changed_idx, arrit = arr.begin() + min_changed_idx,
-              end = condition.end();
+    const auto end = condition.end();
+    for (auto cit = condition.begin() + min_changed_idx, arrit = arr.begin() + min_changed_idx;
          cit != end; ++cit, ++arrit) {
```
fastbodin: Can't we use `std::default_sentinel`  here in-place of `end`?

dwave/optimization/src/nodes/flow.cpp
```diff
@@ -72,14 +72,14 @@ std::span<const Update> ExtractNode::diff(const State& state) const {
...
+    const std::ranges::view auto condition = condition_ptr_->view(state);
+    const std::ranges::view auto arr = arr_ptr_->view(state);
 
     std::vector<double> values;
     values.reserve(condition.size());
 
-    for (auto cit = condition.begin(), arrit = arr.begin(), end = condition.end(); cit != end;
-         ++cit, ++arrit) {
+    const auto end = condition.end();
+    for (auto cit = condition.begin(), arrit = arr.begin(); cit != end; ++cit, ++arrit) {
```
fastbodin: Can't we use `std::default_sentinel`  here in-place of `end`?

tests/cpp/test_iterators.cpp
```diff
@@ -14,30 +14,121 @@
...
+    static_assert(std::is_nothrow_constructible<TestType>::value);
+    static_assert(std::is_nothrow_copy_constructible<TestType>::value);
+    static_assert(std::is_nothrow_move_constructible<TestType>::value);
+    static_assert(std::is_nothrow_copy_assignable<TestType>::value);
+    static_assert(std::is_nothrow_move_assignable<TestType>::value);
+
+    static_assert(std::random_access_iterator<TestType>);
+    static_assert(std::random_access_iterator<TestType>);
+    static_assert(not std::contiguous_iterator<TestType>);
+    static_assert(not std::contiguous_iterator<TestType>);
```
fastbodin, line 72: Duplicates here?

fastbodin (approved):
PR looks good to me baring a few minor comments / clarifications.

dwave/optimization/include/dwave-optimization/iterators.hpp
arcondello: Shoot, I always forget that one. I wish I could get https://github.com/include-what-you-use/include-what-you-use to work in our CI properly. Someday.

dwave/optimization/include/dwave-optimization/iterators.hpp
arcondello: yep!

dwave/optimization/src/nodes/flow.cpp
arcondello: We absolutely can and probably should

dwave/optimization/src/nodes/flow.cpp
arcondello: Yes... I think the old one would work but I was messing with the definition of `assign()` anyway. I think it's OK to leave in.

fastbodin: Agreed. Please leave in. I just wanted to make sure!

fastbodin (approved):
LGTM post revisions.

dwave/optimization/include/dwave-optimization/iterators.hpp
```diff
@@ -27,163 +28,164 @@ namespace dwave::optimization {
...
+    }
 
-    /// Construct a non-contiguous iterator from a shape/strides defined as
-    /// ranges when `From` is `void`.
-    template <DType T>
-    BufferIterator(const T* ptr,
-                   const std::span<const ssize_t> shape,    // shape of the array
-                   const std::span<const ssize_t> strides)  // strides of the array
-        requires(!DType<From>)
+    /// Convenience constructor when using shape/strides specified as ranges.
```
wbernoudy: Comment says ranges but the types are `std::span`?

arcondello: `std::span` can be constructed from a range
```c++
template< class R >
explicit(extent != std::dynamic_extent)
constexpr span( R&& r );
```
but probably should make this comment clearer

dwave/optimization/include/dwave-optimization/iterators.hpp
```diff
@@ -27,163 +28,164 @@ namespace dwave::optimization {
...
+    template <OptionalDType T>
+    BufferIterator(T* ptr, std::span<const ssize_t> shape, std::span<const ssize_t> strides)
             : BufferIterator(ptr, shape.size(), shape.data(), strides.data()) {
-        assert(shape.size() == strides.size());
+        assert(std::ranges::size(shape) == std::ranges::size(strides));
```
wbernoudy, line 131: I think you can leave this as `shape.size()` etc then?

dwave/optimization/include/dwave-optimization/iterators.hpp
```diff
@@ -207,74 +209,60 @@ class BufferIterator {
...
-        } else {
-            // both are contiguous, so they are equal if their pointers match.
-            return lhs.ptr_ == rhs.ptr_;
+        // Check that lhs and rhs are consistent with each other.
+        assert(std::ranges::equal(std::span(lhs.shape_, lhs.ndim_),
+                                  std::span(rhs.shape_, rhs.ndim_)) &&
+               "lhs must be reachable from rhs but lhs and rhs do not have the same shape");
+        assert(std::ranges::equal(std::span(lhs.strides_, lhs.ndim_),
+                                  std::span(rhs.strides_, rhs.ndim_)) &&
+               "lhs must be reachable from rhs but lhs and rhs do not have the same strides");
```
wbernoudy, line 239: Is it worth asserting the `ptr_` is the same here too?

wbernoudy (approved):

dwave/optimization/include/dwave-optimization/iterators.hpp
arcondello: That's true, OTOH this is a bit more future-proof and equivalent :shrug:

dwave/optimization/include/dwave-optimization/iterators.hpp
arcondello: In the case that they are equal? Sure, makes sense.

PR #502, opened 2026-04-09
Build with C++23

arcondello:
**WORK IN PROGRESS. DO NOT MERGE**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -115,7 +115,7 @@ jobs:
 
   python-windows:
     executor:
-      name: win/default
+      name: win/server-2025
 
     environment:
       <<: *global-environment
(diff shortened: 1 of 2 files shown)


arcondello:
Seems like we should just wait for VC2026 support in CircleCI

arcondello:
For future reference,
```
winget install --id Microsoft.VisualStudio.BuildTools --version 18.5.0 `
        --exact --silent --accept-package-agreements --accept-source-agreements `
        --override "--quiet --wait --add Microsoft.VisualStudio.Workload.VCTools --includeRecommended"

$env:CIBW_ENVIRONMENT = 'CL="/std:c++latest"'
```
maybe works?

arcondello:
Ah! Looks like there is secret support see https://discuss.circleci.com/t/latest-version-of-msvc-14-43-xxx/54161/2 and https://circleci.com/developer/machine/image/windows-server-2022-gui

Though we do need meson 1.12 for https://github.com/mesonbuild/meson/pull/15857 first. Getting there though!

arcondello:
Alas, I was wrong, we really do need a different windows machine on circleci, see https://github.com/dwavesystems/dwave-optimization/pull/594

PR #504, opened 2026-04-09
Add dynamic reduce sum propagator

alexzucca90:
Add a bound consistent sum propagator that handles dynamic input array


dwave/optimization/include/dwave-optimization/cp/core/base_node.hpp
```diff
@@ -14,12 +14,70 @@
...
 #include "dwave-optimization/cp/core/interval_array.hpp"
 #include "dwave-optimization/cp/core/propagator.hpp"
+#include "dwave-optimization/cp/propagators/binaryop.hpp"
 #include "dwave-optimization/cp/state/cpvar_state.hpp"
 #include "dwave-optimization/graph.hpp"
+#include "dwave-optimization/nodes.hpp"
+
+namespace dwave::optimization::cp {
+
+// enum class CPConversionSatus { OK, Failed };
```
wbernoudy: Did you mean to leave this commented out code in this PR?

dwave/optimization/src/cp/propagators/reduce.cpp
```diff
@@ -49,6 +51,8 @@ class SumPropagatorData : public PropagatorData {
     std::vector<double> min, max;
 };
 
+// ----- ReducePropagator -----
```
wbernoudy, line 54: Wrong section?

dwave/optimization/src/cp/propagators/reduce.cpp
```diff
@@ -144,4 +148,223 @@ CPStatus ReducePropagator<std::plus<double>>::propagate(CPPropagatorsState& p_st
...
+CPStatus DynamicReducePropagator<std::plus<double>>::propagate(CPPropagatorsState& p_state,
+                                                               CPVarsState& v_state) const {
+    auto data = data_ptr<SumPropagatorData>(p_state);
+
+    std::deque<ssize_t>& indices_to_process = data->indices_to_process();
+
+    while (data->num_indices_to_process() > 0) {
+        ssize_t i = indices_to_process.front();
+        indices_to_process.pop_front();
+        data->set_scheduled(false, i);
```
wbernoudy, line 237: I think this needs to be at the end of loop to prevent the index from getting scheduled again?

dwave/optimization/src/cp/propagators/reduce.cpp
```diff
@@ -144,4 +148,223 @@ CPStatus ReducePropagator<std::plus<double>>::propagate(CPPropagatorsState& p_st
...
+        assert(i == 0);
```
wbernoudy: Is this supposed to be here? Not sure how it's possible that the tests are passing with this haha

dwave/optimization/src/cp/propagators/reduce.cpp
```diff
@@ -144,4 +148,223 @@ CPStatus ReducePropagator<std::plus<double>>::propagate(CPPropagatorsState& p_st
...
+
+template <class BinaryOp>
+void DynamicReducePropagator<BinaryOp>::initialize_state(CPState& state) const {
+    CPPropagatorsState& p_state = state.get_propagators_state();
+    assert(propagator_index_ >= 0);
+    assert(propagator_index_ < static_cast<ssize_t>(p_state.size()));
+    p_state[propagator_index_] = std::make_unique<PropagatorData>(state.get_state_manager(), 1);
+}
+
+/// Compute the upper and lower bounds derived from the definitely present variables
```
wbernoudy, line 172: Probably worth noting for now that this is only for the `+` operator

dwave/optimization/src/cp/propagators/reduce.cpp
wbernoudy, line 241: If this code is meant to only ever work for full reduction, so output size is always scalar, I would delete the while loop.

If you want to make it work for partial reduction later, I would add a comment about that at least to this assert

dwave/optimization/src/cp/propagators/reduce.cpp
```diff
@@ -144,4 +148,223 @@ CPStatus ReducePropagator<std::plus<double>>::propagate(CPPropagatorsState& p_st
...
+        ssize_t i = indices_to_process.front();
+        indices_to_process.pop_front();
+        data->set_scheduled(false, i);
+        assert(i == 0);
+
+        /// ==== Forward propagation ====
+
+        /// Compute the output interval induced by the present variables
+        double lb_present, ub_present;
+        std::tie(lb_present, ub_present) = compute_present_bounds(in_, out_, v_state, 0, 0, 0);
```
wbernoudy:
```suggestion
        double [lb_present, ub_present] = compute_present_bounds(in_, out_, v_state, 0, 0, 0);
```

dwave/optimization/src/cp/propagators/reduce.cpp
```diff
@@ -144,4 +148,223 @@ CPStatus ReducePropagator<std::plus<double>>::propagate(CPPropagatorsState& p_st
...
+        ssize_t min_size_new = in_->max_size(v_state) + 1;
+        ssize_t max_size_new = in_->min_size(v_state) - 1;
+        ssize_t initial_min_size = in_->min_size(v_state);
+
+        double out_min = out_->min(v_state, 0);
+        double out_max = out_->max(v_state, 0);
+
+        for (ssize_t j = 0; j < static_cast<ssize_t>(lb_optional.size()); ++j) {
+            // Update the min size if the domains overlap
+            if (not(ub_optional[j] < out_min or out_max < lb_optional[j])) {
```
wbernoudy, line 290:
```suggestion
            if (not (ub_optional[j] < out_min or out_max < lb_optional[j])) {
```

dwave/optimization/src/cp/propagators/reduce.cpp
```diff
@@ -144,4 +148,223 @@ CPStatus ReducePropagator<std::plus<double>>::propagate(CPPropagatorsState& p_st
...
+            }
+
+            // Update the max size if the domains overlap
+            if (not(ub_acc[j] < out_min or out_max < lb_acc[j])) {
+                max_size_new = std::max(in_->min_size(v_state) + j, max_size_new);
+            }
+        }
+
+        assert(min_size_new >= in_->min_size(v_state));
+        bool changed_min_size = (min_size_new > in_->min_size(v_state)) ? true : false;
```
wbernoudy:
```suggestion
        bool changed_min_size = min_size_new > in_->min_size(v_state);
```

dwave/optimization/src/cp/propagators/reduce.cpp
```diff
@@ -144,4 +148,223 @@ CPStatus ReducePropagator<std::plus<double>>::propagate(CPPropagatorsState& p_st
...
+            ub_acc.clear();
+            compute_bounds_for_size(in_, out_, v_state, lb_optional, ub_optional, lb_acc, ub_acc,
+                                    lb_present, ub_present);
+        }
+
+        // Prune the present variables
+        // first compute the widest possible sum that there can be gven present and optional
+        // variables
+        {
+            double sum_max = *std::max_element(ub_optional.begin(), ub_optional.end());
```
wbernoudy, line 328: Can this fail if `changed_min_size == in_->max_size(state)`?

dwave/optimization/src/cp/propagators/reduce.cpp
```diff
@@ -144,4 +148,223 @@ CPStatus ReducePropagator<std::plus<double>>::propagate(CPPropagatorsState& p_st
...
+
+        // Prune the optional variables, they receive the bounds from as if they were present and
+        // the following optional will contribute with either present or not, whichever yields
+        // bigger gap
+        {
+            for (ssize_t j = 0; j < in_->max_size(v_state) - in_->min_size(v_state); ++j) {
+                assert(not in_->is_active(v_state, in_->min_size(v_state) + j));
+                assert(in_->maybe_active(v_state, in_->min_size(v_state) + j));
+
+                auto smin_it = std::min_element(lb_optional.begin() + j + 1, lb_optional.end());
```
wbernoudy, line 352: Might be worth pre-computing these outside the loop to avoid the O((max_size - min_size)^2) cost here?

tests/cpp/cp/propagators/test_reduce.cpp
```diff
@@ -124,4 +124,129 @@ TEST_CASE("SumPropagator") {
         }
     }
 }
+
+TEST_CASE("DynamicSumPropagator") {
+    GIVEN("A Graph with a list node and its sum") {
+        dwave::optimization::Graph graph;
+
+        // Add an integer node
+        ListNode* x = graph.emplace_node<dwave::optimization::ListNode>(5, 0, 5);
```
wbernoudy, line 133: Possibly worth another test case that allows negative values to ensure the minimums are working properly as well...

dwave/optimization/include/dwave-optimization/cp/core/base_node.hpp
alexzucca90: I think I want to remove this file now :)

dwave/optimization/src/cp/propagators/reduce.cpp
alexzucca90: That should the correct one, below you can find the `ReducePropagator` methods implementation

dwave/optimization/src/cp/propagators/reduce.cpp
alexzucca90: I think it should happen there. Say I prune an adjacent variable, then modifying that variable could trigger another propagation. If I unschedule this index, then I'm preventing this extra propagation. 
There is a caveat in all this. Can we guarantee that all the pruning that is done here, won't need another propagation?

dwave/optimization/src/cp/propagators/reduce.cpp
alexzucca90: believe it or not, clang formatting goes back to 
```cpp
if (not(ub_optional[j] < out_min or out_max < lb_optional[j])) {
```

dwave/optimization/src/cp/propagators/reduce.cpp
alexzucca90: this doesn't compile on my laptop

dwave/optimization/src/cp/propagators/reduce.cpp
wbernoudy: I don't believe it 😠 come on clang-format!!

arcondello: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#spaceafterlogicalnot

PR #508, opened 2026-04-14
remove unsupported constraint labels add_constraint examples in docs/implicit_constraints.rst

m-zarrin:
**Description**

Remove unsupported `label` arguments from `add_constraint` examples in `docs/implicit_constraints.rst` to keep the documentation aligned with the current API.


arcondello:
We never supported a label argument. I should have caught that on review.

arcondello:
@JoelPasvolsky we should add these examples to the doctests.

PR #509, opened 2026-04-14
Refresh docstrings for math functions

JoelPasvolsky:
Will also need a PR to ``dwave-ocean-sdk`` repo's ``conf.py`` file line [intersphinx_mapping = {](https://github.com/dwavesystems/dwave-ocean-sdk/blob/b6cbcacc1c1f36bd8781d728396cb297eddfd75b/docs/conf.py#L267)

### DIFF
--- a/dwave/optimization/mathematical.py
+++ b/dwave/optimization/mathematical.py
@@ -98,6 +98,7 @@ __all__ = [
     "isin",
     "less_equal",
     "linprog",
+    "LPResult",
     "log",
     "logical",
     "logical_and",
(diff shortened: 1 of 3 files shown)


dwave/optimization/mathematical.py
```diff
@@ -98,6 +98,7 @@
     "isin",
     "less_equal",
     "linprog",
+    "LPResult",
```
arcondello: This was deliberately omitted. We don't need/want this class in other namespaces.

dwave/optimization/mathematical.py
```diff
@@ -473,19 +554,24 @@ def atleast_2d(*arrays):
 
 
 def broadcast_shapes(*shapes: int | tuple[int, ...]) -> tuple[int, ...]:
-    """Broadcast the input shapes into a single shape.
+    """Calculate the broadcast shape.
+
+    .. note:: This is a helper function that does not add a symbol to the
```
arcondello: IMO this is a detail that we should not include. Not least because it might change at any time, and some methods do or don't add a symbol in different cases.

arcondello (approved):

PR #515, opened 2026-04-20
Add sparse set domains

alexzucca90:


arcondello:
Seems like we'll want a separate container-like `sparse_set` class that gets used by this class in the long run? Doesn't need to be now but want to understand the vision. Or, more concretely, should we merge this and then should I work on a container version?

dwave/optimization/include/dwave-optimization/cp/core/sparse_set_array.hpp
```diff
@@ -0,0 +1,78 @@
...
+#include "dwave-optimization/cp/core/domain_array.hpp"
+#include "dwave-optimization/cp/state/state.hpp"
+#include "dwave-optimization/cp/state/state_manager.hpp"
+
+namespace dwave::optimization::cp {
+
+class SparseSetArray : public DomainArray {
+ public:
+    SparseSetArray(StateManager* sm, ssize_t min_size, ssize_t max_size, ssize_t lb, ssize_t ub);
+    //  SparseSetArray(StateManager* sm, ssize_t size, ssize_t lb, ssize_t ub);
```
arcondello: lots of commented out functions, here and elsewhere.

dwave/optimization/include/dwave-optimization/cp/core/sparse_set_array.hpp
```diff
@@ -0,0 +1,78 @@
...
+    CPStatus remove(double value, int index, DomainListener* l) override;
+    CPStatus remove_above(double value, int index, DomainListener* l) override;
+    CPStatus remove_below(double value, int index, DomainListener* l) override;
+    CPStatus remove_all_but(double value, int index, DomainListener* l) override;
+    CPStatus update_min_size(int new_min_size, DomainListener* l) override;
+    CPStatus update_max_size(int new_max_size, DomainListener* l) override;
+
+ private:
+    std::vector<ssize_t> n_;
+    std::vector<ssize_t> ofs_;
```
arcondello: This is quite terse. Could we make these clearer or add a comment explaining what they are?

arcondello (approved):
Probably needs an overall documentation pass, but I think this is fine to merge into the development branch

PR #516, opened 2026-04-21
Improve `ReduceNode::prepare_diff()`

fastbodin:
Implemented `SparseFlags` (specialized sparse set) to track which indices have been marked as `invalid`, `updated`, or `unchanged`. Increased memory usage but should speed-up `ReduceNode::prepare_diff()` from `O(node size)` to `O(# of updates)`.

### DIFF
--- a/dwave/optimization/src/nodes/reduce.cpp
+++ b/dwave/optimization/src/nodes/reduce.cpp
@@ -38,46 +38,46 @@ class ReduceNodeData : public NodeStateData {
 
     // A modified `SparseSet` that sorts indices of a `ReduceNode` according
     // to the flags: invalid, updated, and unchanged.
-    struct SparseFlags {
+    struct ReductionFlags {
      public:
         /// All flags are set to "unchanged" by default.
-        SparseFlags(const ssize_t size) {
-            dense_flags.reserve(size);
-            sparse_flags.reserve(size);
+        ReductionFlags(const ssize_t size) {
+            indices_.reserve(size);
+            look_up_table_.reserve(size);
             for (ssize_t i = 0; i < size; ++i) {
-                dense_flags.emplace_back(i);
-                sparse_flags.emplace_back(i);
+                indices_.emplace_back(i);
+                look_up_table_.emplace_back(i);
             }
         }
 
         /// Grow by one. Note, index is set as "unchanged" by default.
         void append_back() {
-            const ssize_t index = sparse_flags.size();
-            sparse_flags.emplace_back(index);
-            dense_flags.emplace_back(index);
-            assert(dense_flags.size() == sparse_flags.size());
+            const ssize_t index = look_up_table_.size();
+            look_up_table_.emplace_back(index);
+            indices_.emplace_back(index);
+            assert(indices_.size() == look_up_table_.size());
         }
 
         /// Pop the flag cached by the last index.
         void pop_back() {
             assert(size() > 0);
-            const ssize_t index = sparse_flags.size() - 1;
+            const ssize_t index = look_up_table_.size() - 1;
             // Currently "invalid", move to "updated" region.
             if (is_invalid(index)) move_invalid_to_updated_(index);
             // Swap into last position of "unchanged" region from "updated" region.
             if (is_updated(index)) {
-                assert(num_updated > 0);
-                --num_updated;
-                swap_positions_(sparse_flags[index], size() - 1);
+                assert(num_updated_ > 0);
+                --num_updated_;
+                swap_positions_(look_up_table_[index], size() - 1);
             } else {
                 // Swap within "unchanged" region to the last position of "unchanged".
                 assert(is_unchanged(index));
-                swap_positions_(sparse_flags[index], size() - 1);
+                swap_positions_(look_up_table_[index], size() - 1);
             }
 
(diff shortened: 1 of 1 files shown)


dwave/optimization/src/nodes/reduce.cpp
```diff
@@ -37,16 +36,199 @@ class ReduceNodeData : public NodeStateData {
     using reduction_type = ufunc_type::reduction_type;
     using result_type = ufunc_type::result_type;
 
-    enum class ReductionFlag {
-        invalid,    // no longer holds the correct value, the whole reduction must be recalculated
-        unchanged,  // no change since the last propagation
-        updated,    // has been updated and holds the correct value
+    // A modified `SparseSet` that sorts indices of a `ReduceNode` according
+    // to the flags: invalid, updated, and unchanged.
+    struct SparseFlags {
```
arcondello: This on the `.cpp` side so it doesn't really matter, but this name is descriptive of the implementation rather than the purpose. I think `ReductionFlags` is a smaller change and is more descriptive.

dwave/optimization/src/nodes/reduce.cpp
```diff
@@ -37,16 +36,199 @@ class ReduceNodeData : public NodeStateData {
...
+        /// Indices of the `ReduceNode` sorted by "flag". `dense_flags`
+        /// is partitioned such that the index at position
+        /// [0, num_invalid) are considered "invalid", meaning they no longer holds
+        ///                  the correct value and the whole reduction must be recalculated,
+        /// [num_invalid, num_invalid + num_updated) are considered "updated",
+        ///                                          meaning they been updated and
+        ///                                          holds the correct value, and
+        /// [num_invalid + num_updated, end) are considered "unchanged", meaning
+        ///                                  they are unchanged since last propagation.
+        std::vector<ssize_t> dense_flags;
```
arcondello: Seems like all of these should be `private` because you only want the user to use the mutation methods?

dwave/optimization/src/nodes/reduce.cpp
```diff
@@ -37,16 +36,199 @@ class ReduceNodeData : public NodeStateData {
...
+
+        /// Current size.
+        size_t size() const {
+            assert(dense_flags.size() == sparse_flags.size());
+            return dense_flags.size();
+        }
+
+        /// A span of the indices whose values have been marked as "invalid" or "updated".
+        std::span<const ssize_t> changed() const {
+            return std::span<const ssize_t>(dense_flags.data(), num_invalid + num_updated);
```
arcondello: This is fine, but I think
```c++
std::span<const ssize_t>(dense_flags).subspan(0, num_invalid + num_updated)
```
is a bit more abstract.

dwave/optimization/src/nodes/reduce.cpp
```diff
@@ -37,16 +36,199 @@ class ReduceNodeData : public NodeStateData {
...
+        }
+
+        /// Reset all flags to "unchanged".
+        void reset() {
+            num_invalid = 0;
+            num_updated = 0;
+        }
+
+        /// Current size.
+        size_t size() const {
```
arcondello: `ssize_t`. I know the `std` containers use `size_t` but [even the C++ core guidelines say they shouldn't](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#res-signed). We use `ssize_t` everywhere in `dwave-optimization`. Also avoids you needing to `static_cast` below.

arcondello: See also https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1491r0.pdf

dwave/optimization/src/nodes/reduce.cpp
fastbodin: I was just looking at that. Completely agreed.

arcondello (approved):

dwave/optimization/src/nodes/reduce.cpp
```diff
@@ -37,16 +36,199 @@ class ReduceNodeData : public NodeStateData {
...
+                }
+            }
+            assert(indices_.size() == look_up_table_.size());
+        }
+
+        /// Mark the given `index` as "invalid".
+        void mark_invalid(const ssize_t index) {
+            assert(validate_(index));
+            // Nothing to do.
+            if (is_invalid(index)) return;
```
fastbodin, line 108: Based on how we use this, this could simply be 
```cpp
assert(!is_invalid(index));  // Else why call this method.
```
I would opt to keep it as-is to be safe though 🤷

dwave/optimization/src/nodes/reduce.cpp
```diff
@@ -37,16 +36,199 @@ class ReduceNodeData : public NodeStateData {
...
+            swap_positions_(look_up_table_[index], num_invalid_);
+            ++num_invalid_;
+            --num_updated_;
+        }
+
+        /// Mark the given `index` as "updated".
+        void mark_updated(const ssize_t index) {
+            assert(validate_(index));
+            // Nothing to do.
+            if (is_updated(index)) return;
```
fastbodin, line 123: As above but 
```cpp
assert(!is_updated(index));  // Else why call this method.
```

PR #517, opened 2026-04-21
Add missing raw string flag in python docs

fastbodin:
Previously getting errors like `SyntaxWarning: invalid escape sequence '\l'` when running `python -m unittest`

### DIFF
--- a/dwave/optimization/mathematical.py
+++ b/dwave/optimization/mathematical.py
@@ -1197,7 +1197,7 @@ class LPResult:
 
 @_binaryop(LessEqual)
 def less_equal(x1: ArraySymbolLike, x2: ArraySymbolLike) -> LessEqual:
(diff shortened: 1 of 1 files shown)


fastbodin:
Failing tests appear to be a `CircleCI` issue.

arcondello (approved):

arcondello:
Which Python version? Wonder if we can detect the error in CI

fastbodin:
Python 3.13.11

arcondello:
Hmm, [we test with `-Werror`](https://github.com/dwavesystems/dwave-optimization/blob/a75d205976da9aa54fee720f71b08729efcbea8c/pyproject.toml#L105), I am surprised this wasn't getting caught.

Oh well, simple enough change so I am not gonna lose sleep over it unless it happens again.

Issue #33, opened 2024-06-14
Consider testing `.pyi` files with `mypy.stubtest()` in CI


arcondello:
`mypy` provides [tools for stubtesting](https://mypy.readthedocs.io/en/stable/stubtest.html). We should investigate using this as part of CI if we can to make sure we keep the `.pyx` and `.pyi` synced. That said, I am not 100% sure it will play nicely with Cython. Needs investigation.

PR #522, opened 2026-04-24
Update symbols/numbers.pyi

fastbodin:
To include all public methods of `BinaryVariable` and `IntegerVariable`.

### DIFF
--- a/dwave/optimization/symbols/numbers.pyi
+++ b/dwave/optimization/symbols/numbers.pyi
@@ -12,12 +12,40 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 
-import numpy as _np
-
+import collections.abc
+import numpy as np
+import typing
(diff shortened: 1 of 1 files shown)


arcondello:
CI errors unrelated, looking into it...

fastbodin:
Unlikely the issue but I just realized I should have included the return type. Fixing now.

arcondello:
Seems like a CircleCI bug. Will let it sit for a bit before investigating further in case it clear itself.

From the broken builds:
image

from their website:
image

arcondello:
> Unlikely the issue but I just realized I should have included the return type. Fixing now.

Also the `def __init__(...)`.

Also also, https://github.com/dwavesystems/dwave-optimization/issues/33 is relevant to this PR so linking it here for future reference.

arcondello (approved):

PR #527, opened 2026-04-27
Refresh docstrings for model methods in ``_model.pyx`` file

JoelPasvolsky:
Completes remaining methods refreshed in [this previous PR](https://github.com/dwavesystems/dwave-optimization/pull/525)

### DIFF
--- a/dwave/optimization/_model.pyx
+++ b/dwave/optimization/_model.pyx
@@ -137,8 +137,7 @@ cdef class _Graph:
                 of the model to be feasible.
 
         Returns:
-            The symbol associated with the constraint; for example, a
-            :class:`~dwave.optimization.symbols.LessEqual` symbol.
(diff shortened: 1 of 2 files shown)


dwave/optimization/_model.pyx
```diff
@@ -129,12 +129,16 @@ cdef class _Graph:
...
+        For a solution to the model to be :term:`feasible <feasible state>`, the
+        :term:`constraint` must be satisfied.
+
         Args:
             value: Value that must evaluate to True for the state
                 of the model to be feasible.
 
         Returns:
-            The constraint symbol.
+            The symbol associated with the constraint; for example, a
```
arcondello: I think the "for example" here is confusing, why not use the "Example:" section for that?

arcondello (approved):

PR #529, opened 2026-04-28
Add various Stride logos for future linking

arcondello:
It's nice to have the images already present and able to be linked statically for future inclusion in the docs.


JoelPasvolsky (approved):
Thanks!

PR #533, opened 2026-04-29
Refresh docstrings for expressions and state methods

JoelPasvolsky:


dwave/optimization/expression.py
```diff
@@ -28,6 +28,40 @@
 class Expression:
     """An expression that can be used as an input to other symbols.
 
+    Examples:
```
arcondello, line 33: Another example of a class we don't want users to use directly so we don't want to document much.

dwave/optimization/states.pyx
```diff
@@ -217,22 +263,36 @@ cdef class States:
...
-        returned by the solver to which your problem model is submitted. This
-        enables asynchronous problem submission.
+        A :class:`~dwave.cloud.computation.Future` object is returned by the
+        :term:`solver` to which you submit your problem model. This enables
+        asynchronous problem submission.
+
+        This method is intended for use by developers of the Ocean SDK.
 
         Args:
-            future: ``Future`` object.
```
arcondello, line 225: We do not require a dwave-cloud-client Future, any Future will do. Though the exact definition of a Future is a bit... soft.

dwave/optimization/states.pyx
```diff
@@ -255,7 +315,41 @@ cdef class States:
                 object encoding a model. Strings are interpreted as a file
                 name.
             version:
-                A 2-tuple indicating which serialization version to use.
+                A 2-tuple indicating which serialization version to use; for
+                example, ``(1, 0)`` represents version 1.0. By default, uses the
+                latest version.
+
+        Examples:
+
```
arcondello: extra space? harmless but we don't usually add it I don't think

dwave/optimization/expression.py
JoelPasvolsky: The example is for the method, not class, and I added an explicit sentence on that now too.

PR #535, opened 2026-04-29
Change sum constraint `kwargs` for `Integer` and `Binary` symbols

fastbodin:
`BinaryNode` and `IntegerNode` kwargs underwent the following change:
`subject_to` -> `sum_subject_to`
`axes_subject_to` -> `axes_sums_subject_to`

#### Additional information
The prior kwargs were not a released feature. However, I added a release note. ~Perhaps the original release note for this feature should be changed?~ <- Previous release note made no mention of the kwargs.

#### AI Generation Disclosure
No AI tools used.


releasenotes/notes/change_sum_constraint_kwarg-b4b27cf01189d033.yaml
```diff
@@ -0,0 +1,5 @@
+---
+other:
+  - |
+    Change `subject_to` and `axes_subject_to` kwarg for `Integer` and `Binary`
```
arcondello: We haven't made a tag with the previous release note, so we should edit the old one.

fastbodin: Since the old one did not include reference to kwargs, it is left unchanged and the above release note has been removed.

arcondello (approved):

PR #540, opened 2026-05-01
Refresh clang-format

arcondello:
We still don't require users to format their files, but this PR gets a bit closer by adding `clang-format` as a dependency group.

We also format all of the existing files using the previous `.clang-format` spec. https://github.com/dwavesystems/dwave-optimization/pull/540/commits/3461074451c9f8fbf90c0479f56e1941d698ef5e has those changes, which shouldn't be interesting from a review perspective.

Finally, we update the `.clang-format` spec to change the way we format argument lists. I find this a lot clearer personally. Might be worth reviewing https://github.com/dwavesystems/dwave-optimization/commit/127ce07429de25a5e5fbe07db323145f964d6019 in isolation to see those changes.

See also https://github.com/dwavesystems/dwave-optimization/issues/113.

#### AI Generation Disclosure
I used Gemini's free version to suggest some clang-format parameters to try changing.


dwave/optimization/src/nodes/_state.hpp
```diff
@@ -67,7 +67,8 @@ class ArrayStateData {
         {
             buffer.reserve(std::ranges::size(values));
             for (ssize_t index = buffer.size(), stop = std::ranges::size(values) + offset;
-                 index < stop; ++index, ++vit) {
+                 index < stop;
+                 ++index, ++vit) {
```
smharwood: Do we want a new line before `) {` to be consistent with multi-line `if` clauses?

arcondello: I am a bit surprised that 
```
BreakBeforeCloseBracketIf: true
```
doesn't do this :thinking:

smharwood: A quick look at the docs says we need `BreakAfterOpenBracketLoop ` and `BreakBeforeCloseBracketLoop`?

arcondello: Oh, duh, will fix.

dwave/optimization/include/dwave-optimization/type_list.hpp
```diff
@@ -64,10 +64,12 @@ struct type_list {
         if constexpr (std::convertible_to<T*, Type>) return true;
 
         // Next check if we could do a dynamic down cast. Do as much at compile-time as possible
-        if constexpr (std::is_pointer_v<Type>                               //
-                      && std::is_polymorphic_v<T>                           //
-                      && std::derived_from<std::remove_pointer_t<Type>, T>  //
-                      && (std::is_const_v<std::remove_pointer_t<Type>> || !std::is_const_v<T>)) {
+        if constexpr (
+            std::is_pointer_v<Type>                               //
```
wbernoudy, line 68: Probably a pain to remove all these, but we don't need these `//` anymore now right?

dwave/optimization/src/nodes/numbers.cpp
```diff
@@ -697,14 +753,18 @@ void check_sum_constraints(const NumberNode& node) {
...
+    std::vector<double> lower_bound,
+    std::vector<double> upper_bound,
+    std::vector<SumConstraint> sum_constraints
+)
+    : ArrayOutputMixin(shape),
+      min_(get_extreme_index_wise_bound<false>(lower_bound)),
+      max_(get_extreme_index_wise_bound<true>(upper_bound)),
+      lower_bounds_(std::move(lower_bound)),
+      upper_bounds_(std::move(upper_bound)),
+      sum_constraints_(std::move(sum_constraints)) {
```
wbernoudy: Could we use `BreakConstructorInitializers: BCIS_AfterColon` to get something like 
```
SomeClass::SomeClass(
    ...
) :
    member(x),
    initializers(y),
    nicely(z),
    indented(w) {
```

wbernoudy (commented):
It is incredible what a massive boon to readability this change is.

arcondello:
> It is incredible what a massive boon to readability this change is.

Yeah, the clang-format folks really cooked with v22. Lots of nice new options.

dwave/optimization/src/nodes/numbers.cpp
arcondello: Sure, though it does have a slightly funny interaction with `requires()`
```c++
        requires(!DType<From>)
        :
        ptr_(ptr),
        format_(format_of<T>()),
        shape_(std::make_unique<ShapeInfo>(ndim, shape, strides)) {}
```
I am happy either way

arcondello: actually, I think it was already pretty ugly
```c++
        requires(!DType<From>)
        : ptr_(ptr),
          format_(format_of<T>()),
          shape_(std::make_unique<ShapeInfo>(ndim, shape, strides)) {}
```
so happy enough to make the change.

dwave/optimization/include/dwave-optimization/type_list.hpp
arcondello: Yes that's right, I'll probably leave them in for now to minimize changes.

dwave/optimization/src/nodes/numbers.cpp
arcondello: Sorry, thinking again, I see that if the initializer list is short it does
```c++
    template <DType T>
    BufferIterator(
        const T* ptr,
        const std::span<const ssize_t> shape,  // shape of the array
        const std::span<const ssize_t> strides
    )  // strides of the array
        requires(!DType<From>)
        : BufferIterator(ptr, shape.size(), shape.data(), strides.data()) {
        assert(shape.size() == strides.size());
    }
```
so I am inclined to keep the original format for consistency. I.e. the initializers "section" always begins with a `:`. But I can be talked out of it :shrug: 

https://github.com/dwavesystems/dwave-optimization/pull/540/commits/b287bb61ab2c50c74aa08600e60fcbcb67e732db shows what it looks like with the change if that's helpful.

wbernoudy: It seems that `BreakConstructorInitializers: AfterColon` has a funny interaction with `requires()` that is not consistent with how it works the rest of the time. This doesn't seem to be controllable or documented.

However, `RequiresClausePosition: SingleLine` does seem to help with the "single colon on its own line issue":
```c++
template <DType T>
BufferIterator(const T* ptr, const std::span<const ssize_t> shape)
    requires(!DType<FromAasdfasdfasdfasdfasdfasdfasdfasdfasdfasdfasdf>)
    : BufferIterator(ptr, shape.size(), shape.data(), strides.data()) {
    assert(shape.size() == strides.size());
}
```

I'm still in favor of using `AfterColon`-- the `requires()` case is not super common, and while the colon at the start of the initializers line instead the start of the previous is inconsistent, I think the consistent indentation of the initializers in all cases is worth it. And hopefully `clang-format` will introduce options to control the colon better.

...but I won't block the PR for this :)

wbernoudy (approved):

dwave/optimization/src/nodes/numbers.cpp
arcondello: Ok, implemented in https://github.com/dwavesystems/dwave-optimization/pull/540/commits/44b369a07f44e44949f0e9b0f3aef8b5f3145b33

arcondello:
Closing in favour of https://github.com/dwavesystems/dwave-optimization/pull/544

PR #541, opened 2026-05-01
Add missing includes to functional.hpp

arcondello:
#### AI Generation Disclosure
No AI used.

### DIFF
--- a/dwave/optimization/include/dwave-optimization/functional.hpp
+++ b/dwave/optimization/include/dwave-optimization/functional.hpp
@@ -14,6 +14,8 @@
 
 #pragma once
 
+#include <algorithm>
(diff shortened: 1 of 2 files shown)


dwave/optimization/include/dwave-optimization/functional.hpp
```diff
@@ -14,6 +14,8 @@
 
 #pragma once
 
+#include <algorithm>
+#include <concepts>
```
alexzucca90, line 18: is concepts needed?

arcondello: For `std::integral<>`, yes.

alexzucca90 (approved):

PR #543, opened 2026-05-05
Try to use pkgconf-pypi for blas dependency

arcondello:
Try to use https://github.com/pypackaging-native/pkgconf-pypi for scipy-openblas configuration

#### AI Generation Disclosure
No AI used.


arcondello:
Hmm, unfortunately the `pkgconf` doesn't play nicely with the lack of universal wheel for `scipy-openblas`. This means that our x86-64 MacOS build fails.

Some options off the top of my head:
* Drop the `scipy-openblas` build for MacOS x86-64. Honestly how many of those users do we still have?
* Just drop this PR entirely and keep the status quo.
* I think we could swap the order of the `delocate-merge` depending on whether we're trying to do an arm or x86-64 build. But yeah, that seems like a mess.
* Rather than merging the wheels, we might be able to install them into individual environments, dump the pkgconf files somewhere, and then append the pkgconf path.

arcondello:
Ok current version just doesn't do a BLAS build for macos x86-64.

I gotta say, with the `pkgconf-pypi` change https://github.com/pypackaging-native/pkgconf-pypi/pull/105, I am inclined to abandon this. At least until/unless `meson` or `meson-python` adds native support https://github.com/pypackaging-native/pkgconf-pypi/issues/98#issuecomment-3587835383. 

Additional discussion in https://github.com/pypackaging-native/pkgconf-pypi/issues/110

PR #550, opened 2026-05-20
Add minimal interval implementation

arcondello:
This does not add [outward rounding](https://standards.ieee.org/ieee/1788/4431/), which we definitely want/need in the long run. But I wanted to get a minimal API up.

Next step is to replace or integrate into the existing `ValuesInfo` class, maybe with something like `std::variant<interval<double>, interval<ssize_t>>`.

#### AI Generation Disclosure
No AI used.

### DIFF
--- a/dwave/optimization/include/dwave-optimization/interval.hpp
+++ b/dwave/optimization/include/dwave-optimization/interval.hpp
@@ -175,7 +175,7 @@ requires(not std::is_const_v<T>) class interval {
         return out;
     }
 
-    bool contains(DType auto&& value) const {
+    bool contains(auto&& value) const {
         // dev note: numeric issues
         return infimum_ <= value and value <= supremum_;
     }
@@ -218,7 +218,7 @@ requires(not std::is_const_v<T>) class interval {
         return 0;
     }
     std::ptrdiff_t size() const noexcept requires(std::integral<T>) {
-        if (supremum_ >= infimum_) return static_cast<std::ptrdiff_t>(supremum_) - infimum_;
+        if (supremum_ >= infimum_) return static_cast<std::ptrdiff_t>(supremum_) - infimum_ + 1;
         return 0;
     }
 
--- a/dwave/optimization/include/dwave-optimization/interval.hpp
+++ b/dwave/optimization/include/dwave-optimization/interval.hpp
@@ -1,0 +1,90 @@
+// Copyright 2026 D-Wave
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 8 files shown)


dwave/optimization/src/functional_.hpp
```diff
@@ -87,6 +90,24 @@ struct BinaryFunctionMixin {
...
+struct Abs : UnaryFunctionMixin<Abs<T>> {
+    using result_type = T;
+
+    constexpr T operator()(DType auto&& x) const noexcept {
+        return std::abs(x);
+    }
+
+    template<DType U>
+    interval<T> operator()(const interval<U>& in) const noexcept {
+        // first get the absolute value, using the type of the input interval
```
wbernoudy, line 104: First?

dwave/optimization/include/dwave-optimization/interval.hpp
```diff
@@ -0,0 +1,266 @@
...
+            supremum_ += std::ceil(rhs.supremum());
+        } else {
+            infimum_ += rhs.infimum();
+            supremum_ += rhs.supremum();
+        }
+        return *this;
+    }
+
+    template <DType U>
+    friend auto operator+(interval lhs, const interval<U>& rhs) {
```
wbernoudy, line 172: Can we overload these operators for scalars as well?

dwave/optimization/src/functional_.hpp
```diff
@@ -87,6 +90,24 @@ struct BinaryFunctionMixin {
...
+        return static_cast<interval<T>>((-in | in) & interval<U>::nonnegative());
+    }
```
wbernoudy, line 106: Partially to check my understanding
```suggestion
    }

    template<DType U>
    interval<T> inverse(const interval<U>& in) const noexcept {
      assert((interval<U>::nonnegative() | in) == in and
             "argument for inverse of abs should be nonnegative");
      return in | -in;
    }
```

dwave/optimization/include/dwave-optimization/interval.hpp
arcondello: Yup!

dwave/optimization/src/functional_.hpp
arcondello: leftover from an old implementation

dwave/optimization/src/functional_.hpp
arcondello: Yup! I think so

dwave/optimization/include/dwave-optimization/interval.hpp
```diff
@@ -0,0 +1,266 @@
...
+        }
+    }
+
+    /// The size of the interval is the max(supremum_ - infimum_, 0)
+    double size() const noexcept requires(std::floating_point<T>) {
+        if (supremum_ >= infimum_) return static_cast<double>(supremum_) - infimum_;
+        return 0;
+    }
+    std::ptrdiff_t size() const noexcept requires(std::integral<T>) {
+        if (supremum_ >= infimum_) return static_cast<std::ptrdiff_t>(supremum_) - infimum_;
```
wbernoudy:
```suggestion
        if (supremum_ >= infimum_) return static_cast<std::ptrdiff_t>(supremum_) - infimum_ + 1;
```
I believe this should be correct for all integral types

arcondello:
Once https://github.com/dwavesystems/dwave-optimization/pull/582 is merged it'll fix a lot of the function signatures.

PR #551, opened 2026-05-22
Update clang-format to organize #include statements

arcondello:
We were kind of all over the place with our `#include` statements. This PR adopts the convention that `""` includes are for internal libraries. All "third-party" `#include`s now use `<>`. I also added some categories so we get nice readable separation.

#### AI Generation Disclosure
No AI used. I found all the clang-format options via stackoverflow like god intended.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -40,7 +40,7 @@ PenaltyReturnTypeOnItsOwnLine: 1000
 IncludeCategories:
   - Regex: '<[^.>]+>'
     Priority: 1
-  - Regex: '<catch2.+\.hpp>'
+  - Regex: '<(catch2|cblas).*\.(h|hpp)>'
     Priority: 3
   - Regex: '<.+\.(h|hpp)>'
     Priority: 2
--- a/.clang-format
+++ b/.clang-format
@@ -29,4 +29,21 @@ RequiresClausePosition: SingleLine
 
 # Prevent putting the return type on its own line most of the time
 PenaltyReturnTypeOnItsOwnLine: 1000
+
+# Our header order should be
+# 1) The "main" include, i.e., the corresponding .hpp file to a .cpp file. clang-format
+#    handles that separately
+# 2) All includes with <> without an extension, i.e., the std
+# 3) All includes with <> with an extension, e.g., #include <sys/types.h>
+# 4) 3rd party includes, i.e., catch2
+# 5) All includes with ""
+IncludeCategories:
+  - Regex: '<[^.>]+>'
+    Priority: 1
+  - Regex: '<catch2.+\.hpp>'
+    Priority: 3
+  - Regex: '<.+\.(h|hpp)>'
+    Priority: 2
+  - Regex: '".+"'
+    Priority: 4
 ...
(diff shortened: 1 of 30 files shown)


arcondello:
Note that this is somewhat in disagreement with the [C++ core guidline](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf12-prefer-the-quoted-form-of-include-for-files-relative-to-the-including-file-and-the-angle-bracket-form-everywhere-else). An alternative would be to use `#include <dwave-optimization/fname.hpp>` in the tests and introduce a new category in the `.clang-format`. So e.g.,
```c++
#include <vector>

#include <catch2/catch_test_macros.hpp>
#include <catch2/matchers/catch_matchers_floating_point.hpp>

#include <dwave-optimization/nodes.hpp>

#include ".../relative_file.hpp"
```
of course, following that advice to it's logical conclusion takes us even further because we current separate the `include/` directory from `src/` directory. So in theory almost all includes should use `<>`.

edit: [Google's C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) doesn't explicitly specify but implicitly suggests using `""` for local headers and `<>` for external.
edit: LLVM Coding Standards likewise doesn't specify but seems to implicitly support the `""` local, `<>` external convention.

arcondello:
Also note that the [LLVM Coding Standards](https://llvm.org/docs/CodingStandards.html#include-style) suggest putting system headers last. So more-or-less the reverse of our current order.

edit: [Google's C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) puts system at the top. And we are using that as our "base" so I think it's fine to maintain.

.clang-format
```diff
@@ -29,4 +29,21 @@ RequiresClausePosition: SingleLine
...
+# 1) The "main" include, i.e., the corresponding .hpp file to a .cpp file. clang-format
+#    handles that separately
+# 2) All includes with <> without an extension, i.e., the std
+# 3) All includes with <> with an extension, e.g., #include <sys/types.h>
+# 4) 3rd party includes, i.e., catch2
+# 5) All includes with ""
+IncludeCategories:
+  - Regex: '<[^.>]+>'
+    Priority: 1
+  - Regex: '<catch2.+\.hpp>'
```
wbernoudy: Why do we need a special case for catch2 here?

arcondello: I want to separate third-party dependencies like `catch2` from other libraries that use `<.h>` or `<.hpp>` like `#include <sys/types.h>`. So we need to make one of them a special case. IMO it's easier to list our explicit third part dependencies than to do the inverse.

wbernoudy: Oh duh, I see now. Just forgot catch2 was the only third party dep.

wbernoudy (approved):

.clang-format
arcondello: Oh! I forgot `blas`. https://github.com/dwavesystems/dwave-optimization/blob/b2918faf8a9ad102e15dcfb0e2a98145c10f0af1/dwave/optimization/src/nodes/linear_algebra.cpp#L27 though it won't actually be affected because we include it within a guard.

Issue #510, opened 2026-04-16
Add checkpointing method(s) for decision array nodes


arcondello:
It would be useful for all decision nodes to have methods for checkpointing their state. I am imagining something like
```c++
struct StateCheckpoint {
    virtual ~StateCheckpoint() = default;
};

class DecisionNode {
    // Get an object from the node that contains the information necessary to return the node to it's current state
    virtual std::unique_ptr<StateCheckpoint> checkpoint(const State& state) const = 0;

    // This sets the current buffer/diff to match the state
    virtual void assign_checkpoint(State& state, std::unique_ptr<StateCheckpoint> checkpoint) const = 0;
};
```
```c++
class MyStateCheckpoint : public StateCheckpoint {
    // In this example we just save the buffer, but we could store whatever other info we need if the node
    // has a more complex state
    std::vector<double> buffer;
};

class MyDecisionNode {
    // Node-specific mutation method.
    void assign(State& state, std::vector<double> values) const;

    std::unique_ptr<StateCheckpoint> checkpoint(const State& state) const override {
        assert(this->diff(state).empty());  // checkpoints should be created between propagations
        std::vector<double> buffer;
        for (const double& val : this->view(state)) buffer.emplace_back(val);
        return std::make_unique<MyStateCheckpoint>(std::move(buffer));
    }

    void assign_checkpoint(State& state, std::unique_ptr<StateCheckpoint> checkpoint) const override {
        const MyStateCheckpoint* my_checkpoint = dynamic_cast<MyStateCheckpoint*>(checkpoint.get());
        this->assign(state, my_checkpoint->buffer);
    }
};
```

*Note*: if we could go further and make `void assign(State& state, std::vector<double> values) const` a required method for all decision nodes, which would allow us to make my example above the "default checkpoint".

arcondello:
Based on https://github.com/dwavesystems/dwave-optimization/pull/552 and other experiments, I think we need to move the checkpoints methods to the `Graph`. So something like
```c++
struct Graph {
    checkpoint_type checkpoint(State& state, std::span<const DecisionNode*> node_ptrs) const;
    
    // sets the value of the decisions that were checkpointed and then does a propagation/commit
    void assign_checkpoint(State& state, checkpoint_type&& checkpoint) const;
};
```
or, if we want to be able to mix and match individual node checkpoints, something like
```c++
struct Graph {
    // returns a vector of node_ptrs.size(), one for each node
[...]

PR #552, opened 2026-05-27
Add proof-of-concept for CollectionNode checkpointing

arcondello:
Proof of concept for checkpointing along the lines of https://github.com/dwavesystems/dwave-optimization/issues/510. Obviously this is a rough-cut implementation and needs a lot of testing and cleanup, but shows the basic idea.

#### AI Generation Disclosure
No AI used.

### DIFF
--- a/dwave/optimization/src/nodes/collections.cpp
+++ b/dwave/optimization/src/nodes/collections.cpp
@@ -66,6 +66,46 @@ std::vector<double> augment_collection(std::vector<double> values, const ssize_t
     return values;
 }
 
+struct CollectionStateData;
+
+struct CollectionStateCheckpoint_ : NodeStateCheckpoint {
+    // We want our checkpoint to be moveable but not copyable or default constructable
+    CollectionStateCheckpoint_() = delete;
+    CollectionStateCheckpoint_(const CollectionStateCheckpoint_&) = delete;
+    CollectionStateCheckpoint_(CollectionStateCheckpoint_&&) = default;
+    CollectionStateCheckpoint_& operator=(const CollectionStateCheckpoint_&) = delete;
+    CollectionStateCheckpoint_& operator=(CollectionStateCheckpoint_&&) = default;
+
+    ~CollectionStateCheckpoint_();
+
+    CollectionStateCheckpoint_(CollectionStateData* state_ptr) noexcept;
+
+    void emplace_updates(std::vector<Update>&& updates) {
+        this->updates.emplace_back(std::move(updates));
+    }
+    void emplace_updates(CollectionStateCheckpoint_& other) {
+        updates.insert(updates.end(), other.updates.begin(), other.updates.end());
+        // assert that we used move?
+    }
+
+    void reset() {
+        updates.clear();
+        state_ptr = nullptr;
+    }
+
+    bool valid() override { return state_ptr != nullptr; }
+
+    CollectionStateData* state_ptr;
+
+    // the size of the array at the time the checkpoint was made
+    const ssize_t size;
+
+    // all of the updates since the checkpoint was made (until the next checkpoint)
+    // we could keep them as a flat vector of updates but this way the move semantics
+    // are fast
+    std::vector<std::vector<Update>> updates;
+};
+
 struct CollectionStateData : NodeStateData {
     explicit CollectionStateData(ssize_t n) : CollectionStateData(n, n) {}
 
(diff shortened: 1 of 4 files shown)


dwave/optimization/src/nodes/collections.cpp
```diff
@@ -66,6 +66,46 @@ std::vector<double> augment_collection(std::vector<double> values, const ssize_t
     return values;
 }
 
+struct CollectionStateData;
+
+struct CollectionStateCheckpoint_ : NodeStateCheckpoint {
+    // We want our checkpoint to be moveable but not copyable or default constructable
```
arcondello, line 72: Lies! We don't want to let it be moved either because we're passing around `unique_ptr`s.

dwave/optimization/src/nodes/collections.cpp
```diff
@@ -85,6 +125,14 @@ struct CollectionStateData : NodeStateData {
         assert(0 <= size && static_cast<std::size_t>(size) <= this->elements.size());
     }
 
+    ~CollectionStateData() {
+        // in case the state gets deallocated before the checkpoints we don't want
+        // a segfault
+        for (CollectionStateCheckpoint_* checkpoint_ptr : checkpoints) {
+            checkpoint_ptr->state_ptr = nullptr;
```
arcondello, line 132: Should be `->reset()`.

dwave/optimization/src/nodes/collections.cpp
```diff
@@ -111,9 +159,59 @@ struct CollectionStateData : NodeStateData {
...
+                    elements[idx] = old;
+                }
+            }
+
+            checkpoints[i]->reset();
+
+            if (checkpoints[i] == checkpoint.get()) {
+                // we found the checkpoint we were looking for!
+                if (checkpoints[i]->size != size) {
+                    assert(false and "not yet implemented");
```
wbernoudy, line 185: Why isn't this just
```c++
size = checkpoint->size;
previous_size = checkpoint->size;
```

arcondello: It is, this is a placeholder for myself until I add a test that hits this path.

dwave/optimization/src/nodes/collections.cpp
```diff
@@ -66,6 +66,46 @@ std::vector<double> augment_collection(std::vector<double> values, const ssize_t
...
 
+struct CollectionStateData;
+
+struct CollectionStateCheckpoint_ : NodeStateCheckpoint {
+    // We want our checkpoint to be moveable but not copyable or default constructable
+    CollectionStateCheckpoint_() = delete;
+    CollectionStateCheckpoint_(const CollectionStateCheckpoint_&) = delete;
+    CollectionStateCheckpoint_(CollectionStateCheckpoint_&&) = default;
+    CollectionStateCheckpoint_& operator=(const CollectionStateCheckpoint_&) = delete;
+    CollectionStateCheckpoint_& operator=(CollectionStateCheckpoint_&&) = default;
```
wbernoudy, line 77: Move assignment is implicitly deleted because of the const `size` member.

dwave/optimization/src/nodes/collections.cpp
```diff
@@ -147,6 +245,23 @@ struct CollectionStateData : NodeStateData {
...
 
+    /// Deconstruct a checkpoint without reverting the state to it.
+    void remove_checkpoint(CollectionStateCheckpoint_* checkpoint) {
+        assert(not checkpoints.empty() and "checkpoints unexpectedly empty");
+
+        for (ssize_t i = 0, num_checkpoints = checkpoints.size(); i < num_checkpoints; ++i) {
+            if (checkpoints[i] == checkpoint) {
+                if (i > 0) {
+                    checkpoints[i - 1]->emplace_updates(*checkpoint);
+                }
```
fastbodin, line 256: Sanity check: This ensures continuity between the checkpoints. i.e. no breakpoints.

**Example:** If I had checkpoints: `A` -> `B` -> `C` and deleted `B`, I would still be able to get from `A` -> `C` b.c. the updates in `B` have been absorbed into `A`?

dwave/optimization/src/nodes/collections.cpp
```diff
@@ -111,9 +159,59 @@ struct CollectionStateData : NodeStateData {
         assert(this->size == size);
     }
 
+    void assign_from_checkpoint(std::unique_ptr<CollectionStateCheckpoint_>&& checkpoint) {
```
wbernoudy, line 162: In this implementation checkpoints are cheap. But if they can be expensive, and assuming we want to be able to go back to a checkpoint and start again, then I think we need at least the option to restore without invalidating the checkpoint.

PR #556, opened 2026-05-28
Fix 1d and non-contiguous matrix multiplication

### LINKED ISSUE #553: `matmul` fails for some cases
#### Original bug report from @fwoodhouse

I'm using matmul and transpose to do a vector self outer product; that is, for a column vector V, form M = VV^T, so M_ij = V_i V_j.

In the following code, in theory xTx and yyT should be identical:
```python
SHAPE = 3

model = Model()

x = model.integer(shape=(1,SHAPE))
xTx = matmul(transpose(x), x)

y = model.integer(shape=(SHAPE,1))
yyT = matmul(y, transpose(y))

with model.lock():
    model.states.resize(1)
    x.set_state(0, [5]*SHAPE)
    y.set_state(0, [5]*SHAPE)
    print(xTx.state(0))
    print(yyT.state(0))
```
But yyT fails. The output is:
```bash
[[25. 25. 25.]
 [25. 25. 25.]
 [25. 25. 25.]]
 ** On entry to DGEMM  parameter number  8 had an illegal value
[[0. 0. 0.]
 [0. 0. 0.]
 [0. 0. 0.]]
```
Am I doing something bad, or is this a (known?) bug?

#### Possible cause from @m-zarrin 

It looks like the (n, 1) column vector path is passing an invalid leading dimension to BLAS: ldb = 1, when it should satisfy ldb ≥ n. As a result, BLAS silently zeros out the output. The (1, n) row vector path seems to avoid this issue somehow.
Until this is fixed, you can use the row-vector form for outer products:
```python
# Instead of:
y = model.integer(shape=(SHAPE, 1))
yyT = matmul(y, transpose(y))          # triggers the bug 
# matmul(y, transpose(y)) where y is (3,1)	(3,1) @ (1,3)	X all zeros (BLAS bug)

# Use:
y = model.integer(shape=(1, SHAPE))
yyT = matmul(transpose(y), y)          # identical math, no bug
# matmul(transpose(v), v) where v is (1,3)	(3,1) @ (1,3)
```
This gives the same mathematical result while avoiding the BLAS issue.

arcondello:
Closes https://github.com/dwavesystems/dwave-optimization/issues/553

I did cheat a bit by doing most of the testing from the Python level. It's just a lot easier :grimacing: 

#### AI Generation Disclosure
A used ChatGPT a bit to help me understand BLAS's requirements on inputs. It was not used for any code generation.

### DIFF
--- a/dwave/optimization/src/nodes/linear_algebra.cpp
+++ b/dwave/optimization/src/nodes/linear_algebra.cpp
@@ -102,77 +102,113 @@ template <>
 void gemm<double>(
     const ssize_t m,
     const ssize_t n,
-    const ssize_t k,  // size of the arrays
+    const ssize_t k,
     const double* const A,
-    std::span<const ssize_t, 2> A_strides,  // lhs matrix
+    std::span<const ssize_t, 2> A_strides,
     const double* const B,
-    std::span<const ssize_t, 2> B_strides,  // rhs matrix
+    std::span<const ssize_t, 2> B_strides,
     double* const C,
     std::span<const ssize_t, 2> C_strides
-) {  // output matrix
-    // OpenBLAS has some requirements for A,B,C.
-    // Specifically, they must be contiguous within each "row" and they must
-    // have positive strides in the first dimension.
-    // So, if they don't satisfy those requirements, we copy the state into
-    // a contiguous array that does.
+) {
+    // dev note: we could check here if we want to actually use dot or gemv, for
+    // now let's keep it simple and always use gemm
 
-    if (A_strides[0] < 0 or A_strides[1] != sizeof(double)) {
+    static constexpr ssize_t itemsize = sizeof(decltype(*A));
+
+    // One special case we do want to handle before anything else is the k==1 case.
+    // In this case A_strides[1] and B_strides[0] can be any value at all, but
+    // blas is pickier than that.
+    // So let's use equivalent strides that blas is happier with and try again.
+    if (k == 1 and (A_strides[1] != itemsize or B_strides[0] != n * itemsize)) {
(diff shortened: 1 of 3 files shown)


dwave/optimization/src/nodes/linear_algebra.cpp
```diff
@@ -102,77 +102,113 @@ template <>
...
+    // now let's keep it simple and always use gemm
 
-    if (A_strides[0] < 0 or A_strides[1] != sizeof(double)) {
+    static constexpr ssize_t itemsize = sizeof(decltype(*A));
+
+    // One special case we do want to handle before anything else is the k==1 case.
+    // In this case A_strides[1] and B_strides[0] can be any value at all, but
+    // blas is pickier than that.
+    // So let's use equivalent strides that blas is happier with and try again.
+    if (k == 1 and (A_strides[1] != itemsize or B_strides[0] != n * itemsize)) {
```
wbernoudy, line 122: Why the second part of this check? I get that it will be handled correctly by the rest of the implementation in the case that `A_strides[1]` and `B_strides[0]` are "correct," but is there another reason?

wbernoudy (approved):

dwave/optimization/src/nodes/linear_algebra.cpp
arcondello: We're making a recursive call in this `if`-branch, without this check we'd just infinitely loop. We could actually loosen it a bit, I think we just need `B_strides[0] >= n * itemsize`, but this way we're just dodging the recursion.

wbernoudy: Got it, thanks

Issue #399, opened 2025-09-25
Style: Consider allowing `and`, `or`, and `not` keywords in C++ code


arcondello:
C++ supports `and`, `or`, `not` [keywords](https://en.cppreference.com/w/cpp/keywords.html). `not` especially seems much more readable to me in `if`-clauses 
```c++
if (not std::is_sorted(range)) {}
```
vs
```c++
if (!std::is_sorted(range)) {}
```
Currently we don't use them anywhere in the code base, mostly because the [Google C++ style guide]( https://google.github.io/styleguide/cppguide.html#Boolean_Expressions) disallows them
> Also note that you should always use the punctuation operators, such as && and ~, rather than the word operators, such as and and compl.

But IMO we should diverge for readability. Though this is coming from someone who was raised on Python.

Issue #45, opened 2024-07-05
Consider adding `Intersection` and `Union` nodes/symbols


arcondello:
It would be useful to have nodes that propagate the intersection or union of two predecessor symbols.

fastbodin:
`Intersection` can now be done with `IsInNode` and `ExtractNode`.

```python
>>> from dwave.optimization.model import Model
>>> from dwave.optimization.mathematical import isin, extract
>>> model = Model()
>>> element = model.integer(3)
>>> test_elements = model.integer(5)
>>> contains = isin(element, test_elements)
>>> intersection = extract(contains, element)
>>> with model.lock():
...     model.states.resize(1)
...     element.set_state(0, [1, 2, 3])
...     test_elements.set_state(0, [2, 3, 4, 5, 6])
...     print(intersection.state(0))
[2. 3.]
```

PR #557, opened 2026-05-29
Add DisjointCover node

smharwood:
Adds a logical-valued node that determines whether its predecessors are a disjoint cover of a given primary set

Implements a `DisjointCover` node, enabling construction of `DisjointLists` via `List` symbols

#### AI Generation Disclosure
No AI generated code was used. AI tools like Gemini were consulted for best practices and architectural suggestions.

### DIFF
--- a/dwave/optimization/include/dwave-optimization/nodes/set_routines.hpp
+++ b/dwave/optimization/include/dwave-optimization/nodes/set_routines.hpp
@@ -22,6 +22,45 @@
 
 namespace dwave::optimization {
 
+class IsDisjointCoverNode : public ScalarOutputMixin<EqualityMixin<ArrayNode>, false> {
+ public:
+    IsDisjointCoverNode(std::span<ArrayNode*> node_ptrs, ssize_t primary_set_size);
+
+    /// @copydoc Array::buff()
+    double const* buff(const State& state) const override;
+
+    /// @copydoc Node::commit()
+    void commit(State& state) const override;
+
+    /// @copydoc Array::diff()
+    std::span<const Update> diff(const State& state) const override;
+
+    /// @copydoc Node::initialize_state()
+    void initialize_state(State& state) const override;
+
+    /// @copydoc Array::integral()
+    bool integral() const override;
+
+    /// @copydoc Array::max()
+    double max() const override;
+
+    /// @copydoc Array::min()
+    double min() const override;
+
+    /// The size of the primary set to be covered
+    ssize_t primary_set_size() const { return primary_set_size_; };
+
+    /// @copydoc Node::propagate()
+    void propagate(State& state) const override;
+
+    /// @copydoc Node::revert()
+    void revert(State& state) const override;
+
+ private:
+    ssize_t primary_set_size_;
+    std::vector<Array*> operands_;
+};
+
 class IsInNode : public ArrayOutputMixin<EqualityMixin<ArrayNode>> {
  public:
     IsInNode(ArrayNode* element_ptr, ArrayNode* test_elements_ptr);
--- a/dwave/optimization/include/dwave-optimization/nodes/set_routines.hpp
+++ b/dwave/optimization/include/dwave-optimization/nodes/set_routines.hpp
@@ -22,7 +22,8 @@
 
 namespace dwave::optimization {
 
-class IsDisjointCoverNode : public ScalarOutputMixin<EqualityMixin<ArrayNode>, false> {
+class IsDisjointCoverNode
+    : public ScalarOutputMixin<EqualityMixin<ArrayNode, IsDisjointCoverNode>, false> {
  public:
     IsDisjointCoverNode(std::span<ArrayNode*> node_ptrs, ssize_t primary_set_size);
 
@@ -34,6 +35,9 @@ class IsDisjointCoverNode : public ScalarOutputMixin<EqualityMixin<ArrayNode>, f
 
     /// @copydoc Array::diff()
     std::span<const Update> diff(const State& state) const override;
+
+    /// @copydoc Node::equal_to()
+    bool equal_to(const IsDisjointCoverNode& rhs) const override;
 
     /// @copydoc Node::initialize_state()
     void initialize_state(State& state) const override;
@@ -59,6 +63,8 @@ class IsDisjointCoverNode : public ScalarOutputMixin<EqualityMixin<ArrayNode>, f
  private:
     ssize_t primary_set_size_;
     std::vector<Array*> operands_;
+
+    void replace_predecessor_(ssize_t index, Node* node_ptr) override;
 };
 
 class IsInNode : public ArrayOutputMixin<EqualityMixin<ArrayNode>> {
--- a/dwave/optimization/mathematical.py
+++ b/dwave/optimization/mathematical.py
@@ -37,6 +37,7 @@ from dwave.optimization.symbols import (
     Exp,
     Expit,
     Extract,
+    IsDisjointCover,
     IsIn,
     LessEqual,
     LinearProgram,
@@ -93,6 +94,7 @@ __all__ = [
     "expit",
     "extract",
     "hstack",
+    "is_disjoint_cover",
     "isin",
     "less_equal",
     "linprog",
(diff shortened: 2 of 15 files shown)


dwave/optimization/include/dwave-optimization/nodes/set_routines.hpp
```diff
@@ -75,4 +75,47 @@ class IsInNode : public ArrayOutputMixin<ArrayNode> {
     const Array* test_elements_ptr_;
 };
 
+class DisjointCoverNode : public ScalarOutputMixin<ArrayNode, false> {
+ public:
+    DisjointCoverNode(ssize_t primary_set_size, std::span<ArrayNode*> node_ptrs);
+
+    // Node overloads **********
```
arcondello: I know this is a pattern used by a lot of other nodes, but I want to move to just alphabetical for all methods going forward.

dwave/optimization/symbols/set_routines.pyx
```diff
@@ -43,3 +50,61 @@ cdef class IsIn(ArraySymbol):
...
 
 _register(IsIn, typeid(IsInNode))
+
+
+cdef class DisjointCover(ArraySymbol):
+    """Tests whether the symbols are disjoint, set-like, and cover a set of integers
+
+    See Also:
+
+    .. versionadded::
```
arcondello: typo

arcondello (commented):
Couple random aesthetic comments

arcondello:
Misc unhelpful notes from trying to find precedent for similar functions in other libraries we like to follow.


* [NumPy's Set routines](https://numpy.org/doc/2.3/reference/routines.set.html). Much closer to https://github.com/dwavesystems/dwave-optimization/issues/45 than this.
* [Python's set methods](https://docs.python.org/3/library/stdtypes.html#frozenset.isdisjoint). One thing to learn there is I think we want `IsDisjointCover` if we're outputting a `bool`? Definitely if we were implementing a `Disjoint` symbol we'd want `IsDisjoint`. In this case it's not clear what `DisjointCover` would be, so maybe adding `Is*` is redundant?

Maybe we want `IsPartition`? Though I am a little nervous about clashing with [np.partition()](https://numpy.org/doc/2.3/reference/generated/numpy.partition.html) which is a totally different thing. [And I think a partition might actually requires that each subset be non-empty](https://en.wikipedia.org/wiki/Partition_of_a_set).

arcondello:
I think if we were writing this from scratch we'd want an API like
```python
from dwave.optimization import Model, is_disjoint_cover

model = Model()

universe = 100
num_lists = 10

lists = [model.list(universe, min_size=0) for _ in range(num_lists)]

model.add_constraint(is_disjoint_cover(lists))
```
open to suggestions

smharwood:
> Misc unhelpful notes from trying to find precedent for similar functions in other libraries we like to follow.
> 
> * [NumPy's Set routines](https://numpy.org/doc/2.3/reference/routines.set.html). Much closer to [Consider adding `Intersection` and `Union` nodes/symbols #45](https://github.com/dwavesystems/dwave-optimization/issues/45) than this.
> * [Python's set methods](https://docs.python.org/3/library/stdtypes.html#frozenset.isdisjoint). One thing to learn there is I think we want `IsDisjointCover` if we're outputting a `bool`? Definitely if we were implementing a `Disjoint` symbol we'd want `IsDisjoint`. In this case it's not clear what `DisjointCover` would be, so maybe adding `Is*` is redundant?
> 
> Maybe we want `IsPartition`? Though I am a little nervous about clashing with [np.partition()](https://numpy.org/doc/2.3/reference/generated/numpy.partition.html) which is a totally different thing. [And I think a partition might actually requires that each subset be non-empty](https://en.wikipedia.org/wiki/Partition_of_a_set).

I'm convinced adding "Is" is the way to go. At least, its consistent with "IsIn", and its probably not a good idea to go the other way and rename that to just "In"

smharwood:
> I think if we were writing this from scratch we'd want an API like
> 
> ```python
> from dwave.optimization import Model, is_disjoint_cover
> 
> model = Model()
> 
> universe = 100
> num_lists = 10
> 
> lists = [model.list(universe, min_size=0) for _ in range(num_lists)]
> 
> model.add_constraint(is_disjoint_cover(lists))
> ```
> 
> open to suggestions

So, add `is_disjoint_cover` to the mathematical namespace?

dwave/optimization/include/dwave-optimization/nodes/set_routines.hpp
smharwood: Should be fixed now

dwave/optimization/mathematical.py
```diff
@@ -1110,6 +1112,44 @@ def hstack(arrays: collections.abc.Sequence[ArraySymbol]) -> ArraySymbol:
         return concatenate(arrays, 1)
 
 
+def is_disjoint_cover(primary_set_size: int, subsets: list[ArraySymbol]) -> IsDisjointCover:
```
smharwood: Can't tell if I should future-proof this interface and make the primary set an `ArraySymbol`? (but for now, just assert its a constant)

smharwood:
> I think if we were writing this from scratch we'd want an API like
> 
> ```python
> from dwave.optimization import Model, is_disjoint_cover
> 
> model = Model()
> 
> universe = 100
> num_lists = 10
> 
> lists = [model.list(universe, min_size=0) for _ in range(num_lists)]
> 
> model.add_constraint(is_disjoint_cover(lists))
> ```
> 
> open to suggestions

Finally revisiting this and I think I understand your suggestion: you would prefer to remove the "primary set size" argument to `is_disjoint_cover` and just infer it from the (common) max value of the lists?

I'm tempted to agree - this would address my other comment about future-proofing the interface. Just wondering if we wouldn't want the flexibility  in the future.

arcondello:
Can make it a kwarg with `primary_set_size=None` where `None` means infer from inputs?

dwave/optimization/src/nodes/set_routines.cpp
```diff
@@ -20,6 +20,199 @@
 
 namespace dwave::optimization {
 
+// IsDisjointCoverNode *******************************************************************
+
+struct IsDisjointCoverNodeData : public NodeStateData {
+ private:
+    // simple wrapper around ssize_t with default value 1, indicating no violations
+    struct Count_ {
```
arcondello, line 28: Could you get the same effect by messing with the `Allocator` template to the `unordered_map`? https://en.cppreference.com/cpp/container/unordered_map

That wouldn't necessarily be better.

dwave/optimization/src/nodes/set_routines.cpp
```diff
@@ -20,6 +20,199 @@
...
+};
+
+IsDisjointCoverNode::IsDisjointCoverNode(
+    std::span<ArrayNode*> node_ptrs,
+    ssize_t primary_set_size
+) :
+    ScalarOutputMixin<EqualityMixin<ArrayNode, IsDisjointCoverNode>, false>(),
+    primary_set_size_(primary_set_size) {
+    for (const auto& node : node_ptrs) {
+        if (!node->integral()) {
```
arcondello:
```cpp
        if (not node->integral()) {
```
https://github.com/dwavesystems/dwave-optimization/issues/399

dwave/optimization/src/nodes/set_routines.cpp
```diff
@@ -20,6 +20,199 @@
...
+    for (const auto& node : operands_) {
+        for (const auto& value : node->view(state)) {
+            ssize_t element = static_cast<ssize_t>(value);
+            // Should not be possible because of checks in constructor
+            assert(element < primary_set_size_);
+            assert(element >= 0);
+            count[element] += 1;
+        }
+    }
+    emplace_data_ptr_<IsDisjointCoverNodeData>(state, count);
```
arcondello:
```cpp
    emplace_data_ptr_<IsDisjointCoverNodeData>(state, std::move(count));
```

dwave/optimization/symbols/set_routines.pyx
```diff
@@ -14,10 +14,75 @@
...
+from dwave.optimization._model cimport _Graph, _register, ArraySymbol, Symbol
+from dwave.optimization.libcpp cimport dynamic_cast_ptr
+from dwave.optimization.libcpp.graph cimport ArrayNode
+from dwave.optimization.libcpp.nodes.set_routines cimport IsDisjointCoverNode, IsInNode
+
+
+cdef class IsDisjointCover(ArraySymbol):
+    """Tests whether the symbols are disjoint, set-like, and cover a set of integers
+
+    See Also:
```
arcondello, line 33: missing link

dwave/optimization/mathematical.py
```diff
@@ -1116,6 +1118,52 @@ def hstack(arrays: collections.abc.Sequence[ArraySymbol]) -> ArraySymbol:
         return concatenate(arrays, 1)
 
 
+def is_disjoint_cover(subsets: list[ArraySymbol], *, primary_set_size: int|None = None) -> IsDisjointCover:
```
arcondello:
```python
def is_disjoint_cover(subsets: list[ArraySymbol], *, primary_set_size: int | None = None) -> IsDisjointCover:
```
spacing around the `|`

arcondello (approved):
Couple minor aesthetic comments, otherwise LGTM!

dwave/optimization/src/nodes/set_routines.cpp
smharwood: I'm not strong enough to look at C++ language specs like that and figure out how to do it, so I asked Gemini and it suggested the custom structure (which is probably where I got the idea in the first place)

dwave/optimization/mathematical.py
smharwood: How do we feel about getting some pre-commit hooks set up?

arcondello: We have generally said no for Ocean packages under the reasoning that it raises the bar for users. See also https://github.com/dwavesystems/dwave-optimization/issues/113

PR #558, opened 2026-06-01
Add a check for release notes

arcondello:
Add a release note check. It's very simple, just checks whether a new PR has a release note and adds a comment if not.

#### AI Generation Disclosure
I used ChatGPT (free public version) to write the first version of the file and then edited it. I also relied on it pretty heavily to help with the syntax.

### DIFF
--- a/.github/workflows/releasenote-check.yml
+++ b/.github/workflows/releasenote-check.yml
@@ -20,8 +20,6 @@ jobs:
         with:
           # We use GitHub commands to check the files rather than git in order to
           # avoiding needing to check out the whole repo.
-          # We use reno for release notes so each note is a new file. Hence why we
-          # only check for files added.
           script: |
             const owner = context.repo.owner;
             const repo = context.repo.repo;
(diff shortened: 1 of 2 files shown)


.github/workflows/releasenote-check.yml
```diff
@@ -0,0 +1,61 @@
...
+                owner,
+                repo,
+                pull_number,
+                per_page: 100,
+              }
+            );
+
+            const hasReleaseNote = files.some(file =>
+              file.status === "added" &&
+              file.filename.startsWith("releasenotes/")
```
randomir, line 41: To make sure the file was actually added (a new release note), and not just significantly modified one, you can try with:

```suggestion
              file.filename.startsWith("releasenotes/") &&
              file.additions > 0 &&
              file.deletions == 0
```

randomir (commented):
Nice addition.

.github/workflows/releasenote-check.yml
arcondello: IMO if one touched the release notes in any way it's safe to not fire the warning. There are a lot of edge cases here but I think it's best to err on simplicity. This is really just because I keep forgetting entirely.

wbernoudy (approved):

.github/workflows/releasenote-check.yml
arcondello: I actually ended up changing the logic. Now any change in the `releasenotes/` will suppress the warning. Under the assumption that if the contributor made any changes, then the regular code review will catch any issues.

I also don't remind the user if there were no changes in `dwave/`. There are a few other cases we'd want a release note but I think this will reduce spam a lot and is simple.

wbernoudy (approved):

Issue #398, opened 2025-09-25
Style: Consider using leading underscores for protected/private attributes and methods in C++


arcondello:
Currently we follow [Google's C++ Style guide](https://google.github.io/styleguide/cppguide.html#Variable_Names) and use trailing underscores for protected/private attributes
```c++
class A {
 public:
    void a_method();
    int a;
 protected:
    void b_method();
    int b_;
 private:
    void c_method();
    int c_;
};
```

I propose that we switch to using leading underscores for both attributes and methods in our C++ code base.
```c++
class A {
 public:
    void a_method();
    int a;
 protected:
    void _b_method();
    int _b;
 private:
    void _c_method();
    int _c;
};
```

*Alternatives*

Use trailing underscores for both methods and attributes. This would be more C++-y and less Pythonic, but IMO less readable.

wbernoudy:
https://stackoverflow.com/a/228797

fastbodin:
Given we are already using this for some private attributes, I am happy for it to be consistent throughout.

arcondello:
Do we also want to use this style for functions and/or classes that are only defined in the `.cpp` file? I.e., behind the compilation barrier.

If we chose to do it for classes then most of the state data classes would have a name change.

wbernoudy:
The one counterargument I can think of is that functions/classes only defined in the `.cpp` can be used by multiple other "public" classes/functions implemented in the `.cpp`, so in a sense they are not as "private". But I think the clarity gained from being able to glance at the function/class name and being able to tell whether it is meant to be exposed in the `.hpp` or not is worth it, so I'm in favor!

And just to clarify, we plan to use this style for private/protected methods as well as members, right? E.g.
```c++
class A {
 public:
    void a_method();
    int a;
 protected:
    void b_method_();
    int b_;
 private:
    void c_method_();
    int c_;
};
```

arcondello:
> The one counterargument I can think of is that functions/classes only defined in the .cpp can be used by multiple other "public" classes/functions implemented in the .cpp, so in a sense they are not as "private". But I think the clarity gained from being able to glance at the function/class name and being able to tell whether it is meant to be exposed in the .hpp or not is worth it, so I'm in favor!

Yeah, the distinction I am going for is "part of the public interface" vs not.

> And just to clarify, we plan to use this style for private/protected methods as well as members, right? E.g.

[...]

PR #562, opened 2026-06-05
Refactor collections nodes

arcondello:
No functional changes.

Refactors to our ~decision~ collections (edit, de-scoping, can make another PR for other decisions) nodes to bring them more in line with our current standards (e.g., https://github.com/dwavesystems/dwave-optimization/issues/399, https://github.com/dwavesystems/dwave-optimization/issues/398, alphabetizing methods, better use of private attributes, etc).

#### AI Generation Disclosure
No AI used. Though honestly it probably would have been useful.

### DIFF
--- a/dwave/optimization/src/nodes/collections.cpp
+++ b/dwave/optimization/src/nodes/collections.cpp
@@ -23,9 +23,9 @@ namespace dwave::optimization {
 // Given a collection, check that it is a valid sub-permutation of range(n),
 // and then "augment" it so that the values not in collection are appended onto
 // the end.
-std::vector<double> augment_collection(std::vector<double> values, const ssize_t n) {
+std::vector<double> augment_collection_(std::vector<double> values, const ssize_t n) {
     // First check that our given values has the values we expect
-    if (!std::ranges::all_of(values, is_integer)) {
+    if (not std::ranges::all_of(values, is_integer)) {
         throw std::invalid_argument("values values must be integers");
     }
     if (std::ranges::any_of(values, [](const auto& val) -> bool { return val < 0; })) {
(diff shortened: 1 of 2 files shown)


dwave/optimization/src/nodes/collections.cpp
```diff
@@ -432,24 +455,37 @@ void DisjointBitSetsNode::swap_between_sets(
...
         from_disjoint_set, to_disjoint_set, element
     );
 }
 
 ssize_t DisjointBitSetsNode::get_containing_set_index(State& state, ssize_t element) const {
-    return data_ptr<DisjointBitSetsNodeData>(state)->get_containing_set_index(element);
+    return data_ptr<DisjointBitSetsNodeData_>(state)->get_containing_set_index(element);
+}
+
+DisjointBitSetNode::DisjointBitSetNode(DisjointBitSetsNode* disjoint_bit_sets_node) :
```
wbernoudy, line 467: Should this be at the top of the `DisjointBitSetNode` section?

arcondello: yes it should, good catch.

arcondello: Oh wait, it is. `DisjointBitSetsNode` (plural) vs `DisjointBitSetNode` (singular).

~Though I suppose we should change the order of nodes for the singular to be above the plural in both cases.~ Edit: I think I'll leave it as-is. While they are out of lexicographic order, this way we don't need to do a forward declaration and IMO this is an example where the exception is helpful rather than harmful. Also also, hopefully these nodes [are going away](https://github.com/dwavesystems/dwave-optimization/pull/557).

arcondello:
@wbernoudy tells me he approves

PR #566, opened 2026-06-18
Move documentation images into a subdirectory

arcondello:
By moving these images into a subdirectory, we can avoid some duplication in the sdk docs.

Also deletes some unused images added in the README refactor (https://github.com/dwavesystems/dwave-optimization/pull/534, https://github.com/dwavesystems/dwave-optimization/pull/529)

#### AI Generation Disclosure
No AI used.


JoelPasvolsky (approved):
Thanks!

PR #567, opened 2026-06-23
Enforce `const state` in `NumberNode::get_value()`

fastbodin:
Modified ``NumberNode::get_value()`` arguments so that it accepts a ``const state`` as opposed to a non-const ``state``.

#### AI Generation Disclosure

No AI tools used.


arcondello (approved):

alexzucca90:
no release notes?

arcondello:
I was hemming and hawing about it. I suppose we should have one. Also still not sure why https://github.com/dwavesystems/dwave-optimization/pull/558 isn't firing.

arcondello (approved):

PR #568, opened 2026-06-23
Fix release note check GitHub Action

arcondello:
I believe this should fix it.

See also https://github.com/dwavesystems/dwave-optimization/pull/558

#### AI Generation Disclosure

Use ChatGPT to help diagnose the problem. Though it insisted the old syntax should work :shrug:

### DIFF
--- a/.github/workflows/releasenote-check.yml
+++ b/.github/workflows/releasenote-check.yml
@@ -16,7 +16,7 @@ jobs:
 
     steps:
       - name: Check for release note files
-        uses: actions/github-script@v7
+        uses: actions/github-script@v9
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/releasenote-check.yml
```diff
@@ -16,7 +16,7 @@ jobs:
 
     steps:
       - name: Check for release note files
-        uses: actions/github-script@v7
+        uses: actions/github-script@v9
```
arcondello, line 19: This is actually unrelated but fixes another warning so sneaking it in.

fastbodin (approved):

Issue #565, opened 2026-06-18
Style: Use qualified auto in C++ code


arcondello:
We obviously use pointers types a lot in the C++ code. For example we do
```c++
auto x_ptr = graph.emplace_node<BinaryNode>();
```

IMO we should be using qualified `auto` in these places
```c++
auto* x_ptr = graph.emplace_node<BinaryNode>();
```
for readability.

Obviously there is a semantic difference if later we need to do say `decltype(auto)` but in most places we never inspect the `auto` and the latter conveys more meaning.

See
- [LLVM Coding Standards](https://llvm.org/docs/CodingStandards.html?utm_source=chatgpt.com#beware-unnecessary-copies-with-auto): "Use auto & for values and auto * for pointers unless you need to make a copy."
- [clang-tidy](https://clang.llvm.org/extra/clang-tidy/checks/readability/qualified-auto.html?utm_source=chatgpt.com)'s readability-qualified-auto

PR #569, opened 2026-06-23
Misc `graph.hpp`/`graph.cpp` QOL and aesthetic improvements

arcondello:
PR for a grab bag of QOL features on Node/Graph that I made while working on another feature

#### AI Generation Disclosure
No AI used.

### DIFF
--- a/dwave/optimization/include/dwave-optimization/graph.hpp
+++ b/dwave/optimization/include/dwave-optimization/graph.hpp
@@ -33,9 +33,10 @@
 namespace dwave::optimization {
 
 class ArrayNode;
-class Node;
+class ConstantNode;
 class DecisionNode;
 class InputNode;
+class Node;
 
 // A decision is an independent variable in the model. The value(s) to be optimized.
 struct Decision {};
@@ -64,6 +65,10 @@ class Graph {
     /// Commit the changes on each changed node.
     void commit(State& state, std::span<const Node*> changed) const;
     void commit(State& state, std::vector<const Node*>&& changed) const;
+
+    /// Return the constants nodes in the graph.
+    std::span<ConstantNode* const> constants() noexcept { return constants_; }
+    std::span<const ConstantNode* const> constants() const noexcept { return constants_; }
 
     /// Return the constraints of the graph as a span of nodes.
     std::span<ArrayNode* const> constraints() noexcept { return constraints_; }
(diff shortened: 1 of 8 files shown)


dwave/optimization/src/graph.cpp
```diff
@@ -240,70 +242,87 @@ void Graph::recursive_reset(State& state, const Node* ptr) {
...
+
+    for (auto& uptr : nodes_ | std::views::reverse) {
+        if (uptr->topological_index_ == keep) continue;  // we marked these to keep
+        if (uptr->successors().size() > 0) continue;  // this node is used by other nodes
+
+        // We have a node with no successors and that we haven't marked it as important.
+        // So let's mark it to be dropped later.
+
+        // Remove the node from its predecessor's successor vectors.
+        // This very leaves the node in an invalid state, until we delete it later.
```
fastbodin: `// This leaves the node ...` <- remove leave.

dwave/optimization/src/graph.cpp
```diff
@@ -240,70 +242,87 @@ void Graph::recursive_reset(State& state, const Node* ptr) {
...
+        for (auto* pred_ptr : uptr->predecessors_) {
+            [[maybe_unused]] ssize_t num_removed = pred_ptr->remove_successor_(uptr.get());
+            assert(num_removed > 0);
```
fastbodin, line 296: This is `> 0` as opposed to `==1` since a node may catch a successor pointer multiple times?

arcondello: Yes that's right. E.g. `Add(x, x)`

arcondello:
For the `auto*` change, see https://github.com/dwavesystems/dwave-optimization/issues/565

dwave/optimization/include/dwave-optimization/graph.hpp
```diff
@@ -307,8 +321,8 @@ class Node {
     friend void Graph::topological_sort();
     friend void Graph::reset_topological_sort();
     template <class NodeType, class... Args>
-    friend NodeType* Graph::emplace_node(Args&&... args);
-    friend ssize_t Graph::remove_unused_nodes(bool ignore_listeners);
+    friend NodeType* Graph::emplace_node(Args&&...);
```
fastbodin, line 324: I believe `Args&&... args` and `Args&&...` for this declaration are equivalent. Is this a stylistic preference I am missing here?

fastbodin (approved):

dwave/optimization/include/dwave-optimization/graph.hpp
arcondello: Yes, I honestly forget why I made this change. Probably just that the `friend` declaration doesn't really need that info. I can revert it if anyone has strong feelings but :shrug:

fastbodin: Fine with the change. I am just trying to update my own hardware 😆 

Was it this: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rf-unused

arcondello: Love me a core guidelines citation :laughing: I think in this case they aren't unused exactly, just not necessary info for this declaration? Honestly I find friend declarations just kind of inherently weird. I can never decide whether I think they should be `private` or `public`.

fastbodin: Ha. Right. Because it technically does not matter 🤦 . I am surprised there is not a style guide here!

PR #571, opened 2026-06-24
Add method for removing redundant nodes

### LINKED ISSUE #563: Add methods to support the removal of redundant nodes
https://github.com/dwavesystems/dwave-optimization/issues/41 and https://github.com/dwavesystems/dwave-optimization/pull/161 added the ability to remove unused nodes from the `Graph`. The logical next step is the ability to remove redundant nodes. This would require two things:
1) We need some method to test whether two nodes are equal (or maybe equivalent?), that is that they will always be guaranteed to take the same value. We could also replace [the Cython-level function](https://github.com/dwavesystems/dwave-optimization/blob/121cd379b8b397dc9ecc21daca1080cc3f40d052/dwave/optimization/_model.pyx#L1098).
2) We need a method to transfer successors from one node to another. So probably something like `Graph::transfer_successors(Node* from_ptr, Node* to_ptr)` which uses `Node::replace_predecessor(...)`. We need the latter because some nodes [keep redundant pointers](https://github.com/dwavesystems/dwave-optimization/blob/121cd379b8b397dc9ecc21daca1080cc3f40d052/dwave/optimization/include/dwave-optimization/nodes/binaryop.hpp#L79).

With those two in place, we can check for equality, and in that case transfer the successors. Having done that as much as possible, we then run the `Graph::remove_redundant_nodes()` function.

arcondello:
Closes https://github.com/dwavesystems/dwave-optimization/issues/563

#### AI Generation Disclosure
No AI used.


dwave/optimization/include/dwave-optimization/nodes/binaryop.hpp
```diff
@@ -46,10 +52,10 @@ class BinaryOpNode : public ArrayOutputMixin<ArrayNode> {
     /// @copydoc Array::max()
     double max() const override;
 
-    using ArrayOutputMixin::shape;
+    using ArrayOutputMixin<EqualityMixin<ArrayNode, BinaryOpNode<BinaryOp>>>::shape;
```
arcondello, line 54: See the commit message for https://github.com/dwavesystems/dwave-optimization/pull/571/commits/5b213bca8a005de7f705f5578c66117a597852ed

> This requires some fiddling, see
https://isocpp.org/wiki/faq/templates#nondependent-name-lookup-members
for the why.

dwave/optimization/include/dwave-optimization/graph.hpp
```diff
@@ -418,4 +473,44 @@ class DecisionNode : public Decision, public virtual Node {
...
+        // lhs and rhs are the same type, so let's go to our type-specific
+        // method.
+        return equal_to(*rhs_ptr);
+    }
+
+    // Nodes must implement `equal_to()` for their own type.
+    virtual bool equal_to(const Derived& rhs) const = 0;
+};
+
+/// Overload for EqualityMixin that does not require a class-specific equal_to()
```
arcondello, line 496: A doubt I have is whether these two overloads are different enough that they should actually be given different names 🤔

dwave/optimization/src/nodes/lp.cpp
```diff
@@ -16,8 +16,12 @@
 
 #include <string>
 
+#include <sys/types.h>
```
arcondello: oops, this needs to be removed

arcondello:
As a note: I didn't try to be inclusive about the `assert`s in the `replace_predecessor_()` method. That way lies madness so I just went with some sanity checks.

tests/cpp/test_graph.cpp
```diff
@@ -316,6 +316,71 @@ TEST_CASE("Graph::objective()") {
...
+
+                    // we kept the ones earlier in the topological order
+                    CHECK(graph.nodes()[2].get() == left_x_plus_y);
+                    CHECK(graph.nodes()[3].get() == negative_left_x_plus_y);
+                }
+            }
+        }
+    }
+
+    // todo: test redundant constants
```
arcondello: Ah, I need to address these todos

dwave/optimization/_model.pyx
```diff
@@ -928,6 +928,46 @@ cdef class _Graph:
...
+            int: Number of symbols removed.
+
+        Examples:
+
+            >>> from dwave.optimization import Model
+            ...
+            >>> model = Model()
+            ...
+            >>> x = model.integer(5)
+            ...
```
fastbodin: Nit: I find that the `...` do not add much to the example. Additionally, most of the other examples do not include `...`. Am I missing something?

dwave/optimization/src/graph.cpp
```diff
@@ -241,6 +242,175 @@ void Graph::recursive_reset(State& state, const Node* ptr) {
...
+            if (not ignore_listeners and ptr->num_listeners() > 0) return false;
+            assert(ptr->topological_index_ != drop or ptr->successors_.empty());
+            return ptr->topological_index() == drop;
+        };
+
+        // Go through our "special" nodes and drop anything
+        assert(std::ranges::none_of(decisions_, dropped));  // should never be dropped
+        assert(std::ranges::none_of(inputs_, dropped));  // should never be dropped
+        std::erase_if(constants_, dropped);
+
```
fastbodin, line 287:
```suggestion
```

dwave/optimization/src/graph.cpp
```diff
@@ -241,6 +242,175 @@ void Graph::recursive_reset(State& state, const Node* ptr) {
...
+        std::erase_if(constraints_, dropped);
+        assert(objective_ptr_ == nullptr or objective_ptr_->topological_index_ != drop);
```
fastbodin, line 289: Duplicate of line 274? Both do not seem necessary?

dwave/optimization/src/graph.cpp
```diff
@@ -241,6 +242,175 @@ void Graph::recursive_reset(State& state, const Node* ptr) {
...
+
+        // This is the step that actually deallocates the node
+        for (auto& uptr : nodes_) {
+            if (not dropped(uptr)) continue;
+
+            // Remove the node from its predecessor's successor vectors.
+            // This very leaves the node in an invalid state, very briefly
```
fastbodin:
```suggestion
            // This leaves the node in an invalid state, very briefly
```

dwave/optimization/include/dwave-optimization/nodes/binaryop.hpp
```diff
@@ -63,20 +68,22 @@ class BinaryOpNode : public ArrayOutputMixin<ArrayNode> {
...
 
  private:
+    void replace_predecessor_(ssize_t previous_index, Node* node_ptr) override;
+
     BinaryOp op;
 
     // There are redundant, because we could dynamic_cast each time from
     // predecessors(), but this is more performant
-    std::array<Array* const, 2> operands_;
+    std::array<Array*, 2> operands_;
```
fastbodin, line 86: Right. No longer `const` because they may be replaced.

Resolved.

dwave/optimization/include/dwave-optimization/nodes/collections.hpp
```diff
@@ -138,6 +138,9 @@ class DisjointBitSetNode : public ArrayOutputMixin<ArrayNode> {
 
     std::span<const Update> diff(const State& state) const override;
 
+    // A DisjointBitSetNode can only ever be equal to itself.
```
fastbodin: Nit.
```suggestion
    /// A DisjointBitSetNode can only ever be equal to itself.
```

dwave/optimization/include/dwave-optimization/nodes/collections.hpp
```diff
@@ -228,6 +231,9 @@ class DisjointListNode : public ArrayOutputMixin<ArrayNode> {
 
     std::span<const Update> diff(const State& state) const override;
 
+    // A DisjointListNode can only ever be equal to itself.
```
fastbodin: Nit.
```suggestion
    /// A DisjointListNode can only ever be equal to itself.
```

dwave/optimization/include/dwave-optimization/nodes/manipulation.hpp
```diff
@@ -128,7 +141,7 @@ class ConcatenateNode : public ArrayOutputMixin<ArrayNode> {
 };
 
 /// An array node that is a contiguous copy of its predecessor.
-class CopyNode : public ArrayOutputMixin<ArrayNode> {
+class CopyNode : public ArrayOutputMixin<EqualityMixin<ArrayNode>> {
```
fastbodin, line 144: Why do we have a `CopyNode`? Will it not always just get removed when we call `remove_redundant_nodes()`?

dwave/optimization/include/dwave-optimization/nodes/reduce.hpp
```diff
@@ -110,7 +116,7 @@ class ReduceNode : public ArrayOutputMixin<ArrayNode> {
 
     BinaryOp op;
 
-    Array* const array_ptr_;
+    Array* array_ptr_;
```
fastbodin, line 119: Unrelated to this PR. Is there any reason this should not be `const Array* array_ptr_;`?

dwave/optimization/src/nodes/creation.cpp
```diff
@@ -422,6 +426,45 @@ void ARangeNode::propagate(State& state) const {
     if (ptr->diff().size()) Node::propagate(state);
 }
 
+void ARangeNode::replace_predecessor_(ssize_t index, Node* node_ptr) {
+    Node::replace_predecessor_(index, node_ptr);
+
+    assert(0 <= index and static_cast<size_t>(index) < predecessors().size());
```
fastbodin, line 432: Nit: I realize this is the same but this avoids `size_t`
```cpp
assert(0 <= index and index < static_cast<ssize_t>(predecessors().size()));
```

dwave/optimization/src/nodes/creation.cpp
```diff
@@ -422,6 +426,45 @@ void ARangeNode::propagate(State& state) const {
...
+    if (std::holds_alternative<const Array*>(start_)) {
+        if (index == 0) {
+            // found it!
+            start_ = array_ptr;
+            return;
+        }
+
+        // If we haven't found the Array* we want to replace, then "normalize"
+        // the index so that it looks like it would if start wasn't an array.
+        index -= 1;
```
fastbodin, line 446: I imagine this felt quite annoying when you realized you had to do this.

dwave/optimization/src/nodes/flow.cpp
```diff
@@ -143,6 +143,19 @@ void ExtractNode::propagate(State& state) const {
...
+void ExtractNode::replace_predecessor_(ssize_t index, Node* node_ptr) {
+    Node::replace_predecessor_(index, node_ptr);
+
+    if (index == 0) {
+        condition_ptr_ = dynamic_cast<ArrayNode*>(node_ptr);
+        assert(condition_ptr_ != nullptr);
+    } else {
+        assert(index == 1);
+        arr_ptr_ = dynamic_cast<ArrayNode*>(node_ptr);
+        assert(condition_ptr_ != nullptr);
```
fastbodin:
```suggestion
        assert(arr_ptr_ != nullptr);
```

dwave/optimization/src/nodes/binaryop.cpp
```diff
@@ -387,9 +428,18 @@ void BinaryOpNode<BinaryOp>::propagate(State& state) const {
     if (ptr->diff().size()) Node::propagate(state);
 }
 
+template <class BinaryOp>
+void BinaryOpNode<BinaryOp>::replace_predecessor_(ssize_t previous_index, Node* node_ptr) {
+    Node::replace_predecessor_(previous_index, node_ptr);
+
+    assert(0 <= previous_index and previous_index < 2);
+    operands_[previous_index] = dynamic_cast<ArrayNode*>(node_ptr);
+    assert(operands_[previous_index] != nullptr);
```
fastbodin, line 437: To be clear, we have to call `Node::replace_predecessor_(...)` and include the additional implementation, because, for this node any many other nodes, we cache the predecessors in multiple places

dwave/optimization/src/nodes/indexing.cpp
```diff
@@ -743,6 +750,39 @@ void AdvancedIndexingNode::commit(State& state) const {
...
+    // to the index of the indexing arrays
+    if (index == 0) {
+        // replace the array
+        array_ptr_ = dynamic_cast<ArrayNode*>(node_ptr);
+        assert(array_ptr_ != nullptr);
+        return;
+    }
+    index -= 1;
+
+    //
```
fastbodin:
```suggestion
    // For each `indexer`, check if we're replacing an `Array`.
    // If we haven't found the `Array*` we want to replace, then "normalize" the index ...
```

dwave/optimization/src/nodes/indexing.cpp
```diff
@@ -743,6 +750,39 @@ void AdvancedIndexingNode::commit(State& state) const {
...
+        if (std::holds_alternative<Slice>(indexer)) continue;  // carry on
+
+        assert(std::holds_alternative<ArrayNode*>(indexer));
+        if (index == 0) {
+            // found the indexer we want to replace
+            indexer = dynamic_cast<ArrayNode*>(node_ptr);
+            assert(std::get<ArrayNode*>(indexer) != nullptr);
+            return;
+        }
+        index -= 1;
```
fastbodin, line 779: I realize this is likely excessive, but I am tempted to put an `assert(index >= 0);` after we decrement `index`.

dwave/optimization/src/nodes/indexing.cpp
```diff
@@ -1461,6 +1501,26 @@ std::span<const Update> BasicIndexingNode::diff(const State& state) const {
     return data_ptr_<BasicIndexingNodeData>(state)->diff;
 }
 
+bool BasicIndexingNode::equal_to(const BasicIndexingNode& rhs) const {
+    if (array_ptr_ == rhs.array_ptr_ and              //
+        ndim_ == rhs.ndim_ and                        //
+        std::ranges::equal(shape(), rhs.shape()) and  //
+        std::ranges::equal(strides(), rhs.strides())) {
```
fastbodin, line 1509:
```suggestion
    if (array_ptr_ == rhs.array_ptr_ and              // itself
        ndim_ == rhs.ndim_ and                        // same ndim
        std::ranges::equal(shape(), rhs.shape()) and  // same shape
        std::ranges::equal(strides(), rhs.strides()))  // same strides
        {
```

fastbodin: Hanging comments or just nicer formatting?

dwave/optimization/src/nodes/indexing.cpp
fastbodin: Hanging comment? Something like? Indifferent, just flagging.

dwave/optimization/src/nodes/creation.cpp
```diff
@@ -422,6 +426,45 @@ void ARangeNode::propagate(State& state) const {
...
+
+    const Array* array_ptr = dynamic_cast<const Array*>(node_ptr);
+    assert(array_ptr != nullptr);
+
+    if (std::holds_alternative<const Array*>(start_)) {
+        if (index == 0) {
+            // found it!
+            start_ = array_ptr;
+            return;
+        }
```
fastbodin, line 442: There is no issue here. This is completely fine as-is!

I just wanted to acknowledge an inconsistency that may be intentional for reasons I am not seeing. Sometimes, when there are multiple predecessors to consider, we check `assert(pred_ptr != nullptr);` after reassignment and other times, we check that the argument `node_ptr` (in this case after a `dynamic_cast`) is not a `nullptr`. The end result is the same to my eye.

dwave/optimization/src/nodes/lambda.cpp
```diff
@@ -301,6 +302,16 @@ void AccumulateZipNode::initialize_state(State& state) const {
     );
 }
 
+bool AccumulateZipNode::equal_to(const AccumulateZipNode& rhs) const {
+    // note that we don't have a notion of Graph equality, so we instead
+    // just check whether we have the same underlying shared ptr
+    return (
+        expression_ptr_ == rhs.expression_ptr_ and  //
+        operands_ == rhs.operands_ and              //
+        initial_ == rhs.initial_
```
fastbodin, line 311: Possibly unintentional placeholder comments? Or just for nicer formatting?

dwave/optimization/src/nodes/lambda.cpp
```diff
@@ -354,6 +365,24 @@ const InputNode* const AccumulateZipNode::accumulate_input() const {
     return expression_ptr_->inputs()[0];
 }
 
+void AccumulateZipNode::replace_predecessor_(ssize_t index, Node* node_ptr) {
+    Node::replace_predecessor_(index, node_ptr);
+
```
fastbodin, line 370: Excessive, but I would add `assert(0 <= index);`

dwave/optimization/src/nodes/lambda.cpp
```diff
@@ -354,6 +365,24 @@ const InputNode* const AccumulateZipNode::accumulate_input() const {
...
+            // we're replacing the initial array
+            initial_ = dynamic_cast<ArrayNode*>(node_ptr);
+            assert(std::get<ArrayNode*>(initial_) != nullptr);
+            return;
+        }
+
+        index -= 1;
+    }
+
+    operands_[index] = dynamic_cast<ArrayNode*>(node_ptr);
```
fastbodin, line 387: Prior to this, I might add `assert(0 <= index and index < static_cast<ssize_t>(operands_.size()));`

dwave/optimization/src/nodes/lambda.cpp
```diff
@@ -354,6 +365,24 @@ const InputNode* const AccumulateZipNode::accumulate_input() const {
...
+
+    if (std::holds_alternative<ArrayNode*>(initial_)) {
+        if (index == 0) {
+            // we're replacing the initial array
+            initial_ = dynamic_cast<ArrayNode*>(node_ptr);
+            assert(std::get<ArrayNode*>(initial_) != nullptr);
+            return;
+        }
+
+        index -= 1;
```
fastbodin, line 382: // If we haven't found the Array* we want to replace, then "normalize"
        // the index so that it looks like it would if `initial_` wasn't an array.

dwave/optimization/src/nodes/linear_algebra.cpp
```diff
@@ -541,6 +544,21 @@ void MatrixMultiplyNode::propagate(State& state) const {
     data->assign(data->output);
 }
 
+void MatrixMultiplyNode::replace_predecessor_(ssize_t index, Node* node_ptr) {
+    Node::replace_predecessor_(index, node_ptr);
+
```
fastbodin, line 549: Excessive but I would throw an asset here `assert(0 <= index and index <= 1);`

dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -539,6 +557,17 @@ void ConcatenateNode::propagate(State& state) const {
...
 }
 
+void ConcatenateNode::replace_predecessor_(ssize_t index, Node* node_ptr) {
+    Node::replace_predecessor_(index, node_ptr);
+
+    ArrayNode* array_ptr = dynamic_cast<ArrayNode*>(node_ptr);
+    assert(array_ptr != nullptr);
+    assert(std::ranges::equal(array_ptrs_[index]->shape(), array_ptr->shape()));
+
+    assert(0 <= index and static_cast<size_t>(index) < array_ptrs_.size());
```
fastbodin, line 567: Nit: Totally fine as-is but could replace with
```cpp
    assert(0 <= index and index) < static_cast<ssize_t>(array_ptrs_.size()));
```
to avoid `size_t`. 🤷

dwave/optimization/src/nodes/naryop.cpp
```diff
@@ -315,6 +315,18 @@ void NaryOpNode<BinaryOp>::propagate(State& state) const {
     }
 }
 
+template <class BinaryOp>
+void NaryOpNode<BinaryOp>::replace_predecessor_(ssize_t index, Node* node_ptr) {
+    Node::replace_predecessor_(index, node_ptr);
+
+    assert(0 <= index and static_cast<size_t>(index) < operands_.size());
```
fastbodin, line 322: See Nits above regarding `size_t`. Fine with it as is.

dwave/optimization/src/nodes/unaryop.cpp
```diff
@@ -185,6 +185,19 @@ std::span<const Update> UnaryOpNode<UnaryOp>::diff(const State& state) const {
     return data_ptr_<ArrayNodeStateData>(state)->diff();
 }
 
+template <class UnaryOp>
+bool UnaryOpNode<UnaryOp>::equal_to(const Node& rhs) const {
+    const auto* rhs_ptr = dynamic_cast<const UnaryOpNode*>(&rhs);
+    if (rhs_ptr == nullptr) return false;  // not same type so not equal
+    return this->equal_to(*rhs_ptr);
```
fastbodin: Confirming that I am not sleeping at the wheel. I had to do a double take here 😆

fastbodin (commented):
Managed to work through all files excluding the `test` files. Thought I would upload these comments before moving onto them.

dwave/optimization/include/dwave-optimization/nodes/binaryop.hpp
fastbodin: Thanks for this.

dwave/optimization/src/nodes/creation.cpp
arcondello: Technically my case can catch an issue that yours can't! If the size of `predcessors()` is larger than the size that `ssize_t` can encode.

Obviously an edge case that will never matter but...

dwave/optimization/include/dwave-optimization/nodes/manipulation.hpp
arcondello: "Contiguous" is the key word! The docs could be better. But there are times where you want a contiguous version of a strided array.

dwave/optimization/include/dwave-optimization/nodes/reduce.hpp
arcondello: Probably should. Was trying to minimize changes.

dwave/optimization/src/nodes/binaryop.cpp
arcondello: Yes exactly! I considered actually making `ArrayOutputMixin` hold an `std::vector<ArrayNode*> operands`. That would let us unify most of these but has downsides. The main one is that most nodes give a meaningful name to their `ArrayNode*`

dwave/optimization/src/nodes/creation.cpp
arcondello: Yeah, some of it was just me changing my pattern over time, sometimes it was to save lines of code, still others it just felt more natural. IMO consistency isn't terribly important.

dwave/optimization/src/nodes/indexing.cpp
arcondello: nicer formatting. IMO the extra comments wouldn't add much but 🤷

dwave/optimization/src/nodes/lambda.cpp
arcondello: yeah nicer formatting. Probably should mess with `.clang-format` some more.

dwave/optimization/src/nodes/linear_algebra.cpp
arcondello: I think better to add an `assert(index == 1)` in the `else` clause.

dwave/optimization/src/nodes/creation.cpp
arcondello: OTOH, such values would be out of bounds for the `index` parameter so it's kind of moot.

dwave/optimization/include/dwave-optimization/nodes/manipulation.hpp
fastbodin: Ah. Yes. I see. Cheers.

Resolved.

dwave/optimization/include/dwave-optimization/nodes/reduce.hpp
fastbodin: That is fine. I agree with that call. Happy to make a separate PR to go through the nodes and make those changes.

dwave/optimization/src/graph.cpp
arcondello: maybe it's just in my head but this separates the roots from the leaves, which I find helpful

dwave/optimization/src/graph.cpp
arcondello: agree it's redundant, though it's partly so that anyone coming here from https://github.com/dwavesystems/dwave-optimization/blob/96eee37ad819330d01dc9c58a8fd9a0cf002a83a/dwave/optimization/include/dwave-optimization/graph.hpp#L202-L209 would be able to see easily that everything is covered. I could also do it with a comment though

dwave/optimization/include/dwave-optimization/nodes/reduce.hpp
arcondello: We should be more consistent about `const` and non-`const` attributes in general. It's been bothering me for a while.

dwave/optimization/src/nodes/creation.cpp
fastbodin: I am just blindly enforcing the "we prefer `ssize_t` over `size_t`" advice I got in my first month. 😆 Given your latter point, id vote `ssize_t` 🙃 

Resolved.

dwave/optimization/src/nodes/binaryop.cpp
fastbodin: Glad I am following along. 

Resolved.

dwave/optimization/src/nodes/indexing.cpp
fastbodin: Agreed. I only included the "suggestion" because such comments were included for `ConstantNode::equal_to`. Was unsure if you want to be uniform 🤷 .

Resolved.

dwave/optimization/src/nodes/unaryop.cpp
arcondello: Probably should change 190 to
```c++
const UnaryOpNode* rhs_ptr = dynamic_cast<const UnaryOpNode*>(&rhs);
```
to make it a little clearer what's happening.

dwave/optimization/src/nodes/creation.cpp
fastbodin: > IMO consistency isn't terribly important.

🤯 . The floor just fell out from under me 🤣

dwave/optimization/src/nodes/linear_algebra.cpp
fastbodin: Oops. Agreed! I somehow missed that. 

Resolved.

dwave/optimization/src/nodes/creation.cpp
arcondello: Haha, fair. Maybe better to say I (mostly) used the one-cast pattern when I wanted to add something to the namespace that I could assert on. I used the local-cast pattern when not. I probably should do more asserts everywhere but that way lies madness for now.

dwave/optimization/src/graph.cpp
fastbodin: Please leave it then! 

..... get it 😄 

Resolved.

dwave/optimization/src/graph.cpp
fastbodin: Fine to keep then. Sometimes I leave extra `assert()`s for this reason too. Helpful for the reader to understand immediately without having to remember that this was checked elsewhere. 

Resolved.

dwave/optimization/include/dwave-optimization/nodes/reduce.hpp
fastbodin: Agreed. Separate PR. Happy to volunteer myself.

Resolved.

dwave/optimization/src/nodes/creation.cpp
fastbodin: Resolved.

dwave/optimization/src/nodes/creation.cpp
fastbodin: Well now I feel better. Agreed that there be madness.

Resolved.

dwave/optimization/src/nodes/lambda.cpp
fastbodin: Resolved.

dwave/optimization/src/nodes/unaryop.cpp
fastbodin: Usually I am fine with `auto` here, but, in this case, that would actually be helpful.

dwave/optimization/src/graph.cpp
arcondello: :deciduous_tree:

dwave/optimization/src/graph.cpp
arcondello: though now it's bothering me that they are not in the same order

tests/cpp/nodes/test_collections.cpp
```diff
@@ -49,6 +49,17 @@ TEST_CASE("DisjointBitSetsNode") {
                 graph.emplace_node<ArrayValidationNode>(sets.at(i));
             }
 
+            THEN("node equality works as expected") {
```
fastbodin, line 52: Test for `DisjointListNode` too?

tests/cpp/nodes/test_interpolation.cpp
```diff
@@ -114,6 +118,50 @@ TEST_CASE("BSpline") {
...
+        int k = 2;
+        std::vector<double> t0 = {0, 1, 2, 3, 4, 5, 6};
+        std::vector<double> t1 = {1, 1, 2, 3, 4, 5, 6};
+        std::vector<double> c0 = {-1, 2, 0, -1};
+        std::vector<double> c1 = {1, 2, 0, -1};
+
+        Node* a_ptr = graph.emplace_node<BSplineNode>(x0_ptr, k, t0, c0);
+        Node* b_ptr = graph.emplace_node<BSplineNode>(x0_ptr, k, t0, c0);
+        Node* c_ptr = graph.emplace_node<BSplineNode>(x1_ptr, k, t1, c0);
+        Node* d_ptr = graph.emplace_node<BSplineNode>(x0_ptr, k, t0, c1);
```
fastbodin, line 138: Just because `BSplineNode::equal_to()` also checks `k`, I would add 
```cpp
        Node* e_ptr = graph.emplace_node<BSplineNode>(x0_ptr, 3, t0, c0);
```
and the corresponding checks like

```cpp
CHECK(not a_ptr->equal_to(*e_ptr));

File: tests/cpp/nodes/test_manipulation.cpp [hunk 32]
```diff
@@ -1623,6 +1913,25 @@ TEST_CASE("SizeNode") {
             }
         }
     }
+
```
fastbodin, line 1916: I realize neither `SizeNode` nor `TransposeNode` has custom `equal_to()` since they get this directly from `EqualityMixin`. That said, is it still worth checking this case for each? <- Nope!

Agreed that it is fine here https://github.com/dwavesystems/dwave-optimization/pull/571#discussion_r3580803492

dwave/optimization/src/nodes/creation.cpp
arcondello: Thinking about this even more, I think 
```c++
assert(0 <= index and static_cast<size_t>(index) < predecessors().size());
```
is "more" correct. Basically the `0 <= index` check means that we can safely cast `index` to `size_t`. But with
```c++
assert(0 <= index and index < static_cast<ssize_t>(predecessors().size()));
```
the cast is not necessarily safe.

So even though it's kind of moot at the end of the day, I think the `size_t` way is slightly more correct.

tests/cpp/nodes/test_manipulation.cpp
```diff
@@ -1623,6 +1913,25 @@ TEST_CASE("SizeNode") {
...
+    SECTION("predecessor replacement") {
```
arcondello, line 1917: For some nodes that use the `EqualityMixin` without needing to implement another overload I stopped bothering to write separate tests for them, reasoning that it is already extremely well covered elsewhere.

dwave/optimization/src/nodes/creation.cpp
fastbodin: This logic makes sense. Consider me convinced.

dwave/optimization/_model.pyx
```diff
@@ -928,6 +928,46 @@ cdef class _Graph:
         """
         return self.num_nodes()
 
+    def remove_redundant_symbols(self, time_limit_s=None):
```
fastbodin, line 931: Worth adding a note in here that this is ineffective on named symbols?

tests/cpp/nodes/test_statistics.cpp
```diff
@@ -155,5 +157,38 @@ TEST_CASE("MeanNode") {
...
+        CHECK(not a_ptr->equal_to(*c_ptr));
+    }
+
+    SECTION("predecessor replacement") {
+        auto graph = Graph();
+
+        auto* c0_ptr = graph.emplace_node<ConstantNode>(std::vector{0, 1, 2, 3});
+        auto* c1_ptr = graph.emplace_node<ConstantNode>(std::vector{7, 6, 5, 4});
+
+        auto* argsort_ptr = graph.emplace_node<MeanNode>(c0_ptr);
```
fastbodin: `mean_ptr`

tests/cpp/nodes/test_unaryop.cpp
```diff
@@ -244,6 +244,37 @@ TEMPLATE_TEST_CASE(
...
+        CHECK(not a_ptr->equal_to(*c_ptr));
+    }
+
+    SECTION("predecessor replacement") {
+        auto graph = Graph();
+
+        auto* c0_ptr = graph.emplace_node<ConstantNode>(std::vector{0, 1, 2, 3});
+        auto* c1_ptr = graph.emplace_node<ConstantNode>(std::vector{7, 6, 5, 4});
+
+        auto* argsort_ptr = graph.emplace_node<UnaryOpNode<TestType>>(c0_ptr);
```
fastbodin: `unaryop_ptr`

tests/cpp/test_graph.cpp
```diff
@@ -316,6 +318,106 @@ TEST_CASE("Graph::objective()") {
...
 
+TEST_CASE("Graph::remove_redundant_nodes()") {
+    GIVEN("A model with two redundant nodes") {
+        auto graph = Graph();
+
+        auto* x_ptr = graph.emplace_node<BinaryNode>();
+        auto* y_ptr = graph.emplace_node<BinaryNode>();
+
+        auto* left_x_plus_y = graph.emplace_node<AddNode>(x_ptr, y_ptr);
+        auto* right_x_plus_y = graph.emplace_node<AddNode>(x_ptr, y_ptr);
```
fastbodin:
```suggestion
        auto* right_x_plus_y = graph.emplace_node<AddNode>(y_ptr, x_ptr);
```
As per the variable name?

fastbodin (commented):
I have now gone through all the files. I would like to go through `graph.cpp` again. 

The patience you had when writing these tests is a marvel.

tests/cpp/nodes/test_manipulation.cpp
fastbodin: That is fine with me. 👍  You were correct in predicting that I would make such a comment 🤣

dwave/optimization/_model.pyx
arcondello: I personally find it more readable this way, but agree we generally don't do it so will remove for consistency

tests/cpp/nodes/test_interpolation.cpp
arcondello: `k` interacts with the other parameters in such a way that you cannot vary only `k` (because it must be the sum of the degree) so technically that check is redundant.

Added a comment to note that.

dwave/optimization/_model.pyx
```diff
@@ -934,6 +934,9 @@ cdef class _Graph:
         Symbols are redundant if they are the same type, they share the
         same predecessors, and they encode the same operation.
 
+        If a symbol will not be removed if it has :class:`ArraySymbol` object(s)
```
fastbodin: I believe `If` is not necessary.

fastbodin (approved):
Baring the above typo, LGTM! 🥳 . It took me a minute because I wanted to go over `graph.cpp` again in some detail. 👍

PR #580, opened 2026-07-08
Add `type_list_union`

wbernoudy:
This PR implements `type_list_union` and `type_list_union_v`, which allow you to take the union of two or more `type_list`s to aggregate lists of types. It also implements `type_list::add` and `type_list::is_same` helper methods.

#### AI Generation Disclosure

No AI tools used

### DIFF
--- a/dwave/optimization/include/dwave-optimization/type_list.hpp
+++ b/dwave/optimization/include/dwave-optimization/type_list.hpp
@@ -48,6 +48,49 @@
 
 namespace dwave::optimization {
 
+template <typename Type, typename... Types>
+struct type_list;
+
+namespace {
+/// Helper struct for taking the union of two or more ``type_list``s
+template <typename type_list_x, typename type_list_y, typename... remaining_type_lists>
+struct type_list_union;
+
+template <typename type_list_x, typename type_list_y>
+struct type_list_union<type_list_x, type_list_y> {
+ private:
+    template <typename tl, typename type_tuple, typename U>
+    struct union_;
+
+    template <typename tl, typename type_tuple, std::size_t i>
+    struct union_<tl, type_tuple, std::index_sequence<i>> {
+        using value = tl::template add<std::tuple_element_t<i, type_tuple>>;
+    };
+
+    template <typename tl, typename type_tuple, std::size_t i, std::size_t... remaining_indices>
+    struct union_<tl, type_tuple, std::index_sequence<i, remaining_indices...>> {
+        using added = tl::template add<std::tuple_element_t<i, type_tuple>>;
+        using value = std::conditional_t<
+            sizeof...(remaining_indices) == 0,
+            added,
+            typename union_<added, type_tuple, std::index_sequence<remaining_indices...>>::value>;
+    };
+
+ public:
+    using value = union_<
+        type_list_x,
+        typename type_list_y::template to<std::tuple>,
+        std::make_index_sequence<type_list_y::size()>>::value;
+};
+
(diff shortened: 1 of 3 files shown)


dwave/optimization/include/dwave-optimization/type_list.hpp
```diff
@@ -96,6 +102,12 @@ struct type_list {
         return count;
     }
 
+    template <typename OtherTypeList>
+    static constexpr bool is_same() {
```
arcondello: Could we do this with a `constexpr bool operator==()`?

dwave/optimization/include/dwave-optimization/type_list.hpp
```diff
@@ -158,4 +170,41 @@ struct type_list {
     using to = U<Type, Types...>;
 };
 
+/// Take the union of two or more ``type_list``s, producing a new ``type_list``
+template <typename x, typename y, typename... zs>
+struct type_list_union;
```
arcondello: I'd rather make this a "method" like https://docs.python.org/3/library/stdtypes.html#set.union

dwave/optimization/include/dwave-optimization/type_list.hpp
wbernoudy: I don't believe so for a few reasons, but mainly that `operator==` cannot be a static member method. I was trying to follow `std::is_same` here roughly (which I believe was implemented with an explicit function rather than `==` for the same reason).

arcondello: The methods `issuperset`  and `issubset` are based on [the python methods](https://docs.python.org/3/library/stdtypes.html#frozenset.issubset). I don't love the mix of yes and no underscore :thinking:

dwave/optimization/include/dwave-optimization/type_list.hpp
wbernoudy: I can take a look at how this might work. I was trying to follow the more standard lib approach of having separate functions/structs (e.g. `std::tuple_cat`) but I also understand that the rest of `type_list` wasn't really written that way.

arcondello: Yeah, looking at it with fresh eyes I am not 100% sure why I chose to write it the way it is

wbernoudy: And `union` alone is a reserved keyword so I don't think we can have `type_list::union` :disappointed:

dwave/optimization/include/dwave-optimization/type_list.hpp
wbernoudy: Good point, happy to change to `issame`... or should I give `type_list` a rewrite in more standard lib style?

arcondello: I'd go with `equal()`

dwave/optimization/include/dwave-optimization/type_list.hpp
arcondello: Maybe [update](https://docs.python.org/3/library/stdtypes.html#set.update)?

releasenotes/notes/add-type_list_union-9465f17e156f5481.yaml
```diff
@@ -0,0 +1,7 @@
+---
+features:
+  - |
+    Add C++ ``type_list_union`` and ``type_list_union_v`` templated structs
```
arcondello: need to update this text too

arcondello (approved):

PR #582, opened 2026-07-10
Rework dtypes with additional concepts and fixes

### LINKED ISSUE #406: Incorrect format character
For a signed long long we should use `q`, not `Q`

https://github.com/dwavesystems/dwave-optimization/blob/7534f0966462ab6382c684bdc8e42d506c613c64/dwave/optimization/include/dwave-optimization/typing.hpp#L59

arcondello:
Closes #406 

#### AI Generation Disclosure
~No AI used.~ I asked ChatGPT to review the code, which resulted in https://github.com/dwavesystems/dwave-optimization/pull/582/commits/dc512a1c2c1a0e9281919dd5c45b5bc2b388aef9

### DIFF
--- a/dwave/optimization/include/dwave-optimization/typing.hpp
+++ b/dwave/optimization/include/dwave-optimization/typing.hpp
@@ -16,28 +16,50 @@
 
 #include <concepts>
 #include <cstdint>
+#include <type_traits>
+
+#include "dwave-optimization/type_list.hpp"
 
 namespace dwave::optimization {
 
+// We want to support an enumerated list of data types, corresponding to a
+// subset supported by NumPy.
+// See https://numpy.org/doc/stable/reference/arrays.dtypes.html for more
+// information.
+
+/// Supported dtypes as a type_list
+using DTypes = type_list<
+    float,          // np.float32
+    double,         // np.float64
+    bool,           // np.bool
+    std::int8_t,    // np.int8
+    std::int16_t,   // np.int16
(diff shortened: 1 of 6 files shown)


dwave/optimization/include/dwave-optimization/typing.hpp
```diff
@@ -56,21 +78,21 @@ enum class FormatCharacter : char {
     signedchar_ = 'b',
     signedshort_ = 'h',
     signedlong_ = 'l',
-    signedlonglong_ = 'Q',
+    signedlonglong_ = 'q',
```
arcondello, line 81: There isn't really a good test I can do for this because we don't currently expose this stuff to the Python level.

tests/cpp/test_typing.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+#include <utility>
+
+#include <catch2/catch_test_macros.hpp>
+
+#include "dwave-optimization/typing.hpp"
+
+namespace dwave::optimization {
+
+void forwarding_reference_func(DTypeLike auto&& lhs) { }
```
alexzucca90: I don't have anything smart to add other than why `lhs` that stands for left hand side?

alexzucca90 (approved):
LGTM

tests/cpp/test_typing.cpp
arcondello: ah, copy and paste error. Will fix

PR #586, opened 2026-07-17
Fix state management of root symbols

### LINKED ISSUE #584: Segfault when setting the state of an `Input` for an unlocked model
```python
from dwave.optimization import Model

model = Model()

inp = model.input((3,), lower_bound=0, upper_bound=10, integral=True)

model.states.resize(1)
inp.set_state(0, [5, 6, 4])
```
```bash
dwave-optimization/graph.hpp:370: void dwave::optimization::Node::emplace_data_ptr_(dwave::optimization::State&, Args&& ...) const [with StateData = dwave::optimization::ArrayNodeStateData; Args = {std::vector<double, std::allocator<double> >}; dwave::optimization::State = std::vector<std::unique_ptr<dwave::optimization::NodeStateData> >]: Assertion `index >= 0 and "must be topologically sorted"' failed
```

### LINKED ISSUE #585: Error when loading a model with some decision states uninitialized
```python
from dwave.optimization import Model

model = Model()

_, sets = model.disjoint_bit_sets(10, 5)
inp = model.input((3,), lower_bound=0, upper_bound=10, integral=True)

model.lock()

model.states.resize(1)

inp.set_state(0, [5, 6, 4])

with model.to_file(max_num_states=1) as f:
    Model.from_file(f)
```
```bash
KeyError: "There is no item named 'nodes/0/states/0/set0' in the archive"
```

arcondello:
Closes https://github.com/dwavesystems/dwave-optimization/issues/584
Closes https://github.com/dwavesystems/dwave-optimization/issues/585

#### AI Generation Disclosure
~No AI used.~ I did as Claude for a code review and https://github.com/dwavesystems/dwave-optimization/pull/586/commits/b3a58c4d2a3e39ead8d0432abb7ae0f182d135b8 was the result.


dwave/optimization/symbols/inputs.pyx
```diff
@@ -93,6 +93,11 @@ cdef class Input(ArraySymbol):
             ...     print(z.state(0))
             10.0
         """
+        if not self.model.is_locked() and self.node_ptr.topological_index() < 0:
+            raise TypeError(
```
arcondello, line 97: Honestly forget why this should be a `TypeError`, but it's consistent with `.state()` https://github.com/dwavesystems/dwave-optimization/blob/96eee37ad819330d01dc9c58a8fd9a0cf002a83a/dwave/optimization/_model.pyx#L1216

dwave/optimization/symbols/inputs.pyx
```diff
@@ -93,6 +93,11 @@ cdef class Input(ArraySymbol):
...
+                "the state of an intermediate variable cannot be set without "
```
smharwood: I know this is just mirroring other messages out there, but an `input` node is not really an "intermediate" variable- might be a little misleading
```suggestion
                "the state of a variable cannot be set without "
```

arcondello: good call. Probably "non-decision" or something like that

elham-azd:
LGTM

dwave/optimization/symbols/inputs.pyx
```diff
@@ -93,6 +93,11 @@ cdef class Input(ArraySymbol):
             ...     print(z.state(0))
             10.0
         """
+        if not self.model.is_locked() and self.node_ptr.topological_index() < 0:
```
alexzucca90, line 96: Can you have locked model and negative topological index? I thought as the model is locked topological index is built and basically never < 0. 
But I'm probably missing something

arcondello: decisions have a non-negative topological index even when the model is unlocked. So this is basically checking that we're not a decision (albeit in a slightly more future-proof way).

alexzucca90: oh right!

alexzucca90 (approved):

PR #587, opened 2026-07-17
Implement state checkpointing for decision nodes

### LINKED ISSUE #510: Add checkpointing method(s) for decision array nodes

arcondello:
Closes https://github.com/dwavesystems/dwave-optimization/issues/510
Closes https://github.com/dwavesystems/dwave-optimization/pull/552 by replacing it

#### AI Generation Disclosure
No AI was used to write the code. I intend to use it to review this PR.


dwave/optimization/src/nodes/collections.cpp
```diff
@@ -213,8 +303,52 @@ class CollectionStateData_ : public NodeStateData {
...
     ssize_t size_;
     ssize_t previous_size_;
+
+    friend CollectionCheckpoint_;
+    CollectionCheckpoint_* older_checkpoint_ptr_ = nullptr;
 };
 
+CollectionCheckpoint_::CollectionCheckpoint_(CollectionStateData_* state_ptr) :
+    updates_(),
+    drop_(state_ptr->all_updates_.size()),  // so we ignore any updates added before we're made
```
wbernoudy: What happens if some mutations are made to the node, then you make a checkpoint (causing the `drop_` to be set > 0), and then you revert, make more mutations, and commit? Won't this cause some updates to be "dropped" that are actually relevant to the checkpoint?

wbernoudy: Actually, I think I mean "make another checkpoint" as final step in the example rather than commit.

arcondello: Yup! Good catch. Fixed in https://github.com/dwavesystems/dwave-optimization/pull/587/commits/214c5e23952649ba5adb24e8c74f3bcffe55e9e2

dwave/optimization/include/dwave-optimization/state.hpp
```diff
@@ -36,4 +36,20 @@ struct NodeStateData {
...
+    NodeStateCheckpoint() = default;
+    NodeStateCheckpoint(const NodeStateCheckpoint&) = default;
+    NodeStateCheckpoint(NodeStateCheckpoint&&) = delete;
+    NodeStateCheckpoint& operator=(const NodeStateCheckpoint&) = default;
+    NodeStateCheckpoint& operator=(NodeStateCheckpoint&&) = delete;
+
+    virtual ~NodeStateCheckpoint() = default;
+
+    /// Whether the checkpoint is still available to be used.
+    virtual bool valid() const = 0;
```
arcondello: This is unused and I should drop it.

dwave/optimization/include/dwave-optimization/nodes/testing.hpp
```diff
@@ -105,6 +100,28 @@ class DynamicArrayTestingNode : public ArrayOutputMixin<ArrayNode>, public Decis
     void revert(State&) const override;
     void update(State&, int) const override;
 
+    // Overloads required by the DecisionNode ABC *****************************
+
+    // DynamicArrayTestingNode does not impement checkpointing
```
arcondello: See https://github.com/dwavesystems/dwave-optimization/issues/593

Also I see the typo here.

dwave/optimization/src/graph.cpp
```diff
@@ -196,6 +195,11 @@ void Graph::propose(
     }
 }
 
+void Graph::propose(State& state) const {
```
arcondello, line 233: Could be smarter with https://github.com/dwavesystems/dwave-optimization/pull/589 or https://github.com/dwavesystems/dwave-optimization/pull/592, but that's for another PR.

tests/cpp/nodes/test_collections.cpp
```diff
@@ -725,6 +790,326 @@ TEST_CASE("SetNode") {
...
+            }
+
+            THEN("We can commit, mutate, then revert") {
+                graph.propose(state);
+
+                set_ptr->exchange(state, 1, 2);
+
+                set_ptr->assign_from_checkpoint(state, std::move(checkpoint0));
+
+                // CHECK_THAT(set_ptr->view(state), RangeEquals({0, 1}));
```
arcondello: oops, will restore

tests/cpp/nodes/test_collections.cpp
```diff
@@ -725,6 +790,326 @@ TEST_CASE("SetNode") {
...
+                AND_WHEN("We revert to the second checkpoint") {
+                    set_ptr->assign_from_checkpoint(state, std::move(checkpoint1));
+                    graph.propose(state);
+
+                    CHECK_THAT(set_ptr->view(state), RangeEquals({3, 2, 1, 0}));
+                }
+            }
+        }
+
+        // TODO: within one propagation
```
arcondello: this is done, need to remove this comment

tests/cpp/nodes/test_numbers.cpp
```diff
@@ -294,6 +295,18 @@ TEST_CASE("BinaryNode") {
                     CHECK(static_cast<int>(ptr->diff(state).size()) == 2 * exchange_count_ground);
                 }
             }
+
+            AND_WHEN("We create a checkpoint to that state") {
+                auto checkpoint = ptr->checkpoint(state);  // 0, 1, 0, 1, 0, 1, 0, 1, 0, 1
+
+                THEN("We can mutate and then assign from that checkpoint") {
+                    ptr->set_value(state, 0, 1);  // 0, 1, 0, 1, 0, 1, 0, 1, 0, 1
```
fastbodin:
```cpp
// 1, 1, 0, 1, 0, 1, 0, 1, 0, 1
```
since you called `set_value()`

tests/cpp/nodes/test_numbers.cpp
```diff
@@ -1985,6 +2000,55 @@ TEST_CASE("IntegerNode") {
...
                 }
             }
+
+            AND_WHEN("We checkpoint the state and then mutate") {
+                auto checkpoint = ptr->checkpoint(state);  // [-4, -4, -2, -2, 0, 0, 2, 2, 4, 4]
+
+                ptr->exchange(state, 0, 2);   // [-2, -4, -4, -2, 0, 0, 2, 2, 4, 4]
+                ptr->set_value(state, 3, 1);  // [-2, -4, -4, 1, 0, 0, 2, 2, 4, 4]
+
+                THEN("We can commit, then assign from the checkpoint") {
```
fastbodin, line 2010: Nit:
```cpp
                THEN("We can propose, then assign from the checkpoint") {
```
I realize `propose` is `propagate` and `commit`. As this is a Nit, I am not going to comment at all duplicate instances.

tests/cpp/nodes/test_numbers.cpp
```diff
@@ -294,6 +295,18 @@ TEST_CASE("BinaryNode") {
                     CHECK(static_cast<int>(ptr->diff(state).size()) == 2 * exchange_count_ground);
                 }
             }
+
+            AND_WHEN("We create a checkpoint to that state") {
+                auto checkpoint = ptr->checkpoint(state);  // 0, 1, 0, 1, 0, 1, 0, 1, 0, 1
+
+                THEN("We can mutate and then assign from that checkpoint") {
```
fastbodin: Nit
```cpp
                THEN("We can mutate, then propose, and then assign from that checkpoint") {
```

tests/cpp/nodes/test_numbers.cpp
```diff
@@ -2209,14 +2275,110 @@ TEST_CASE("IntegerNode") {
...
+                        inode_ptr->assign_from_checkpoint(state, std::move(checkpoint));
+
+                        sum_constraints_lhs = inode_ptr->sum_constraints_lhs(state);
+                        CHECK(sum_constraints_lhs.size() == 1);
+                        CHECK(sum_constraints_lhs.data()[0].size() == 1);
+                        CHECK_THAT(sum_constraints_lhs[0], RangeEquals({48}));
+                        CHECK_THAT(inode_ptr->view(state), RangeEquals({8, 8, 8, 8, 8, 8, -3, 3}));
+                    }
+                }
+            }
```
fastbodin, line 2381: A bit of a pain, but I think it might be worth adding a single test for `BinaryNode` with `checkpointing` to check that all the correct indices are cached with `1`s and `0`s (I am referring to the `BinaryNode` `DisjointSparseSet`).

```cpp
THEN("Sum constraint sums, tracked indices, and state are correct") {
    CHECK(bnode_ptr->sum_constraints_lhs(state).size() == 1);
    CHECK(bnode_ptr->sum_constraints_lhs(state).data()[0].size() == 1);
    CHECK_THAT(bnode_ptr->sum_constraints_lhs(state)[0], RangeEquals({3}));
    check_indices<true>(state, bnode_ptr, 0, 0, {1, 2, 5}); // <- ***** This check is the important one ******
    check_indices<false>(state, bnode_ptr, 0, 0, {0, 3, 4}); // <- ***** And this one ******
    CHECK_THAT(bnode_ptr->view(state), RangeEquals(expected_init));
}
```

dwave/optimization/src/nodes/numbers.cpp
```diff
@@ -72,8 +73,79 @@ NumberNode::SumConstraint::Operator NumberNode::SumConstraint::op(const ssize_t
     return operators_[slice];
 }
 
+class NumberNodeCheckpoint_ : public DiffCheckpoint {
+ public:
+    using slice_cache_type = std::vector<std::vector<ssize_t>>;
```
fastbodin, line 78: Wishing I had used this elsewhere 😂

dwave/optimization/src/nodes/numbers.cpp
```diff
@@ -142,10 +228,16 @@ void NumberNodeStateData::revert() {
...
 
-    ArrayNodeStateData::revert();  // Revert changes to the buffer.
+    if (auto* checkpoint_ptr = this->checkpoint_ptr<NumberNodeCheckpoint_>()) {
+        checkpoint_ptr->revert_updates(
+            ArrayNodeStateData::revert_and_detach(), std::move(slice_cache_)
+        );
+    } else {
+        slice_cache_.clear();          // Empty the slice cache.
+        ArrayNodeStateData::revert();  // Revert changes to the buffer.
+    }
```
fastbodin, line 313: Worth adding 
```cpp
        // everything should have been cleared out regardless of which path we took
        assert(this->diff().empty());
        assert(slice_cache_.empty());
```
like in `commit()`?

dwave/optimization/src/nodes/numbers.cpp
```diff
@@ -953,6 +1049,56 @@ IntegerNode::IntegerNode(
         std::move(sum_constraints)
     ) {}
 
+void IntegerNode::assign_from_checkpoint(State& state, checkpoint_type& checkpoint) const {
+    auto state_data = data_ptr_<NumberNodeStateData>(state);
+
+    auto* checkpoint_ptr = static_cast<NumberNodeCheckpoint_*>(checkpoint.get());
+
+    // todo: assert that this checkpoint is the latest
```
fastbodin: Hanging todo?

dwave/optimization/src/nodes/numbers.cpp
```diff
@@ -953,6 +1049,56 @@ IntegerNode::IntegerNode(
...
+    // A better way would be to implement a partial revert on our state class, but this
+    // is not a path we care about greatly so let's err on the side of simple and well-
+    // tested.
+    if (ssize_t excess_updates = state_data->diff().size() - checkpoint_ptr->drop()) {
+        assert(excess_updates > 0);
+        for (
+            const auto& [idx, old, _] :
+            state_data->diff() | std::views::reverse | std::views::take(excess_updates)
+        ) {
+            state_data->set(idx, old);
```
fastbodin: I would just add a note that this by-passes (currently) the sum constraint `lhs` tracking.

dwave/optimization/src/nodes/_state.hpp
```diff
@@ -91,6 +91,22 @@ class ArrayStateData {
         assert(size_ >= 0 && static_cast<std::size_t>(size_) == buffer.size());
     }
 
+    // Commit the changes and clear the diff by returning the diff buffer.
```
fastbodin, line 94: Confused where the `commit()` is and am feeling like I am missing something.

dwave/optimization/src/nodes/_state.hpp
```diff
@@ -149,6 +165,29 @@ class ArrayStateData {
         size_ = buffer.size();
     }
 
+    // Commit the changes and clear the diff by returning the diff buffer.
```
fastbodin: Same doc string comments as `commit_and_detach()`. Intentional?

dwave/optimization/src/nodes/_checkpoints.cpp
```diff
@@ -0,0 +1,93 @@
...
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#include "_checkpoints.hpp"
+
+namespace dwave::optimization {
+
+CheckpointableState::~CheckpointableState() {
+    if (prev_ptr_ == nullptr) return;  // nothing to clean up
+    prev_ptr_->next_ptr_ = static_cast<CheckpointableState*>(nullptr);
```
fastbodin, line 21: Checking my understanding: if we are destructing this `CheckpointableState`, then the previously linked state (at `prev_ptr_`) that points to this one now points to `nullptr`? i.e. the previous `CheckpointableState` in the linked-list is now the end of the list.

dwave/optimization/src/nodes/_checkpoints.cpp
```diff
@@ -0,0 +1,93 @@
...
+}
+
+// Place self between the state and any checkpoint it's currently holding
+LinkedListCheckpoint::LinkedListCheckpoint(CheckpointableState& state) :
+    prev_ptr_(state.prev_ptr_), next_ptr_(&state) {
+    if (prev_ptr_ != nullptr) prev_ptr_->next_ptr_ = this;
+    state.prev_ptr_ = this;
+}
+
+LinkedListCheckpoint::~LinkedListCheckpoint() {
```
fastbodin, line 31: Testing my understanding. Given `A -> B -> C`, if the destructor was called on `B`, this results in `A->C`?

arcondello: Yes that's right

dwave/optimization/src/nodes/_checkpoints.cpp
arcondello: Yes that's right. This one is an edge case because the list is now invalid in some sense, but different destruction patterns can cause it to happen temporarily.

dwave/optimization/src/nodes/_checkpoints.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+namespace dwave::optimization {
+
+class LinkedListCheckpoint;
+
+class CheckpointableState {
+ public:
+    CheckpointableState() = default;
+
+    CheckpointableState(const CheckpointableState&) {}  // the checkpoint pointer is not copied
```
fastbodin: Can you explain why this is not simply
```cpp
CheckpointableState(const CheckpointableState&) = delete
```
It is unclear why we would use `void` copy constructors?

dwave/optimization/src/nodes/_checkpoints.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+DiffCheckpoint::DiffCheckpoint(CheckpointableState& state, ssize_t drop) :
+    LinkedListCheckpoint(state), updates_(), drop_(drop) {}
+
+DiffCheckpoint::DiffCheckpoint(CheckpointableState& state, std::span<const Update> diff) :
+    LinkedListCheckpoint(state), updates_(), drop_(diff.size()) {
+    if (auto* prev_ptr = static_cast<DiffCheckpoint*>(prev_ptr_)) {
+        prev_ptr->commit_updates(std::vector<Update>(diff.begin(), diff.end()));
+        assert(prev_ptr->drop_ == 0);
+    }
```
fastbodin, line 52: To be clear, we are doing the `static_cast` because the `prev_ptr_` member is a `LinkedListCheckpoint` which does not have the method `commit_updates()`. Functionally, the only reason the `static_cast` would return a `nullptr` is if `prev_ptr_` is a `nullptr` already. 

I am just comparing this to the destructor which checks `if (prev_ptr_ == nullptr)` and does not have a `if` around `static_cast<DiffCheckpoint*>(prev_ptr_)`.  It feels like this should either be

```cpp
    if (auto* prev_ptr = static_cast<DiffCheckpoint*>(prev_ptr_)) {
        prev_ptr->commit_updates(std::vector<Update>(diff.begin(), diff.end()));
        assert(prev_ptr->drop_ == 0);
        return;
    }
    assert(prev_ptr_ == nullptr);
``` 
or should resemble the destructor
```cpp
    if (prev_ptr_ == nullptr) return;

    auto* prev_ptr = static_cast<DiffCheckpoint*>(prev_ptr_);
    assert(prev_ptr); // <- I added this
    prev_ptr->commit_updates(std::vector<Update>(diff.begin(), diff.end()));
    assert(prev_ptr->drop_ == 0);
```

dwave/optimization/src/nodes/_checkpoints.cpp
```diff
@@ -0,0 +1,93 @@
...
+}
+
+DiffCheckpoint::~DiffCheckpoint() {
+    // if we're the oldest checkpoint, just let whatever information we're
+    // holding get destructed with us
+    if (prev_ptr_ == nullptr) return;
+
+    // otherwise we need to transfer our info over
+    auto* prev_ptr = static_cast<DiffCheckpoint*>(prev_ptr_);
```
fastbodin: See comment above. If we keep this structure, I would consider adding
```cpp
assert(prev_ptr);
```

dwave/optimization/src/nodes/_checkpoints.cpp
```diff
@@ -0,0 +1,93 @@
...
+    assert(prev_ptr->drop_ == 0);
+    for (auto& updates : updates_) prev_ptr->commit_updates(std::move(updates));
```
fastbodin: The reason we do not do
```cpp
prev_ptr->commit_updates(this->detach_updates());
```
is because `drop_` (which I would love a doc string for 😄) says how many of the `vectors` in `vector<vector<Updates>>` we drop?

arcondello: how many updates of the last vector. And agree `drop_` needs a docstring, @alexzucca90 had the same feedback

dwave/optimization/src/nodes/_checkpoints.hpp
arcondello: Adding `= delete` would preclude inheriting classes from having copy constructors which we don't want. This way we allow copying, but don't copy the checkpoint. Useful for `NodeStateData::copy()`.

dwave/optimization/src/nodes/_state.hpp
arcondello: Should say "revert"

dwave/optimization/src/nodes/_state.hpp
arcondello: Commiting the changes just means keeping the status quo 😄

dwave/optimization/src/nodes/_checkpoints.cpp
fastbodin: The `last vector`! Okay. In that case, why do the loop as opposed to 
```cpp
prev_ptr->commit_updates(this->detach_updates());
```

dwave/optimization/src/nodes/_checkpoints.hpp
fastbodin: Thank you!

dwave/optimization/src/nodes/_checkpoints.cpp
arcondello: `prev_ptr->commit_updates()` takes a single diff (usually from the node state). Here we want to transfer a sequence of diffs. We could flatten, or add an overload I suppose..

arcondello:
I think I need to take another documentation pass. I'll try to do that today

dwave/optimization/src/nodes/_checkpoints.cpp
fastbodin: It does not make a difference really to me. Just trying to understand. I had thought that
```cpp
    auto detach_updates() {
        auto updates = std::move(updates_) | std::views::join;
        assert(updates_.empty());
        return updates;
    }
```
would result in a `std::vector<Update>` since `updates_` is a `std::vector<std::vector<Update>>`. I had not thought you would need an overload?

dwave/optimization/src/nodes/_state.hpp
fastbodin: Wait. Given a committed `state`, if we then make a series of changes (and create a `diff`) this method detaches the `diff`. Is this considered a silent `commit()`? The `state` is still modified but now we don't have the `diff` to now how it was modified?

dwave/optimization/src/nodes/collections.cpp
```diff
@@ -109,11 +124,68 @@ class CollectionStateData_ : public NodeStateData {
...
     }
 
+    void assign(std::unique_ptr<NodeStateCheckpoint>& checkpoint) {
+        // convert the checkpoint into something we can read
+        auto* checkpoint_ptr = static_cast<CollectionCheckpoint_*>(checkpoint.get());
+
+        // Right now, you can only revert to the most recent checkpoint. It's
+        // pretty straightforward to support going further back, but this is all
+        // we need right now.
+        assert(this->checkpoint_ptr<CollectionCheckpoint_>() == checkpoint_ptr);
```
fastbodin, line 134: Took me a hot minute to realize this was templated from
```cpp
class CheckpointableState {
 protected:
    template <std::derived_from<LinkedListCheckpoint> T>
    T* checkpoint_ptr() {
        return static_cast<T*>(prev_ptr_);
    }
```

dwave/optimization/src/nodes/collections.cpp
```diff
@@ -329,7 +440,17 @@ ssize_t CollectionNode::size_diff(const State& state) const {
     return data_ptr_<CollectionStateData_>(state)->size_diff();
 }
 
-struct DisjointBitSetsNodeData_ : NodeStateData {
+// DisjointBitSetsNode is on the way out, so let's do the simplest possible
+// implementation for now.
```
fastbodin, line 444: Agreed.

dwave/optimization/src/nodes/collections.cpp
```diff
@@ -445,6 +581,28 @@ void DisjointBitSetsNode::initialize_state(
     );
 }
 
+void DisjointBitSetsNode::assign_from_checkpoint(
+    State& state,
+    std::unique_ptr<NodeStateCheckpoint>& checkpoint
+) const {
+    const DisjointBitSetsCheckpoint_* checkpoint_ptr =
```
fastbodin: `const auto*`?

tests/cpp/nodes/test_collections.cpp
```diff
@@ -725,6 +790,326 @@ TEST_CASE("SetNode") {
...
+                        if (flip(rng)) {
+                            graph.commit(state);
+                        } else {
+                            graph.revert(state);
+                        }
+                    } else {
+                        check.reset();
+                    }
+                }
+            }
```
fastbodin, line 1130: I love this test and wish there was a final `CHECK()` we could do :rofl

dwave/optimization/src/nodes/numbers.cpp
fastbodin: Testing my understanding. You mean
```cpp
        // Right now, you can only revert to the most recent checkpoint. It's
        // pretty straightforward to support going further back, but this is all
        // we need right now.
        assert(this->checkpoint_ptr<NumberNodeCheckpoint_>() == checkpoint_ptr);
```

fastbodin (commented):
Officially gone through everything. I need to look at the `number.cpp` and `collections.cpp` again. On the whole, looks good. Most of my comments above are clarifications. I will resolved them ASAP to avoid clogging the thread on this PR.

dwave/optimization/src/nodes/_checkpoints.hpp
```diff
@@ -0,0 +1,126 @@
...
+    auto detach_updates() {
+        auto updates = std::move(updates_) | std::views::join;
+        assert(updates_.empty());
+        return updates;
+    }
+
+    // The current "drop". The drop is used when a checkpoint is created while
+    // a node has some mutations already applied. This tells the checkpoint
+    // how to handle the diff associated with those mutations, i.e., the ones
+    // the checkpoint shouldn't be tracking.
```
fastbodin, line 111: Nit: This doc string is good but omits some information. Possible alternative (that could be improved) based on my (possibly incorrect) understanding of `drop`.

```cpp
// The current `drop` where `drop` is equal to the number of `Updates` 
// in the last `diff` cached by `updates_` that the checkpoint 
// is NOT responsible for tracking. Functionally, there are two cases. 
// 1. `drop == 0` and the checkpoint is responsible for all `Updates` 
//     stored in the last `diff` of `updates_`.
// 2. `drop > 0` and the checkpoint is *not* responsible for the last
//    `drop` number of `Updates` in the the last `diff` of `updates_`.

fastbodin: I just realized the above was in-fact, incorrect. Will try again.

dwave/optimization/src/nodes/numbers.cpp
fastbodin: Looking at this again, this is a bug. This does not take into account the running `slice` sums (should they be tracked). This was not caught in the tests because the relevant tests use `exchange()` which does not affect the running `slice` sums`.

dwave/optimization/src/nodes/_checkpoints.cpp
arcondello: `static_cast` is not like `dynamic_cast`, the only way this would be `nullptr` is if `prev_ptr_ == nullptr`. Anyway I went with the approach of putting them together in an `if`.

dwave/optimization/src/nodes/_checkpoints.cpp
arcondello: No it's returning whatever is returned by `std::views::join` which is not really a visible type.

dwave/optimization/src/nodes/_state.hpp
arcondello: Either this or `::commit()` should be used. They have the same meaning. The difference is that this one returns the `diff` to the caller, whereas `::commit()` does not. The reason you wouldn't always use `::commit_and_detach()` is by detaching the `updates_` vector it means the `updates_` is now empty (i.e. has no space allocated). Whereas `updates_.clear()` leaves the memory allocated (usually).

dwave/optimization/src/nodes/_checkpoints.cpp
fastbodin: Gotcha. Thanks for clarifying.

tests/cpp/nodes/test_numbers.cpp
arcondello: It seems that `BinaryNode::assign_checkpoint()` is very broken because it has a different update method. I think we need to refactor the various state data classes to use virtual methods.

DeezDevelopes (approved):

PR #589, opened 2026-07-22
Make `State` a class and add `Graph::mutated()`

wbernoudy:
This PR adds the `Graph::mutated()` method which returns the list of decision nodes with "pending" changes (see docstring for more details). It also makes `State` a class rather than an alias to a vector of data pointers in order to help implement `Graph::mutated()`.

#### Additional information
For now, I left the implementation of `Graph::mutated()` simple, and it will return only decision nodes, and not specific successors of `DisjointListsNode` and `DisjointBitSetsNode`. However, if we plan to use this directly with `descendants()`/`propagate()` etc., we may need to consider returning specific successors because of the known performance hit on models with large amounts of lists/sets on individual `DisjointListsNode`/`DisjointBitSetsNode`s.

#### AI Generation Disclosure
No AI tools used

### DIFF
--- a/dwave/optimization/include/dwave-optimization/graph.hpp
+++ b/dwave/optimization/include/dwave-optimization/graph.hpp
@@ -40,8 +40,6 @@ class Node;
 struct Decision {};
 
 class Graph {
-    friend class State;
-
  public:
     Graph() noexcept = default;
     ~Graph() noexcept = default;
--- a/dwave/optimization/include/dwave-optimization/graph.hpp
+++ b/dwave/optimization/include/dwave-optimization/graph.hpp
@@ -406,7 +406,7 @@ class Node {
     StateData* data_ptr_(State& state) const {
         const ssize_t index = topological_index();
         assert(index >= 0 and "must be topologically sorted");
-        assert(state.size() > static_cast<std::size_t>(index) and "unexpected state length");
+        assert(state.size() > index and "unexpected state length");
         assert(state[index] != nullptr and "uninitialized state");
 
         return static_cast<StateData*>(state[index].get());
@@ -415,7 +415,7 @@ class Node {
     const StateData* data_ptr_(const State& state) const {
         const ssize_t index = topological_index();
         assert(index >= 0 and "must be topologically sorted");
-        assert(state.size() > static_cast<std::size_t>(index) and "unexpected state length");
+        assert(state.size() > index and "unexpected state length");
         assert(state[index] != nullptr and "uninitialized state");
 
         return static_cast<const StateData*>(state[index].get());
@@ -425,7 +425,7 @@ class Node {
     void emplace_data_ptr_(State& state, Args&&... args) const {
         const ssize_t index = topological_index();
         assert(index >= 0 and "must be topologically sorted");
-        assert(state.size() > static_cast<std::size_t>(index) and "unexpected state length");
+        assert(state.size() > index and "unexpected state length");
         assert(state[index] == nullptr and "already initialized state");
 
         state[index] = std::make_unique<StateData>(std::forward<Args&&>(args)...);
--- a/dwave/optimization/include/dwave-optimization/graph.hpp
+++ b/dwave/optimization/include/dwave-optimization/graph.hpp
@@ -40,6 +40,8 @@ class Node;
 struct Decision {};
 
 class Graph {
+    friend class State;
+
  public:
     Graph() noexcept = default;
     ~Graph() noexcept = default;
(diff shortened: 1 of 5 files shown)


dwave/optimization/include/dwave-optimization/graph.hpp
```diff
@@ -40,6 +40,8 @@ class Node;
 struct Decision {};
 
 class Graph {
+    friend class State;
```
arcondello, line 43: I don't think we need this?

dwave/optimization/include/dwave-optimization/state.hpp
```diff
@@ -34,6 +34,32 @@ struct NodeStateData {
...
+    auto& operator[](index_type index) {
+        return node_data_[index];
+    }
+
+    template <typename index_type>
+    auto& operator[](index_type index) const {
+        return node_data_[index];
+    }
+
+    void resize(ssize_t size) { node_data_.resize(size); }
```
arcondello, line 56: Since we're granting friendship to `Graph`, do we want to make this private?

dwave/optimization/src/graph.cpp
```diff
@@ -156,6 +157,39 @@ void Graph::initialize_state(State& state) {
     static_cast<const Graph*>(this)->initialize_state(state);
 }
 
+std::span<const DecisionNode*> Graph::mutated(State& state) const {
+    // We will want to eventually replace this implementation with an approach where
```
arcondello, line 161: IMO we want to do this as part of this PR. Probably means keeping a boolean array of flags as well to determine which decisions have already been added?

dwave/optimization/include/dwave-optimization/state.hpp
```diff
@@ -34,6 +34,32 @@ struct NodeStateData {
...
 
-using State = typename std::vector<std::unique_ptr<NodeStateData>>;
+// Foward declaration for storing the mutated decision nodes on State
+class DecisionNode;
+
+class State {
+    friend class Graph;
+
+ public:
+    State(ssize_t size = 0) : node_data_(size) {}
```
arcondello, line 44: Do we want to make the constructors private since we're granting friendship to `Graph`? We might want to leave the default (empty) constructor public for Cython reasons.

dwave/optimization/include/dwave-optimization/state.hpp
```diff
@@ -34,6 +34,32 @@ struct NodeStateData {
...
+    }
+
+    template <typename index_type>
+    auto& operator[](index_type index) const {
+        return node_data_[index];
+    }
+
+    void resize(ssize_t size) { node_data_.resize(size); }
+
+    ssize_t size() const { return node_data_.size(); }
```
arcondello, line 58: Need to include `dwave-optimization/common.hpp` for `ssize_t`, this will fix the current windows CI failures.

dwave/optimization/include/dwave-optimization/state.hpp
wbernoudy: Yes, makes sense to me, I will split the empty constructor out.

dwave/optimization/include/dwave-optimization/state.hpp
wbernoudy: Looks like this is used by the cython as well (e.g. https://github.com/dwavesystems/dwave-optimization/blob/master/dwave/optimization/states.pyx#L305). Perhaps we could change the behavior of `Graph::initialize_state(State&)` to resize the state if the given size is zero?

arcondello: Ok, I am fine to leave this here then.

dwave/optimization/src/graph.cpp
```diff
@@ -157,6 +158,39 @@ void Graph::initialize_state(State& state) {
...
+    // decision nodes "eagerly" add themselves to the list of mutated nodes after they
+    // are mutated. This will avoid the need to iterate over all decision nodes every
+    // time this method is called.
+    state.mutated_nodes_.clear();
+
+    for (const DecisionNode* dec_ptr : decisions()) {
```
arcondello, line 168: I think adding a generic `Node::updated()` method (see https://github.com/dwavesystems/dwave-optimization/pull/596) would help a lot here. I can pull that feature out of #596 and we can rebase this? Or I guess we can merge this as-is and migrate later.

arcondello: Thinking about this more, it's not clear what the expected behaviour of `updated()` is. As written it'll be `true` before propagation for decisions, but for all other nodes it'll only be true after `::propagate()` has been called on them... So [never is often better than *right* now.](https://peps.python.org/pep-0020/). We can do it as a followup maybe.

arcondello (approved):

wbernoudy:
May come back to this later but closing for now.

PR #597, opened 2026-07-29
Fix typing issues in a few .py/.pyi files

arcondello:
Fix some (but not all) typing errors in a few of the Python/stub files. This is not meant to be complete but rather to just move us in the right direction.

#### AI Generation Disclosure
No AI used.


dwave/optimization/mathematical.py
```diff
@@ -731,7 +730,12 @@ def broadcast_to(x: ArraySymbol, shape: tuple[int, ...] | int) -> ArraySymbol:
     return BroadcastTo(x, shape)
 
 
-def bspline(x: ArraySymbol, k: int, t: list, c: list) -> ArraySymbol:
+def bspline(
+    x: ArraySymbol,
+    k: int,
+    t: np.typing.ArrayLike,
+    c: np.typing.ArrayLike,
```
JoelPasvolsky, line 737: Shouldn't these be``list | np.typing.ArrayLike``?

Note that line 751/755 uses a ``list`` rather than a NumPy array

JoelPasvolsky (approved):

dwave/optimization/mathematical.py
arcondello: Good news! A `list` is an `np.typing.ArrayLike`! You might be thinking of `np.typing.NDArray`?

JoelPasvolsky: Of course, yes, thanks!

PR #598, opened 2026-08-04
Don't serialize the states of intermediate variables

arcondello:
This is a backwards compatibility breaking change.

Don't serialize the states of intermediate variables. This simplifies some aspects of serialization significantly. Models serialized before this change will have the states of intermediate variables ignored.

#### AI Generation Disclosure
No AI used.


smharwood:
I've tried serializing and loading a model with different versions and they all act as expected.

Though, I keep wondering if there is a test we can add - for instance, commit a small model to the repo that was serialized with the old behavior and make sure it can be loaded with the latest version?

arcondello:
Closing in favour of https://github.com/dwavesystems/dwave-optimization/pull/611

PR #600, opened 2026-08-04
Add Graph::pop_decision() and Graph::swap_decisions() methods

arcondello:
#### AI Generation Disclosure
No AI used.

### DIFF
--- a/dwave/optimization/src/graph.cpp
+++ b/dwave/optimization/src/graph.cpp
@@ -159,19 +159,36 @@ void Graph::initialize_state(State& state) {
 
 void Graph::pop_decision() {
     assert(not topologically_sorted_ and "cannot pop a decision from a locked model");
+    assert(not decisions_.empty() and "need at least one decision");
 
-    // Get the index of the node we're going to delete
-    const ssize_t i = std::ranges::ssize(decisions_) - 1;
-    assert(0 <= i and "need at least one decision");
+    // Get a pointer to the node we want to remove
+    const Node* target_ptr = decisions_.back();
 
-    // Check that nodes_ and decisions_ are consistent (should always be true)
-    assert(nodes_[i].get() == decisions_[i]);
+    [[maybe_unused]] auto is_target = [&target_ptr](const auto* ptr) {
+        return static_cast<const Node*>(ptr) == target_ptr;
+    };
 
-    // Confirm that removing the decision won't leave anything dangling
-    assert(decisions_[i]->successors().empty() and "cannot remove a decision with successors");
+    // Make sure the last decision is unused
+    assert(target_ptr->successors().empty() and "cannot remove a decision with successors");
+    assert(not is_target(objective_ptr_) and "cannot remove the objective");
+    assert(std::ranges::none_of(constraints_, is_target) and "cannot remove a constraint");
 
+    // Ok, stop tracking our target in the decisions_ list
     decisions_.pop_back();
-    nodes_.erase(nodes_.begin() + i);
+
+    // Should never have the same decision twice and decisions are not inputs/constants
+    assert(std::ranges::none_of(decisions_, is_target));
+    assert(std::ranges::none_of(inputs_, is_target));
+    assert(std::ranges::none_of(constants_, is_target));
+
+    // Finally, we need to remove it from our node list. We do the swap and pop trick
+    // because we're not topologically sorted so it's OK for us to mess with the node order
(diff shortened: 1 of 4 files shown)


dwave/optimization/src/graph.cpp
```diff
@@ -157,6 +157,23 @@ void Graph::initialize_state(State& state) {
...
+    assert(0 <= i and "need at least one decision");
+
+    // Check that nodes_ and decisions_ are consistent (should always be true)
+    assert(nodes_[i].get() == decisions_[i]);
+
+    // Confirm that removing the decision won't leave anything dangling
+    assert(decisions_[i]->successors().empty() and "cannot remove a decision with successors");
+
+    decisions_.pop_back();
+    nodes_.erase(nodes_.begin() + i);
```
smharwood: Musing out loud: Should we be worried about the complexity here? Probably insignificant compared to everything else that's going on

smharwood (approved):
LGTM

dwave/optimization/src/graph.cpp
arcondello: I can run a few tests at the 2M node scale. My guess is that ripping through a vector like this isn't too bad but worth checking.
Also your comment made me realize that this logic is incorrect. The order of `decisions_` is deterministic, but not of `nodes_` (prior to topologically sorting).

dwave/optimization/src/graph.cpp
```diff
@@ -157,6 +157,40 @@ void Graph::initialize_state(State& state) {
...
+    decisions_.pop_back();
+
+    // Should never have the same decision twice and decisions are not inputs/constants
+    assert(std::ranges::none_of(decisions_, is_target));
+    assert(std::ranges::none_of(inputs_, is_target));
+    assert(std::ranges::none_of(constants_, is_target));
+
+    // Finally, we need to remove it from our node list. We do the swap and pop trick
+    // because we're not topologically sorted so it's OK for us to mess with the node order
+    auto it = std::find_if(nodes_.begin(), nodes_.end(), [&target_ptr](const auto& uptr) {
```
arcondello, line 186: This is still `O(num_nodes)` a bit of testing implies that it's OK

tests/cpp/test_graph.cpp
```diff
@@ -501,4 +550,36 @@ TEST_CASE("Graph::remove_unused_nodes()") {
...
+        CHECK_THAT(graph.decisions(), RangeEquals({x_ptr, z_ptr, y_ptr}));
+
+        CHECK(graph.nodes()[0].get() == x_ptr);
+        CHECK(graph.nodes()[1].get() == z_ptr);
+        CHECK(graph.nodes()[2].get() == y_ptr);
+
+        CHECK(x_ptr->topological_index() == 0);
+        CHECK(y_ptr->topological_index() == 2);
+        CHECK(z_ptr->topological_index() == 1);
+    }
```
smharwood, line 606: Maybe add that new intermediate variable before final decision test case here too

dwave/optimization/src/graph.cpp
```diff
@@ -545,6 +579,32 @@ void Graph::set_objective(ArrayNode* objective_ptr) {
...
+    assert(0 <= y_idx and static_cast<size_t>(y_idx) < decisions_.size());
+
+    assert(decisions_[x_idx] == x_ptr);
+    assert(decisions_[y_idx] == y_ptr);
+
+    assert(0 <= x_idx and static_cast<size_t>(x_idx) < nodes_.size());
+    assert(0 <= y_idx and static_cast<size_t>(y_idx) < nodes_.size());
+
+    assert(nodes_[x_idx].get() == static_cast<Node*>(x_ptr));
+    assert(nodes_[y_idx].get() == static_cast<Node*>(y_ptr));
```
smharwood: Since we are not topologically sorted yet, this might fail?

dwave/optimization/src/graph.cpp
```diff
@@ -545,6 +579,32 @@ void Graph::set_objective(ArrayNode* objective_ptr) {
...
+
+    using std::swap;  // ADL shouldn't matter here, but a good habit nonetheless
+    swap(nodes_[x_idx], nodes_[y_idx]);
```
smharwood: might need to fix the logic here too

arcondello: you're totally right 😳

smharwood:
LGTM for real this time :D

PR #604, opened 2026-08-10
Use SciPy for LinearProgram symbol calculations when available

### LINKED ISSUE #599: Use SciPy's LP solver when available
We should update the `LinearProgram` symbol and `LinearProgramNode` to use the SciPy's [linprog()](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.linprog.html) function when it's available.

This is an alternative way to address https://github.com/dwavesystems/dwave-optimization/issues/549 when combined with https://github.com/dwavesystems/dwave-optimization/pull/598

arcondello:
Closes https://github.com/dwavesystems/dwave-optimization/issues/599

#### AI Generation Disclosure
No AI used.


dwave/optimization/src/nodes/lp.cpp
```diff
@@ -346,17 +354,50 @@ void LinearProgramNode::propagate(State& state) const {
...
+            data->lp.b_lb,
+            data->lp.A,
+            data->lp.b_ub,
+            data->lp.A_eq,
+            data->lp.b_eq,
+            data->lp.lb,
+            data->lp.ub,
+            FEASIBILITY_TOLERANCE
+        );
+    }
```
arcondello, line 398: We could rework this section by changing our fallback simplex implementation to just return a solution. But that would mean rewriting most of the tests. IMO for now this is OK.

pyproject.toml
```diff
@@ -41,6 +41,7 @@ dependencies = [
 all = [
     "dimod>=0.12.0",  # oldest minor that passes tests
     "scikit-learn>=1.6.0",  # oldest minor that passes tests
+    "scipy>=1.10.1",  # last that supports our min NumPy version
```
arcondello: I'd like to rework the way we manage our optional dependency versions. Not to mention increasing `NumPy>=2`. But for now, I went with consistency with what's there.

tests/test_symbols.py
```diff
@@ -2741,6 +2742,63 @@ def test_serialization_with_states(self):
...
+    def test_fallback(self):
+        expected_c = np.asarray([1, 2, 3])
+        expected_A_eq = np.asarray([[4, 5, 6], [7, 8, 9]])
+        expected_b_eq = np.asarray([1, 2])
+        expected_lb = np.asarray([-1, -2, -3])
+        expected_ub = np.asarray([100, 101, 102])
+        expected_x = np.asarray([50, 51, 52], dtype=np.double)
+        count = [0]
+
+        def hi(**kwargs):
```
wbernoudy: Maybe `linprog_tester`?

wbernoudy (approved):

tests/test_symbols.py
arcondello: hah! I already made that exact change but hadn't pushed it yet (need to rebase off https://github.com/dwavesystems/dwave-optimization/pull/605 to fix the windows error).

PR #607, opened 2026-08-11
Add const overload for DisjointBitSetsNode::get_containing_set_index()

hurdwave:
#### Reference issue
N/A

Make DisjointBitSetsNode::get_containing_set_index() take const State&

#### Additional information
N/A

#### AI Generation Disclosure
No AI tools used.

### DIFF
--- a/dwave/optimization/include/dwave-optimization/nodes/collections.hpp
+++ b/dwave/optimization/include/dwave-optimization/nodes/collections.hpp
@@ -116,7 +116,6 @@ class DisjointBitSetsNode : public DecisionNode {
 
     void commit(State&) const override;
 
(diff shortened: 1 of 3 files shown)


alexzucca90 (commented):
should we just remove fully the non-const method?

arcondello:
> should we just remove fully the non-const method?

Agree. We should just mark the existing one as `const`.

hurdwave:
> > should we just remove fully the non-const method?
> 
> Agree. We should just mark the existing one as `const`.

Done

arcondello (approved):

alexzucca90 (approved):

PR #608, opened 2026-08-12
Use `PyCapsule` for creating `Symbol`s from `Node*`s

arcondello:
We need a way to smuggle a C++ pointer through Python. Before we used a (malformed) `Symbol` but Python provides [capsules](https://docs.python.org/3/c-api/capsule.html) for this exact purpose. It doesn't really change the amount of code, but IMO this is a nicer mechanism.

Eventually this will help remove a circular dependency between `_Graph` and `Symbol`.

#### AI Generation Disclosure
No AI used.

### DIFF
--- a/dwave/optimization/symbols/naryop.pyx
+++ b/dwave/optimization/symbols/naryop.pyx
@@ -56,14 +56,14 @@ cdef class NaryAdd(ArraySymbol):
         self.initialize_arraynode(model, self.ptr)
 
     @classmethod
-    def _from_symbol(cls, Symbol symbol):
-        cdef NaryAddNode* ptr = dynamic_cast_ptr[NaryAddNode](symbol.node_ptr)
-        if not ptr:
-            raise TypeError(f"given symbol cannot construct a {cls.__name__}")
-        cdef NaryAdd x = NaryAdd.__new__(NaryAdd)
-        x.ptr = ptr
-        x.initialize_arraynode(symbol.model, ptr)
-        return x
+    def _from_ptr(cls, model, capsule):
+        cdef NaryAdd sym = super()._from_ptr(model, capsule)
+
+        sym.ptr = dynamic_cast_ptr[NaryAddNode](sym.node_ptr)
+        if not sym.ptr:
+            raise TypeError(f"given pointer cannot construct a NaryAdd")
+
+        return sym
 
     def __iadd__(self, rhs):
         if not self.node_ptr.successors().empty():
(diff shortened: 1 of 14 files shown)


dwave/optimization/symbols/set_routines.pyx
```diff
@@ -82,7 +81,12 @@ cdef class IsDisjointCover(ArraySymbol):
         args.update(primary_set_size=int(self.primary_set_size))
         zf.writestr(directory + "args.json", encoder.encode(args))
 
-    cdef Py_ssize_t primary_set_size
+    @property
+    def primary_set_size(self):
```
arcondello, line 85: This way we can use the node as the source of truth.

dwave/optimization/symbols/naryop.pyx
```diff
@@ -169,14 +169,14 @@ cdef class NaryMultiply(ArraySymbol):
...
-    def _from_symbol(cls, Symbol symbol):
-        cdef NaryMultiplyNode* ptr = dynamic_cast_ptr[NaryMultiplyNode](symbol.node_ptr)
-        if not ptr:
-            raise TypeError(f"given symbol cannot construct a {cls.__name__}")
-        cdef NaryMultiply x = NaryMultiply.__new__(NaryMultiply)
-        x.ptr = ptr
-        x.initialize_arraynode(symbol.model, ptr)
-        return x
+    def _from_ptr(cls, model, capsule):
+        cdef NaryMultiply sym = super()._from_ptr(model, capsule)
```
fastbodin, line 173: Unrelated to this PR: Is there a reason `NaryMinimum` and `NaryMaximum` do not have `_from_symbol()` (now `_from_ptr()`)?

fastbodin (approved):
LGTM.

dwave/optimization/symbols/naryop.pyx
arcondello: The better question is why do `NaryAdd` and `NaryMultiply` have it? 😄 

The answer is that `NaryAdd` and `NaryMultiply` need access to a `NaryAddNode*` and `NaryMultiplyNode*` respectively, so they hold a pointer of that type on the class (we could refactor that away, but that's a different question). They need it to implement `__iadd__()` and `__imul__()`. The other two are OK only having an `ArrayNode*` so they don't need to implement a method to set `self.ptr`.

fastbodin: Makes sense. Thanks for explaining!

PR #610, opened 2026-08-12
Remove `StateView` class in favour of an inline `Py_buffer` creation

arcondello:
Avoid a redundant class when reading the state of an `ArrayNode`. This helps address (but does not solve) a circular dependency between `states.pyx` and `_model.pyx`.

#### AI Generation Disclosure
Claude helped with the `memset()` trick.


dwave/optimization/_model.pyx
```diff
@@ -25,11 +25,12 @@ import zipfile
 
 import numpy as np
 
-from cpython cimport Py_buffer
```
arcondello, line 28: Turns out this wasn't needed and it was bringing in a whole pile of unneeded `#include`s

dwave/optimization/_model.pyx
```diff
@@ -2495,7 +2495,47 @@ cdef class ArraySymbol(Symbol):
             raise TypeError("the state of an intermediate variable cannot be accessed without "
                             "locking the model first. See model.lock().")
 
-        return np.array(StateView(self, index), copy=copy)
+        # Make sure there is a state to access
+        cdef States states = self.model.states
+        states.resolve()
```
fastbodin, line 2500: Previously, this call had the following comment: 
```py
# todo: inspect/respect/test flags
```
Still relevant?

fastbodin (approved):

dwave/optimization/_model.pyx
arcondello: Wasn't even relevant in the old code! We should have removed that todo in https://github.com/dwavesystems/dwave-optimization/pull/301

Also, it's extra not relevant here because there are no flags requested when we make the `memoryview` directly from a buffer. Another benefit of the approach in this PR.

fastbodin: Perfect. I had been confused what additional things were left to "inspect/respect/test" in the old code 🤣

PR #612, opened 2026-08-13
Clean up imports in `states.pyx`/`states.pxd`

arcondello:
There were a bunch of unused imports and we were using the old `cpp*` convention that I'd like to move away from.

#### AI Generation Disclosure
No AI used.

### DIFF
--- a/dwave/optimization/states.pyx
+++ b/dwave/optimization/states.pyx
@@ -17,13 +17,9 @@ import tempfile
 import weakref
 import zipfile
 
-cimport cpython.buffer
-
 from libcpp.utility cimport move
 
-from dwave.optimization.libcpp.array cimport Array as cppArray
-from dwave.optimization.model cimport ArraySymbol, _Graph
-from dwave.optimization.model import Model
+from dwave.optimization.model cimport _Graph
 from dwave.optimization.utilities import _file_object_arg
 
 __all__ = ["States"]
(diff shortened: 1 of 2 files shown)


dwave/optimization/states.pyx
```diff
@@ -87,15 +83,13 @@ cdef class States:
         :attr:`~dwave.optimization.model.Model.states`
     """
     def __init__(self, model):
-        if not isinstance(model, Model):
```
arcondello, line 90: The removal of this error message is the only functional change. `import`-ing `Model` just for this type check seems like overkill for a class that no user constructs directly and for which we already have `typing` https://github.com/dwavesystems/dwave-optimization/blob/2ba3d79fb306bd68d385222167ea2b709b51c6cf/dwave/optimization/states.pyi#L22

dwave/optimization/states.pyx
```diff
@@ -459,7 +453,7 @@ cdef class States:
                 version=version,
             )
 
-    cdef _Graph _model(self):
+    def _model(self):
```
smharwood, line 456: presumably/technically there is overhead in changing this to `def` but who cares?

arcondello: Right, it's kind of in the rounding error of the deref. Anywhere that wants to access it can still do
```cython
cdef _Graph model = self._model()
```
which amounts to the same thing. With the very very minor overhead of the Python function call.

smharwood (approved):

PR #615, opened 2026-08-13
Remove `Symbol.equals()` and `Symbol.maybe_equals()` methods

arcondello:
This is a backwards compatibility break

#### Description

These functions had several issues
* Confusing name/purpose. They were meant to allow comparison between symbols across different models (e.g., for testing serialization) but the name is confusing and they incidentally worked within models.
* We didn't correctly implement them. Most symbols didn't have the overloads they were meant to.

We might want to add some new methods in the future for this purpose, but for now IMO we should just drop these.

#### AI Generation Disclosure
No AI used.


tests/test_serialization.py
```diff
@@ -45,8 +45,6 @@ def assertModelEqual(self, lhs: Model, rhs: Model):
 
             # All nodes in the model need to match, and to have the same states
             for ls, rs in zip(lhs.iter_symbols(), rhs.iter_symbols()):
-                self.assertTrue(ls.maybe_equals(rs))
```
arcondello, line 48: I should do `type(ls) is type(rs)` here at least

fastbodin (approved):
Baring above small change, LGMT

PR #616, opened 2026-08-13
Add Symbol and ArraySymbol to dwave.optimization.typing

### LINKED ISSUE #603: Re-export `ArraySymbol` from `dwave.optimization.typing`
AFAIK `ArraySymbol` is not publicly re-exported anywhere, and I'd rather not import it directly from `dwave.optimization._model`.

We commonly do `from dwave.optimization.model import ArraySymbol` but pylance flags that as ["private import usage"](https://github.com/microsoft/pylance-release/blob/main/docs/diagnostics/reportPrivateImportUsage.md)

Since `ArraySymbolLike` is defined as
```
ArraySymbolLike: typing.TypeAlias = ArraySymbol | numpy.typing.ArrayLike
```
type checkers like pylance don't like it when you use `ArraySymbol` methods like `.ndim()` because its not defined on the `ArrayLike`.
Changing the definition to `ArraySymbolLike: typing.TypeAlias = ArraySymbol | numpy.typing.NDArray` might not help because `ndim`, `shape`, etc are attributes rather than callables on the `NDArray`

arcondello:
image

Closes https://github.com/dwavesystems/dwave-optimization/issues/603

#### AI Generation Disclosure
No AI used.

### DIFF
--- a/dwave/optimization/typing.py
+++ b/dwave/optimization/typing.py
@@ -28,8 +28,7 @@ __all__ = [
 ]
 
 ArraySymbolLike: typing.TypeAlias = ArraySymbol | numpy.typing.ArrayLike
-"""Either a :class:`~dwave.optimization.ArraySymbol` or a NumPy
-`array-like <https://numpy.org/doc/stable/glossary.html#term-array_like>`_.
+"""Either a :class:`~dwave.optimization.ArraySymbol` or a NumPy |array-like|_.
 """
 
 ShapeLike: typing.TypeAlias = int | tuple[int, ...]
--- a/dwave/optimization/typing.py
+++ b/dwave/optimization/typing.py
@@ -16,15 +16,20 @@ import typing
 
 import numpy.typing
 
-from dwave.optimization._model import ArraySymbol
+from dwave.optimization._model import ArraySymbol, Symbol
 
+# Maintenance note: classes that are imported into this namespace must be
+# added manually to the typing.rst docs.
 __all__ = [
+    "ArraySymbol",
     "ArraySymbolLike",
+    "ShapeLike",
+    "Symbol",
 ]
 
 ArraySymbolLike: typing.TypeAlias = ArraySymbol | numpy.typing.ArrayLike
 """Either a :class:`~dwave.optimization.ArraySymbol` or a NumPy
-`array-like <https://numpy.org/devdocs/glossary.html#term-array_like>`_.
+`array-like <https://numpy.org/doc/stable/glossary.html#term-array_like>`_.
 """
 
 ShapeLike: typing.TypeAlias = int | tuple[int, ...]
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -101,5 +101,5 @@ intersphinx_mapping = {
 
 rst_epilog = """
 .. |array-like| replace:: array-like
-.. _array-like: https://numpy.org/devdocs/glossary.html#term-array_like
+.. _array-like: https://numpy.org/doc/stable/glossary.html#term-array_like
 """
--- a/docs/typing.rst
+++ b/docs/typing.rst
@@ -4,7 +4,20 @@
 Typing
 ======
 
+Types Aliases
+=============
+
 .. currentmodule:: dwave.optimization
 
 .. automodule:: dwave.optimization.typing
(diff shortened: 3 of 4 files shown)


dwave/optimization/typing.py
```diff
@@ -16,15 +16,20 @@
...
+    "ArraySymbol",
     "ArraySymbolLike",
+    "ShapeLike",
+    "Symbol",
 ]
 
 ArraySymbolLike: typing.TypeAlias = ArraySymbol | numpy.typing.ArrayLike
 """Either a :class:`~dwave.optimization.ArraySymbol` or a NumPy
-`array-like <https://numpy.org/devdocs/glossary.html#term-array_like>`_.
+`array-like <https://numpy.org/doc/stable/glossary.html#term-array_like>`_.
```
JoelPasvolsky: we have |array-like|_ so this should be the substitution and this update should happen [here](https://github.com/dwavesystems/dwave-ocean-sdk/blob/5574c9704cb20786360fb5031e154cf3e9da9092/docs/conf.py#L434)

JoelPasvolsky (commented):
We don't run doctest here?

arcondello:
> We don't run doctest here? 

They're running right now, just gated behind the linux tests 
image

JoelPasvolsky:
```
/home/circleci/project/docs/philosophy.rst:7: ERROR: Undefined substitution referenced: "TM". [docutils]
<rst_epilog>:3: WARNING: Duplicate explicit target name: "array-like". [docutils]
```

arcondello:
Those are unrelated to this PR. I have another PR on the way to fix doc warnings/errors.

dwave/optimization/typing.py
arcondello: We have the same substitution in the `dwave-optimization` `conf.py` file https://github.com/dwavesystems/dwave-optimization/blob/2ba3d79fb306bd68d385222167ea2b709b51c6cf/docs/conf.py#L103-L104
but it doesn't seem to work.

Should it instead be
```
.. |array-like| replace:: array-like_
```
with the trailing underscore?

arcondello: ^ this is what I went with. Can revert it if it's wrong.

JoelPasvolsky: to use it, you add the underscore after the substitution in the RST file: |substitution text|_

arcondello: That seems worse. Should we just do `array-like_` then?

JoelPasvolsky: In the online docs, the build uses the SDK `conf.py` file so if your docstrings use something that is only defined in `dwave-optimization` they links won't work

arcondello: > That seems worse.

In the sense that it's an abstraction leak. Like I am saying both that I know it will be substituted, and that that substitution will result in a valid link. Whereas with just `|array-like|` I am just leaving it up to substitution to decide whether it's a link or not.

arcondello: > In the online docs, the build uses the SDK conf.py file so if your docstrings use something that is only defined in dwave-optimization they links won't work

Yes, I understand that. We should change it upstream as well.

JoelPasvolsky: it's the standard way of doing this

arcondello: > it's the standard way of doing this

citation needed?

I am just going to revert to the inline link for now.

JoelPasvolsky: https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#substitution-references

```
A substitution reference may also be a [hyperlink reference](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#hyperlink-reference) by appending a _ (named) or __ ([anonymous](https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#anonymous)) suffix```
```

See example in that definitive text, sir

JoelPasvolsky (approved):
Thank you

dwave/optimization/typing.py
arcondello: Boo. I hate it. But OK, I was wrong. Thank you for schooling me.

PR #620, opened 2026-08-14
Add keep argument for Graph::remove_unused_nodes()

arcondello:
#### AI Generation Disclosure
No AI used.

### DIFF
--- a/dwave/optimization/src/graph.cpp
+++ b/dwave/optimization/src/graph.cpp
@@ -303,20 +303,20 @@ ssize_t Graph::remove_redundant_nodes(bool ignore_listeners, double time_limit_s
     // already done.
     // They are arbitrary values, but we steer away from -1 because that's used to
     // indicate unsorted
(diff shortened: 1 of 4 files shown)


dwave/optimization/src/graph.cpp
```diff
@@ -303,20 +303,20 @@ ssize_t Graph::remove_redundant_nodes(bool ignore_listeners, double time_limit_s
     // already done.
     // They are arbitrary values, but we steer away from -1 because that's used to
     // indicate unsorted
-    constexpr ssize_t seen = -2;  // already checked for duplicates
-    constexpr ssize_t drop = -3;  // marked for deletion
+    constexpr ssize_t SEEN = -2;  // already checked for duplicates
```
arcondello, line 306: This is to keep a standard convention between the two `Graph::remove_` methods.

smharwood (approved):

PR #624, opened 2026-08-18
Support C++23

### LINKED ISSUE #577: Add C++23 support
We're relatively close to being able to support C++23 (maybe). The main holdup is Windows. I believe CircleCI has support now, but we need `meson` 1.12

See https://github.com/dwavesystems/dwave-optimization/issues/502#issuecomment-4724329807

arcondello:
To take advantage of most C++23 features, we require GCC 14+. We can't go higher than 14 because we build `manylinux2_28` wheels [which use GCC 14](https://github.com/pypa/manylinux#manylinux_2_28-almalinux-8-based).

Closes https://github.com/dwavesystems/dwave-optimization/issues/577

#### AI Generation Disclosure
I consulted Claude about the feasibility of using LLVM for Windows wheels because testing live in CI is a pain. And it was very helpful in diagnosing a bug that came from using `clang++` rather than `clang-cl`.

I also asked Claude to find places where switching to C++23 might help performance or clarity. Most of the changes were identified by Claude and I incorporated some of its suggestions. Though I wrote all of the code myself.

### DIFF
--- a/dwave/optimization/include/dwave-optimization/nodes/manipulation.hpp
+++ b/dwave/optimization/include/dwave-optimization/nodes/manipulation.hpp
@@ -270,9 +270,9 @@ class ReshapeNode : public ArrayOutputMixin<EqualityMixin<ArrayNode, ReshapeNode
     ///
     /// @param array_ptr The array to be reshaped.
     /// @param shape The new shape. Must have the same size as the original shape.
-    template <std::ranges::range Range>
-    ReshapeNode(ArrayNode* node_ptr, Range&& shape) :
-        ReshapeNode(node_ptr, std::vector<ssize_t>(shape.begin(), shape.end())) {}
+    template <std::ranges::range R>
+    ReshapeNode(ArrayNode* node_ptr, R&& shape) :
+        ReshapeNode(node_ptr, std::ranges::to<std::vector<ssize_t>>(std::forward<R>(shape))) {}
 
     /// @copydoc Array::buff()
     double const* buff(const State& state) const override;
@@ -348,9 +348,13 @@ class ResizeNode : public ArrayOutputMixin<EqualityMixin<ArrayNode, ResizeNode>>
     /// @param array_ptr The array to be resized.
     /// @param shape The new shape. Must not be dynamic.
     /// @param fill_value The value to use for missing values.
-    template <std::ranges::range Range>
-    ResizeNode(ArrayNode* node_ptr, Range&& shape, double fill_value = 0) :
-        ResizeNode(node_ptr, std::vector<ssize_t>(shape.begin(), shape.end()), fill_value) {}
+    template <std::ranges::range R>
+    ResizeNode(ArrayNode* node_ptr, R&& shape, double fill_value = 0) :
+        ResizeNode(
+            node_ptr,
+            std::ranges::to<std::vector<ssize_t>>(std::forward<R>(shape)),
+            fill_value
+        ) {}
 
     /// @copydoc Array::buff()
     double const* buff(const State& state) const override;
--- a/dwave/optimization/src/nodes/binaryop.cpp
+++ b/dwave/optimization/src/nodes/binaryop.cpp
@@ -19,6 +19,7 @@
 #include <ranges>
 #include <stdexcept>
 #include <type_traits>
+#include <utility>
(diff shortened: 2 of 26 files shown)


arcondello:
I'd like to sweep clang versions like we do gcc in CI, but I only see an unofficial clang docker. For posterity, search/claude tells me that our floor is likely clang17 (as roughly corresponding to gcc14).

meson.build
```diff
@@ -19,12 +19,14 @@ project(
     ).stdout(),
 )
 
-# We want some debugging symbols
-add_project_arguments(['-g1'], language: ['cpp'])
-
 cpp = meson.get_compiler('cpp')
 py = import('python').find_installation(pure: false)
 
+# We want some debugging symbols
```
wbernoudy, line 25: Any specific reason you're including this?

wbernoudy (approved):
So exciting!!

wbernoudy:
> I'd like to sweep clang versions like we do gcc in CI, but I only see an unofficial clang docker. For posterity, search/claude tells me that our floor is likely clang17 (as roughly corresponding to gcc14).

By pure count of feature support, I think it takes clang 20 or 21 to match gcc 14 based on [the tables at cppreference](https://en.cppreference.com/cpp/compiler_support/23). But I think clang 17 should be fine.

meson.build
arcondello: The (lack of) `if`-guard wasn't broken but the compilers were throwing a warning and I reasoned that was reasonably in scope. It moved because we need access to `cpp`.

arcondello:
I'll leave https://github.com/dwavesystems/dwave-optimization/blob/11fcfd9cc9cfd4bf4e1b7debf47e0405dd39e219/dwave/optimization/include/dwave-optimization/array.hpp#L732-L733 to a separate PR. IMO this is a good time to change `deduplicate_diff()` to not act on a vector in-place.

dwave/optimization/include/dwave-optimization/nodes/manipulation.hpp
```diff
@@ -270,9 +270,9 @@ class ReshapeNode : public ArrayOutputMixin<EqualityMixin<ArrayNode, ReshapeNode
     ///
     /// @param array_ptr The array to be reshaped.
     /// @param shape The new shape. Must have the same size as the original shape.
-    template <std::ranges::range Range>
-    ReshapeNode(ArrayNode* node_ptr, Range&& shape) :
-        ReshapeNode(node_ptr, std::vector<ssize_t>(shape.begin(), shape.end())) {}
```
arcondello, line 275: technically these weren't compliant before! Because `std::vector(begin, end)` requires `begin` and `end` to have the same type and `std::ranges::range` does not

dwave/optimization/src/nodes/binaryop.cpp
```diff
@@ -268,10 +268,8 @@ void BinaryOpNode<BinaryOp>::initialize_state(State& state) const {
         // This is the easy case - all we need to do is iterate over both as flat arrays
         values.reserve(lhs_ptr->size(state));
 
-        auto it = lhs_ptr->begin(state);
-        for (const double val : rhs_ptr->view(state)) {
-            values.emplace_back(op(*it, val));  // order is important
```
wbernoudy, line 272: Dropped the comment, but it's probably not necessary

dwave/optimization/include/dwave-optimization/nodes/manipulation.hpp
```diff
@@ -270,9 +270,9 @@ class ReshapeNode : public ArrayOutputMixin<EqualityMixin<ArrayNode, ReshapeNode
...
+    template <std::ranges::range R>
```
wbernoudy, line 273: This is just for naming consistency?

wbernoudy (approved):

dwave/optimization/include/dwave-optimization/nodes/manipulation.hpp
arcondello: A few reasons:
* Internal consistency
* `std` seems to usually use `R`? But isn't terribly consistent
* Make the subsequent `std::ranges::to<std::vector<ssize_t>>(std::forward<R>(...))` line slightly shorter

PR #625, opened 2026-08-20
Replace `deduplicate_diff_view` with `views::deduplicate_diff`

arcondello:
Builds off of https://github.com/dwavesystems/dwave-optimization/pull/624 but requires a backwards compatibility break so making it a separate PR.

#### AI Generation Disclosure
Claude was helpful in helping to understand what the difference between a view and a range adaptor closure is.

### DIFF
--- a/dwave/optimization/src/nodes/manipulation.cpp
+++ b/dwave/optimization/src/nodes/manipulation.cpp
@@ -257,7 +257,7 @@ void BroadcastToNode::propagate(State& state) const {
 
     ssize_t size_diff = 0;
     if (this->size() != 0) {
-        for (Update update : deduplicate_diff_view(from_diff)) {
+        for (Update update : from_diff | views::deduplicate_diff) {
             // we need to convert from our predecessor's index to ours
             const ssize_t index = convert_predecessor_index_(update.index);
             assert(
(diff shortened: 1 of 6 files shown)


dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -1401,13 +1401,7 @@ void RollNode::propagate(State& state) const {
             if (array_ptr_->dynamic()) {
                 // One last case we need to worry about. If the array may have grown and then
                 // shrunk, we need to deduplicate the diff.
-                // We'd really like to use some of the nice C++20 range stuff and
-                // deduplicate_diff_view(), but alas some of the old Python images
-                // don't support deduplicate_diff_view with std::ranges::transform.
-                // So we have to do it manually with a copy.
```
fastbodin, line 1407: Past you knew the way

dwave/optimization/src/nodes/manipulation.cpp
```diff
@@ -257,7 +257,7 @@ void BroadcastToNode::propagate(State& state) const {
 
     ssize_t size_diff = 0;
     if (this->size() != 0) {
-        for (Update update : deduplicate_diff_view(from_diff)) {
+        for (Update update : from_diff | views::deduplicate_diff) {
```
fastbodin, line 260: Love it. So fresh and so clean

fastbodin (approved):
Love it

PR #626, opened 2026-08-26
Add interval class

arcondello:
Alternative to https://github.com/dwavesystems/dwave-optimization/pull/550

#### AI Generation Disclosure
Had Claude Opus review the code before pushing it. It identified a few minor issues.

### DIFF
--- a/dwave/optimization/include/dwave-optimization/interval.hpp
+++ b/dwave/optimization/include/dwave-optimization/interval.hpp
@@ -1,0 +1,195 @@
+// Copyright 2026 D-Wave
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+#pragma once
+
+#include <cassert>
+#include <compare>
+#include <concepts>
+#include <iosfwd>
+#include <limits>
+
+#include "dwave-optimization/typing.hpp"
+
+namespace dwave::optimization {
+
+/// An interval encodes a range of possible values.
+///
+/// Note that this class does not (yet) implement outward rounding.
+template <DType T>
+struct interval {
+    /// Construct an empty interval.
+    constexpr interval() = default;
+
+    /// Construct an interval of values between inf and sup (inclusive).
+    /// When ``sup < inf`` the interval is treated as empty.
+    constexpr interval(T inf, T sup) noexcept : infimum(inf), supremum(sup) {}
+
+    /// Copy constructor.
+    interval(const interval&) = default;
+
+    /// Create an ``interval<double>`` from another interval.
+    template <DType U>
+    requires(std::same_as<T, double>)
+    interval(const interval<U>& other) noexcept : interval(other.infimum, other.supremum) {}
+    // dev note: we could expand this. E.g., we could support all promotions
+    // allowed by NumPy promotion.
+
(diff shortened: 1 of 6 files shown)


dwave/optimization/include/dwave-optimization/interval.hpp
```diff
@@ -0,0 +1,195 @@
...
+
+    /// Copy constructor.
+    interval(const interval&) = default;
+
+    /// Create an ``interval<double>`` from another interval.
+    template <DType U>
+    requires(std::same_as<T, double>)
+    interval(const interval<U>& other) noexcept : interval(other.infimum, other.supremum) {}
+    // dev note: we could expand this. E.g., we could support all promotions
+    // allowed by NumPy promotion.
```
fastbodin, line 47: Confirming, this comment is specifically applicable to 
```cpp
requires(std::same_as<T, double>)
```

dwave/optimization/include/dwave-optimization/interval.hpp
```diff
@@ -0,0 +1,195 @@
...
+
+    /// An interval evalutes to `true` if it is not empty.
+    explicit constexpr operator bool() const noexcept { return infimum <= supremum; }
+
+    /// Two intervals are treated as equal if they are the same type and have the same endpoints
+    /// or if they are both null.
+    constexpr bool operator==(const interval& rhs) const {
+        if (not static_cast<bool>(*this) and not static_cast<bool>(rhs)) return true;  // both null
+        return infimum == rhs.infimum and supremum == rhs.supremum;
+    }
```
fastbodin, line 69: Confirming my understanding: Despite that there is no explicit handling of "if they are the same type", this is handled by the compiler, i.e.
```cpp
    constexpr bool operator==(const interval& rhs) const {
```
really expands to
```cpp
    constexpr bool operator==(const interval<T>& rhs) const {
```
So we would get a compiler error if we tried to compare a `bool` interval with a `double` (unless we defined an explicit `operator==` for each `DType` pair). Hence the "dev note: we could support other type combinations in the future"

dwave/optimization/include/dwave-optimization/interval.hpp
```diff
@@ -0,0 +1,195 @@
...
+        using limits = std::numeric_limits<T>;
+        if constexpr (limits::has_infinity) {
+            return interval(-limits::infinity(), limits::infinity());
+        } else {
+            return interval(limits::lowest(), limits::max());
+        }
+    }
+
+    /// Test whether `x` is a value in the interval.
+    constexpr bool contains(const T& x) const { return infimum <= x and x <= supremum; }
```
fastbodin, line 174: To handle the edge-case of an empty interval, should this not be
```cpp
    constexpr bool contains(const T& x) const { return static_cast<bool>(*this) and infimum <= x and x <= supremum; }
```

fastbodin (approved):
LGTM. Very neat and tidy.

dwave/optimization/include/dwave-optimization/interval.hpp
arcondello: yes, exactly

dwave/optimization/include/dwave-optimization/interval.hpp
fastbodin: After finishing the review, I realized this should be fine as-is (i.e. without the above change) because of the default `inf` and `sup`.

dwave/optimization/include/dwave-optimization/interval.hpp
arcondello: yes. So we allow `interval<double>(interval<int16_t>(0, 5))` but not  `interval<int16_t>(interval<double>(0, 5))`

dwave/optimization/include/dwave-optimization/interval.hpp
arcondello: It's empty exactly when `supremum < infimum` in which case this check will always fail.

dwave/optimization/include/dwave-optimization/interval.hpp
fastbodin: Exactly what I thought. Cheers.

PR #627, opened 2026-08-26
Rework `UnaryOpNode`

arcondello:
Make `UnaryOpNode` header-only.

Builds off of https://github.com/dwavesystems/dwave-optimization/pull/632, I will rebase once that one is merged.

#### AI Generation Disclosure
No AI used for the `UnaryOpNode` changes, see https://github.com/dwavesystems/dwave-optimization/pull/632 for AI disclosure for the `functional.hpp` changes.


tests/cpp/nodes/test_unaryop.cpp
```diff
@@ -139,13 +139,15 @@ TEMPLATE_TEST_CASE(
...
+        auto* a_ptr = graph.emplace_node<IntegerNode>(
             std::span<const ssize_t>{},
             -100,
             100
         );  // Scalar output
 
-        auto p_ptr = graph.emplace_node<UnaryOpNode<TestType>>(a_ptr);
+        auto* p_ptr = graph.emplace_node<UnaryOpNode<TestType>>(a_ptr);
+
+        graph.emplace_node<ArrayValidationNode>(p_ptr);
```
arcondello, line 150: This test was modified more than the others as part of fixing a bug. I figure the changes are for the better but are not so important that we need to propagate them to all other tests.

dwave/optimization/include/dwave-optimization/nodes/unaryop.hpp
```diff
@@ -69,31 +100,220 @@ class UnaryOpNode : public ArrayOutputMixin<ArrayNode> {
...
+using ExpitNode = UnaryOpNode<functional::expit>;
+extern template class UnaryOpNode<functional::expit>;
+
+using LogNode = UnaryOpNode<functional::log>;
+extern template class UnaryOpNode<functional::log>;
+
+using LogicalNode = UnaryOpNode<functional::logical>;
+extern template class UnaryOpNode<functional::logical>;
+
+using NegativeNode = UnaryOpNode<functional::negate>;
```
arcondello: I think now that these are all ops we define rather than `std::` ones, we should rename them to be consistent with the nodes.

dwave/optimization/include/dwave-optimization/nodes/unaryop.hpp
```diff
@@ -15,51 +15,82 @@
 #pragma once
 
 #include <cassert>
-#include <functional>
-#include <optional>
+#include <concepts>
```
alexzucca90: can you help me check where this is used?

PR #632, opened 2026-09-08
Rework unary ops

arcondello:
Rework the ops in `functional.hpp` to carry more information. This is then used in some bigger changes to:
* `UnaryOpNode` https://github.com/dwavesystems/dwave-optimization/pull/627

#### AI Generation Disclosure

Claude helped a lot with this one. Mostly by catching edge cases in the funcs, implementing a few of them once I showed it the pattern (e.g., `rint`) and by helping to write the tests. In the end, most of the code was actually written by me once I made all the changes I wanted.


dwave/optimization/include/dwave-optimization/functional.hpp
```diff
@@ -15,45 +15,185 @@
...
+    static auto operator()(const interval<T>& domain) {
+        using return_type = interval<decltype(UnaryOp::operator()(T()))>;
+
+        // op(empty domain) -> empty domain
+        if (not static_cast<bool>(domain)) return return_type();
+
+        assert(
+            domain <= UnaryOp::template domain<T> and
+            "input domain must be a subset of the func's domain"
+        );
```
fastbodin, line 45: Took me a HOT minute to realize the two `domain` here were different 🤦

dwave/optimization/include/dwave-optimization/functional.hpp
```diff
@@ -15,45 +15,185 @@
...
+    template <DType T>
+    static constexpr interval<T> domain = interval<T>::nonnegative();
+
+    static constexpr Monotonicity monotonic = Monotonicity::Increasing;
+};
+
+struct logical : UnaryOpMixin<logical> {
+    static bool operator()(const DType auto& x) { return x; }
+
+    static interval<bool> operator()(const interval<bool>& domain) { return domain; }
```
fastbodin, line 165: Confirming that, the reason we do not need
```cpp
        if (not static_cast<bool>(domain)) return {};  // op(empty domain) -> empty domain
``` 
here is because it is absorbed by returning `domain`.

PR #633, opened 2026-09-09
Add concepts/classes for NumPy-style type promotions

arcondello:
The next step after https://github.com/dwavesystems/dwave-optimization/pull/632 is to support binary ops. But for that, we need to codify our type promotion rules (following NumPy as usual).

I also used the new type promotion concept in the interval class added in https://github.com/dwavesystems/dwave-optimization/pull/626, mostly to show the pattern.

#### AI Generation Disclosure
Used Claude to review the code. No major changes from that review.


dwave/optimization/include/dwave-optimization/interval.hpp
```diff
@@ -39,12 +39,10 @@ struct interval {
...
 
-    /// Create an ``interval<double>`` from another interval.
-    template <DType U>
-    requires(std::same_as<T, double>)
-    interval(const interval<U>& other) noexcept : interval(other.infimum, other.supremum) {}
-    // dev note: we could expand this. E.g., we could support all promotions
-    // allowed by NumPy promotion.
+    /// Create an one interval from another, allowing for safe type promotions
+    template <can_cast<T> U>
+    constexpr interval(const interval<U>& other) noexcept :
```
arcondello, line 44: No release note needed for this change because the interval class is not yet released.

Also there are other interval methods that might benefit from type promotion, but I've decided to leave those alone for now.

PR #634, opened 2026-09-09
Drop support for Python 3.10

arcondello:
#### AI Generation Disclosure
I asked Claude to find places where we can benefit from dropping 3.10, and it didn't find anything. I also looked myself obviously.


.circleci/config.yml
```diff
@@ -387,7 +387,7 @@ jobs:
 
   deploy:
     docker:
-      - image: cimg/python:3.10
+      - image: cimg/python:3.14
```
randomir, line 390: Might as well use official python image for consistency:

```suggestion
      - image: python:3.14
```

randomir (approved):
LGTM!

.circleci/config.yml
arcondello: We're somewhat deliberately inconsistent. For this `deploy` step it doesn't matter, but for `python-linux` we need the `cimg/` one. So there isn't one style to be consistent with :shrug:

PR #637, opened 2026-09-11
Fix empty number nodes with axis bounds

### LINKED ISSUE #635: Using `axes_sums_subject_to` on a `BinaryVariable` with size 0 results in a crash when initializing a state
```python
In [1]: from dwave.optimization import Model

In [2]: model = Model()

In [3]: b = model.binary(shape=0, axes_sums_subject_to=[(0, ">=", 1)])

In [4]: model.lock()
Out[4]: <contextlib._GeneratorContextManager at 0x7ffff13952b0>

In [5]: model.states.resize(1)

In [6]: b.state(0)
zsh: floating point exception (core dumped)  ipython
```

arcondello:
Closes https://github.com/dwavesystems/dwave-optimization/issues/635

#### AI Generation Disclosure
No AI used.


dwave/optimization/src/nodes/numbers.cpp
```diff
@@ -643,6 +643,21 @@ void construct_state_given_exactly_one_sum_constraint(
...
+    if (node.size() == 0) {
+        assert(values.size() == 0);
+
+        double delta = sum_constraint_delta(0, constraint.op(0), constraint.bound(0));
+        if (delta) {
+            throw std::invalid_argument("Infeasible sum constraint.");
+        }
+
+        return;  // we've trivially satisfied the constraint
+    }
```
fastbodin, line 659: Having got to this point in the code, we know that the sum constraint is defined with respect to an `axis`. Given this, there may be different constraints per `slice`. In this snippet of code, we check the `delta` for the `0`th slice. Because of this, users can technically define infeasible constraints that "pass" this check. e.g.
```cpp
GIVEN("A (4,0)-BinaryNode with an invalid sum constraint on axis 0") {
    std::vector<SumConstraint> sum_constraints{{0, {GreaterEqual, Equal, Equal, Equal}, {1, 1, 2, 1}}};
    REQUIRE_THROWS_WITH(
        graph.emplace_node<BinaryNode>(
            std::vector<ssize_t>{4, 0}, std::nullopt, std::nullopt, sum_constraints
        ),
        "Infeasible sum constraint."
    );
}
```

fastbodin: If we want to avoid that, we could do something like
```cpp
    // handle the special case that our array is empty
    if (node.size() == 0) {
        assert(values.size() == 0);

        for (ssize_t i = 0, stop = std::max(static_cast<ssize_t>(1), node_shape[*axis]); i < stop; ++i) {
            const double delta = sum_constraint_delta(0, constraint.op(i), constraint.bound(i));
            if (delta) {
                throw std::invalid_argument("Infeasible sum constraint.");
            }
        }

        return;  // we've trivially satisfied the constraint
    }
```

arcondello: I should have figured that out from the `0` arg to the `sum_constraint_delta`. My fault for not reading it more carefully.

arcondello: Ok, added

fastbodin (approved):

